PR #483, opened 2013-08-22
fix \&/#bifoldMap

xuwei-k:
https://github.com/scalaz/scalaz/pull/452#issuecomment-21901865


wsh:
I'd like to respectfully suggest that this same commit be cherry-picked onto the 7.0.x branch, too. My team is trying to move to 7.1.x but, in the interim, we had a major bug in production because of the...surprising behavior of ```bifoldMap``` in 7.0.x. What do you think? @purefn @xuwei-k @larsrh

PR #608, opened 2013-12-17
Task: Catch Throwable, not just Exception

jessitron:
Task.Try catches Exception, but not Throwable.
The consequence is that when trampolining, if a Throwable pops up then the thread dies and the program never terminates.

It's a good rule of thumb to never catch Throwable, but this is an exception. If the intention of "Don't catch Throwable" is "Let the program die", then it is more useful to wrap the Throwable, give it back to the other thread, and throw it again there. 
As it is, the contents of the Throwable are lost in the aether of the dying thread, never to be diagnosed.

Paul says:  it should catch all Throwables, that is a bug. It's up to the consumer of `Task` to rethrow any non-Exception Throwables if that's the behavior they want. 

from discussion:
https://groups.google.com/forum/#!topic/scalaz/D3ZyY1Xu_PQ

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Task.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Task.scala
@@ -306,6 +306,6 @@ object Task {
 
   /** Utility function - evaluate `a` and catch and return any exceptions. */
   def Try[A](a: => A): Throwable \/ A =
-    try \/-(a) catch { case e: Exception => -\/(e) }
+    try \/-(a) catch { case e: Throwable => -\/(e) }
 }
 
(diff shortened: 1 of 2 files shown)


kevinmeredith:
Naive question, is it a good or bad idea to, upon pattern matching on a `java.lang.Throwable` that's not a `scala.util.control.NonFatal`, to automatically shut down the JVM?

Example (taken from this PR):

```scala
 def Try[A](a: => A): Throwable \/ A =
   try \/-(a) catch { 
          case scala.util.control.NonFatal(e) => -\/(e)
          case e                                               => System.exit(-1)
   }
```

Please note that I'm not suggesting this behavior, but asking in order to learn.

Thanks!

PR #680, opened 2014-03-30
Tidying up Foldable.collapse

wemrysi:
This PR removes `Foldable.collapse{2-7}` in favor of clients calling collapse on composed Foldables themselves. It also removes the unnecessary requirement of implicit Foldable evidence from the signature of `Foldable.collapse`.

### DIFF
--- a/core/src/main/scala/scalaz/Foldable.scala
+++ b/core/src/main/scala/scalaz/Foldable.scala
@@ -243,8 +243,8 @@ trait Foldable[F[_]]  { self =>
       }
     })._1
 
-  def collapse[X[_], A](x: F[A])(implicit F: Foldable[F], A: ApplicativePlus[X]): X[A] =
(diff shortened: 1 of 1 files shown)


aloiscochard:
Does anyone remember why this change was made?
There is discussion about reverting it.

wemrysi:
Unfortunately I do not remember why I thought removing collapseN was a good idea, but I don't see any harm in reverting that part (should probably keep the bit that removes the `Foldable` evidence from `collapse` though).

vmarquez:
Is this the kind of thing that should be in syntax rather than the Foldable trait?

tonymorris:
I think collapseN needs to be reinstated. Ideally, we'd compose `Foldable` at a minimal/zero penalty, but that's not how scala works.

Prefer more work by the library creator to ease burden on the user.

PR #817, opened 2014-09-19
Remove useless try-catch block

jedesah:
This check should never be necessary if this class is implemented correctly and it is being used properly (as documented).


aloiscochard:
@pchiusano ping

PR #826, opened 2014-10-06
Avoid catching Fatal exceptions in Task error handling

jedesah:


S11001001:
The current behavior is explained in #608.

PR #829, opened 2014-10-14
Change After method to use ScheduledExecutorService

nlim:
I was trying out the `Task#retry` functionality in Task and I found that it uses a `scalaz.concurrent.Timer` to implement its logic, and it appears `scalaz.concurrent.Timer` uses `Thread.sleep`.

Following the example of `Future#timed`, which uses the `ScheduledExecutorService`, we can avoid the use of `Thread.sleep` in the implementation of `Task#retry`


xuwei-k:
I'm sorry. We have released 7.2.0 final. This change is not binary compatible.

PR #845, opened 2014-11-15
Reduce memory footprint of actors, change strategies to be more efficient & safe for actors and async calculations

plokhotnyuk:
Proposed changes and additions:
1. Reduce 40 bytes per actor instance by replacing `val` by `def` in `toEffect` definition.
2. Improve an executor strategies that baked by FJ pools to be more efficient in submitting tasks for execution.
3. Add `def apply(a: => Unit): Unit` to strategies to be yet more efficient with actors (or other message-passing systems).
4. Improvements of default error handiling in actors: fix `InterruptedException` to be rethown properly, add sending of uncatched errors to `UncaughtExceptionHandler` of thread that by default prints stack trace. 
5. Add creation on demand for default strategy, thread factory and executor services to minimize usage of resources.
6. Use default thread factory that creates deamon threads for the naive strategy.
7. Add `DefaultLIFOForkJoinPool` and `DefaultFIFOForkJoinPool` as more suitable thread pool (for async calculations & actors accordingy) than `DefaultExecutorService` backed by `ThreadPoolExecutor`.
8. Add `Executor(execService: ExecutorService, batchSize: Long): Strategy` to allow fine tuning of trade off between fairness and throughput of actors.
9. Reduce default batch size of actors from 1024 to 10 that allows to improve fairness of predefined strategies.
10. Fix the sequential strategy to run actors without stack overflow.
11. Improve the naive strategy to minimize forking of new threads in message handing loop of actors.
12. Improve tests for actors and strategies.

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Actor.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Actor.scala
@@ -43,30 +43,47 @@ final case class Actor[A](handler: A => Unit, onError: Throwable => Unit = Actor
 
   private def schedule(n: Node[A]): Unit = strategy(act(n))
 
+  private def act(n: Node[A]): Unit = {
+    val b = strategy.batch
+    if (b < 0) nonRecLoop(n)
+    else loop(n, b)
+  }
+
   @annotation.tailrec
-  private def act(n: Node[A], i: Long = strategy.batch, f: A => Unit = handler): Unit = {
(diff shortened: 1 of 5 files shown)


xuwei-k:
I'm sorry. We will release 7.2.0 final soon. This change will conflict to new release and not have binary compatibility. Please resend to `series/7.3.x` branch if you are still interested in this. Thanks.

PR #849, opened 2014-11-19
Add liftEitherT to \/ instances.

noelmarkham:
Implemented as a new EitherTOps class, rather than adding as a function to \/, because doing it there needed to specify all three types (lifted monad, left, right) due to covariant types in \/ and invariant types in EitherT.

Test and compilation check added.

### DIFF
--- a/core/src/main/scala/scalaz/syntax/EitherTOps.scala
+++ b/core/src/main/scala/scalaz/syntax/EitherTOps.scala
@@ -2,8 +2,8 @@ package scalaz
 package syntax
 
 final class EitherTOps[A, B](val self: A \/ B) extends AnyVal {
-  def liftEitherT[M[_]](implicit M0: Monad[M]): EitherT[M, A, B] =
-    EitherT(M0.point(self))
+  def liftEitherT[F[_]](implicit A0: Applicative[F]): EitherT[F, A, B] =
+    EitherT(A0.point(self))
 }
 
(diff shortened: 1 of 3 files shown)


aloiscochard:
I can see this as being useful.

So :+1:, it would be nice to think about what would be a general solution we could apply for all transformers. Or do we want to repeat that pattern?

aloiscochard:
@noelmarkham after thinking about what @puffnfresh mentioned there about implicit shadowing:
https://github.com/scalaz/scalaz/issues/895#issuecomment-73894408

We should find a naming convention that prevent clashing as much as possible, `ToEitherTOps` is not specific enough.

aloiscochard:
`EitherToEitherTOps` should be fine?

noelmarkham:
Thanks @aloiscochard - I've updated my changes. I suppose there's still a wider point here about having this work for other transformers, too.

pbrant:
This is something I'd definitely find useful. If we're happy with the pattern, I'd be happy to add elsewhere. Maybe with a more generic name? liftMT? It's something I've looked for on other monad transformers too.

aloiscochard:
I'm :+1: on this one in favor of #962, as the syntax feel more intuitive and less verbose to me.

xuwei-k:
I'm sorry. We have released 7.2.0 final. This change have conflict and not binary compatible.

PR #850, opened 2014-11-19
Add gather function for Task

jedesah:
Because it is useful and maintains consistency; NonDeterminism also has both a gatherUnordered and a gather. Task should be the same.


aloiscochard:
This PR follow: https://github.com/scalaz/scalaz/pull/848

aloiscochard:
That LGTM, but would be great to have @pchiusano or @runarorama feedback as well.

jedesah:
`gather` is already available for `Task` through `Nondeterminism`. Failed to realize that at the time.

`schedule` would still be nice. But don't have a need for it right now and since it's pretty straightforward to implement, will issue a PR if I ever need it.

jedesah:
Actually, schedule was already added through another PR of mine, I have forgotten...

PR #851, opened 2014-11-21
Moving NonEmptyList to be backed by IList instead of List

vmarquez:
This is a big change, so I'm expecting this one may be controversial.  I've made NEL invariant and had to change the way we are flatMapping so there may be slight performance regression.


vmarquez:
To keep everyone up to date the latest change mitigates the performance degradation of flatMap, it's now comparable to the mutating version.  Also made the small formatting changes that @ceedubs's recommended. 

Wanted to also let everyone know I did do a similar PR https://github.com/scalaz/scalaz/pull/825 (for DList). 

@xuwei-k  Since you reviewed the last PR could you take a look whenever you have time?  I'd appreciate it.  

Thanks everyone for taking time to give feedback.

PR #871, opened 2015-01-02
Fix documentation of ==>>.insertWith{,Key}

fthomas:
The documentation of the `f` parameter in `==>>.insertWith` and `==>>.insertWithKey` has mixed up the `existingValue` and `insertedValue` parameters:
```scala
scala> ==>>.singleton(0, 10).insertWith((existing, inserted) => inserted + 1, 0, 20)
res8: scalaz.==>>[Int,Int] = Bin(0,11,Tip,Tip)

scala> ==>>.singleton(0, 10).insertWith((existing, inserted) => existing + 1, 0, 20)
res9: scalaz.==>>[Int,Int] = Bin(0,21,Tip,Tip)
```
The first is actually the `insertedValue` and the second is actually the `existingValue`. This PR swaps the parameter names in the docstring.


xuwei-k:
:+1:

PR #872, opened 2015-01-05
Free applicative functors

runarorama:


runarorama:
Review by @ekmett would be good

core/src/main/scala/scalaz/Frap.scala
```diff
@@ -0,0 +1,118 @@
+package scalaz
+
+/**
+ * Free applicative functors. Less expressive than free monads, but more
+ * flexible to inspect and interpret.
+ */
+sealed trait Frap[F[_],A] {
```
jedws: `FreeAp`?

core/src/main/scala/scalaz/Frap.scala
```diff
@@ -0,0 +1,118 @@
...
+   * Not tail-recursive unless `G` is a free monad.
+   */
+  def foldMap[G[_]:Applicative](f: F ~> G): G[A] =
+    this match {
+      case Pure(x) => Applicative[G].pure(x)
+      case x@Ap() => Applicative[G].ap(f(x.v()))(x.k() foldMap f)
+    }
+
+  /** Provides access to the first instruction of this program, if present */
+  def para[B](pure: A => B, ap: ({type λ[α] = (F[α], Frap[F, α => A])})#λ ~>
```
jedws: is scalaz going to use kind-projector? would really help…

folone:
Nice! I tried doing [something similar](https://github.com/folone/free.scala/blob/master/src/main/scala/free/applicative/Applicative.scala) some time ago, but failed miserably :)

core/src/main/scala/scalaz/Frap.scala
runarorama: That would be awesome. I think you just volunteered to go through the codebase and replace type lambdas with that.

jdegoes:
:open_mouth: :+1:

core/src/main/scala/scalaz/Frap.scala
jedws: bastard! https://github.com/scalaz/scalaz/pull/875

runarorama: Wait, what? Oh, can KP not support f[x] = (x,x) ?

jedws: I can't find this on GH, replying by email. Yes, KP can do that via the
type function syntax, eg: `λ[α => (α, α)]`

PR #873, opened 2015-01-06
reassociation for 3-way disjunction

tpolecat:
This lets you reassociate `A \/ (B \/ C)` as `(A \/ B) \/ C` and vice-versa. I don't know how useful this is, but I ran into it twice recently so here you go.

### DIFF
--- a/core/src/main/scala/scalaz/Either.scala
+++ b/core/src/main/scala/scalaz/Either.scala
@@ -316,6 +316,26 @@ sealed abstract class \/[+A, +B] extends Product with Serializable {
       b => \&/.That(b)
     )
 
+  /** Reassociate a 3-way disjunction to the left; A \/ (B \/ C) => (A \/ B) \/ C. */
+  def reassocL[BB, CC](implicit ev: B <~< (BB \/ CC)): (A \/ BB) \/ CC =
+    this match {
+      case -\/(a) => -\/(-\/(a))
+      case \/-(b) => ev(b) match {
+        case -\/(bb) => -\/(\/-(bb))
+        case \/-(cc) => \/-(cc)
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


puffnfresh:
Seems like it could be a useful type-class:

```scala
trait Associative[F[_, _]] {
  def reassociateLeft[A, B, C](f: F[A, F[B, C]]): F[F[A, B], C]
  def reassociateRight[A, B, C](f: F[F[A, B], C]): F[A, F[B, C]]
}
```

What do you think?

puffnfresh:
First laws would of course be:

```
reassociateLeft . reassociateRight = id
reassociateRight . reassociateLeft = id
```

mpilquist:
+1 for a new typeclass.

aloiscochard:
I can definitely see it can be useful, but just out of curiosity I was trying to workout which types would have instances for that one?

puffnfresh:
@aloiscochard I can name two:

1. `Either`
2. `\/`

Let me think of some others.

puffnfresh:
Oh, `Tuple2`!

aloiscochard:
yay, that is nice :)

On 8 January 2015 at 15:14, Brian McKenna <notifications@github.com> wrote:

> Oh, Tuple2!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/scalaz/scalaz/pull/873#issuecomment-69183831>.
>



-- 
*Λ\ois*
http://twitter.com/aloiscochard
http://github.com/aloiscochard

puffnfresh:
`Validation` but not `\&/` (think it would need a Semigroup)

puffnfresh:
Some initial work: https://github.com/puffnfresh/scalaz/compare/feature/associative-type-class

tpolecat:
Oh awesome. This is **exactly** what scalaz is about. That solution is so much better. I'm closing this and can backfill the `Associative[\/]` or @puffnfresh can do it, doesn't matter.

PR #874, opened 2015-01-06
example of free with coyoneda

tpolecat:
I have had this gist out there for a while and people seem to understand it, so here it is as a usage example.


stew:
:+1:

PR #876, opened 2015-01-09
Fixed a minor race between the two actors in the message loss test

krishnabhargav:
I believe there is a race condition between actor2 and actor1 when the ping pong happens. The countdown should happen before you ping a message to the actor otherwise, you can potentially encounter a hang.

PS: This is my first 1) contribution to an OSS project ever, 2) first pull request ever. so please ignore my mistakes if any.


runarorama:
Thanks for this, @krishnabhargav

PR #878, opened 2015-01-16
optimize IList#sorted, sortBy

xuwei-k:


aloiscochard:
Hi Kenji,

Have you done some microbenchmark?

I highly recommend jmh, it would be nice to have some metrics to compare.

Especially on small arrays, I'm curious if it's really faster in such
cases?

Thanks!

PR #881, opened 2015-01-23
Fix a cast exception

bryce-anderson:
A generic param A was being forced into a String => Trampoline[Unit].

This shouldn't be encountered under normal use, and I found it when using ```Task.async``` incorrectly. Nevertheless, it is always nice to remove a cast.


xuwei-k:
:+1:

PR #885, opened 2015-01-28
Add listT to ListT

justjoheinz:
Inline with the behaviour of OptionT and MaybeT

### DIFF
--- a/core/src/main/scala/scalaz/ListT.scala
+++ b/core/src/main/scala/scalaz/ListT.scala
@@ -102,13 +102,20 @@ sealed abstract class ListTInstances extends ListTInstances1 {
     new ListTHoist {}
 }
 
-object ListT extends ListTInstances {
+trait ListTFunctions {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/scalaz/ListT.scala
```diff
@@ -102,7 +102,13 @@ sealed abstract class ListTInstances extends ListTInstances1 {
     new ListTHoist {}
 }
 
-object ListT extends ListTInstances {
+trait ListTFunctions {
+  def listT[M[_]] = new (λ[α => M[List[α]]] ~> ListT[M, ?]) {
```
jedws: very, very minor nit, but could the type sig be added, and the impl put on the next line?

justjoheinz:
Sorry, I messed up the commit history somewhat. Will try to fix that soon.

xuwei-k:
:+1:

PR #887, opened 2015-01-30
override ==>>#equals and ==>>#hashCode

xuwei-k:


aloiscochard:
:+1: nice catch

core/src/main/scala/scalaz/Map.scala
```diff
@@ -857,6 +857,17 @@ sealed abstract class ==>>[A, B] {
...
 
+  override final def equals(other: Any): Boolean =
+    other match {
+      case that: ==>>[A, B] =>
+        ==>>.mapEqual[A, B](Equal.equalA, Equal.equalA).equal(this, that)
+      case _ =>
+        false
+    }
+
+  override final def hashCode: Int =
```
jedws: could/should this be a lazy val?

PR #890, opened 2015-01-31
Capture typeclass

tpolecat:
This is mostly to start a discusssion.

So, we have a variety of IO-like types in the wild (including non-scalaz types like `remotely.Response` and @aloiscochard's [matterhorn](https://github.com/aloiscochard/matterhorn)) that provide an effect-capturing unit as well as an unsafe extract. Because there is currently no way to abstract over these operations we end up with overspecialized code, such as this example from `scalaz.stream.io`:

```scala
  def fillBuffer[A](buf: collection.mutable.Buffer[A]): Sink[Task,A] =
    channel((a: A) => Task.delay { buf += a })
```

Using `Capture` we can abstract over the context:

```scala
  def fillBuffer[F[_], A](buf: collection.mutable.Buffer[A])(implicit F: Capture[F]): Sink[F,A] =
    channel((a: A) => F.capture { buf += a })
```
 
I have used a similar abstraction in my own work with good results.

Question: What is the law of `Capture`? It would need to be stated in terms of when a side-effect happens, which is kind of outside the language we usually use to state laws. (See also the "law" of `Catchable`.)

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Task.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Task.scala
@@ -4,6 +4,7 @@ import java.util.concurrent.{ScheduledExecutorService, ConcurrentLinkedQueue, Ex
 import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
 
 import scalaz.{Catchable, Maybe, MonadError, Nondeterminism, Reducer, Traverse, \/, -\/, \/-}
+import scalaz.effect.Capture
 import scalaz.syntax.monad._
 import scalaz.std.list._
 import scalaz.Free.Trampoline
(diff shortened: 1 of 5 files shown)


aloiscochard:
I can see this as being useful, could we have an identity law at least: 
`x => unsafeRun(capture(x)) == x` ?

`IO` don't seems to have much encoded laws either unless I'm missing something.

tpolecat:
@puffnfresh yeah the idea is to abstract over the FFI that constructs primitive effects for a particular IO implementation. It's really just a recognition that there are many such types and no sign that they're going away. If `IO` were the one true and holy data type for this kind of thing it wouldn't be an issue. 

The version of this I'm using for doobie doesn't have `unsafeRun` so maybe it's not needed. I added it here because it would make it possible to at least write a test that verifies an implementation does what is expected.

Maybe @runarorama or @pchiusano could comment since this problem looms large over streams.

jdegoes:
Why isn't this `LiftIO`?

```scala
 def fillBuffer[F[_], A](buf: collection.mutable.Buffer[A])(implicit F: LiftIO[F]): Sink[F,A] =
    channel((a: A) => F.liftIO(IO { buf += a }))
```

One can argue that `IO` in Scala can and does have lots of competing implementations. I guess that's a problem and one should be able to abstract over those implementations, which may end up looking something like `Capture`.

tpolecat:
Sure, you could just write a bogus `LiftIO[Task]` that called `.unsafePerformIO` and that would work. I guess it seems less distasteful to me to abstract over effect capture. Dunno.

aloiscochard:
After more thought, I'm :-1: on this.

I think we should instead try to create an equivalent of `IOSpec`, seems to me that it would be the best way to have different "IO implementation" with a single interface.

If there is a good `IO` I don't see the point of anything else nor any typeclass around it, if one have to abstract other effects in his library, it seems reasonable to do it specialized for now and aim for a unified `IO` interface in the future.

PR #896, opened 2015-02-08
more Foldable short circuiting

xuwei-k:
There are same issues mostly `Foldable` instances https://github.com/scalaz/scalaz/pull/796


puffnfresh:
:+1:

PR #897, opened 2015-02-10
mproduct combinator in Bind

wedens:
`fproduct` like combinator on a Monad. If somebody can suggest better name, I'm open to suggestions :)

### DIFF
--- a/core/src/main/scala/scalaz/Bind.scala
+++ b/core/src/main/scala/scalaz/Bind.scala
@@ -36,11 +36,15 @@ trait Bind[F[_]] extends Apply[F] { self =>
     lazy val f = ifFalse
     bind(value)(if(_) t else f)
   }
-    
+
   /**
    * Repeats a monadic action infinitely
    */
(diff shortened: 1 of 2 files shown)


xuwei-k:
Could you move to `Bind` from `Monad`? because this method does not requires `Monad#point`.  Furthermore could you add `BindOps#mproduct` ?

wedens:
@xuwei-k done

aloiscochard:
:+1: 

I like that name and I don't see a better one, but if someone else have suggestions...

(Nitpick: The commits could be squashed)

wedens:
@aloiscochard policies on rebasing PRs varies from project to project :)

aloiscochard:
@wedens Sorry if I make it sound like I was trying to enforce a policy!

I tried to emphasis it was really just a remark by using parenthesis and "nitpick", which basically mean "Just ignore me if you want to" ;)

xuwei-k:
:+1:

PR #903, opened 2015-02-17
adding liftF syntactic sugar to help lift to the Free monad over the Ind...

vmarquez:
adding liftF syntactic sugar to help lift to the Free monad over the IndexedReaderWriterStateT functor.  This is useful for stack safety, similar PR was accepted here:

https://github.com/scalaz/scalaz/pull/777


xuwei-k:
This change has conflicts and I think this does not need because there is [`Free.liftFU`](https://github.com/scalaz/scalaz/blob/v7.2.0-RC1/core/src/main/scala/scalaz/Free.scala#L29-L31)

PR #905, opened 2015-02-18
new combinator in BindSyntax. bind with ignored result of function

wedens:
I often use this to do something in monad context without sensible output value. e.g. with State or IO.

### DIFF
--- a/core/src/main/scala/scalaz/syntax/BindSyntax.scala
+++ b/core/src/main/scala/scalaz/syntax/BindSyntax.scala
@@ -18,6 +18,8 @@ final class BindOps[F[_],A] private[syntax](val self: F[A])(implicit val F: Bind
 
   def >>[B](b: => F[B]): F[B] = F.bind(self)(_ => b)
 
+  def >>![B](f: A => F[B]): F[A] = F.bind(self)(a => F.map(f(a))(_ => a))
+
   def ifM[B](ifTrue: => F[B], ifFalse: => F[B])(implicit ev: A === Boolean): F[B] = {
     val value: F[Boolean] = ev.subst(self)
     F.ifM(value, ifTrue, ifFalse)


japgolly:
Is that not called `tap`?

japgolly:
In `Id` there is `<|` aka `unsafeTap`  which takes an `A => Any`. This new operator wouldn't be considered unsafe in `IO` but would in many other contexts (right?) so I don't know about `unsafeTap` vs `tap`. It should probably renamed to be more consistent with `<|`.

wedens:
@japgolly It's completely different than `unsafeTap`. It's more like `>>` but right hand result is ignored instead of left hand. So name is consistent with `>>`.

japgolly:
unsafeTap = `Id[A] → (A → Id[B]) → Id[A]` = `A → (A → Any) → A`
This pull req = `F[A] → (A → F[B]) → F[A]`

It's the same thing :)

wedens:
yes. If you write it in this more abstract way, they are the same. I was planning to name it `<<`. I don't remember why I changed it :)

tonymorris:
This appears to be `<*` which can be written using `Apply`.

puffnfresh:
Yeah, this definitely looks like `<*`.

wedens:
no. `<*` is `Applicative f => f a -> f b -> f a` and this is `Monad f => (a -> f b) f a`. actually it should be ``Monad f => (a -> f ()) f a`. simplifies writing `fa >>= (a => f(a) >| a)` where `f` is executed just for the effect

tpolecat:
Yeah it's analogous to tap. You can say `readLn >>! putLn` and get the string back. Seems ok to me.

puffnfresh:
:+1:

tonymorris:
That'll teach me for relying more on English ("more like >> but right hand result is ignored instead of left hand") than the type, which I overlooked.

I agree this is not (<*).

> :t \self f -> self >>= \a -> fmap (const a) (f a)
Monad m => m a -> (a -> m b) -> m a

aloiscochard:
:+1:

S11001001:
:+1:

PR #906, opened 2015-02-24
Enable Partial Lenses with Seq

lbruand-poctu:
Hi!

I am looking for the ability to use partial lenses with `Seq` collections. I have not found anything that would be suitable for that (implicits and the like). So I propose to add the attached methods.

Cheers,
Lucas

### DIFF
--- a/core/src/main/scala/scalaz/PLens.scala
+++ b/core/src/main/scala/scalaz/PLens.scala
@@ -395,6 +395,26 @@ trait PLensFunctions extends PLensInstances with PLensFamilyFunctions {
   def lazyRightPLens[A, B]: LazyEither[A, B] @?> B =
     plens(_.fold(_ => None, b => Some(Store(lazyRight(_), b))))
 
+  def seqHeadPLens[A]: Seq[A] @?> A =
+    plens {
+      case Nil => None
+      case h :: t => Some(Store(_ :: t, h))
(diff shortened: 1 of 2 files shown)


puffnfresh:
Not making a comment on whether this is useful or not but I recommend avoiding `Seq` in any case:

https://orrsella.com/2014/02/27/the-default-scala-seq-is-mutable-wait-what/

tpolecat:
The head/tail matching only works for runtime class `List`.

```scala
scala> seqHeadPLens[Int].get(Vector(1, 2)) 
scala.MatchError: Vector(1, 2) (of class scala.collection.immutable.Vector)
	at scalaz.PLensFunctions$$anonfun$seqHeadPLens$1.apply(PLens.scala:399)
	at scalaz.PLensFunctions$$anonfun$seqHeadPLens$1.apply(PLens.scala:399)
	at scalaz.PLensFunctions$$anon$3.run(PLens.scala:309)
	at scalaz.PLensFamily.get(PLens.scala:44)
```

Assuming you turn the head/tail matching into something that works for all `Seq` types and isn't horribly expensive for things like `Vector`, we still have the mutability concern @puffnfresh raises.

NightRa:
@lbruand-poctu You should look into Monocle Traversals for this.

PR #907, opened 2015-02-25
Performance improvement for Free/Trampoline (~10%)

vmarquez:
Part of what makes Free slow is the expansion of Gosubs in the resume method.  This change drastically reduces the number of necessary resume calls made(which happens when folding a Free structure or calling go on a trampoline) and results in about a 8-10% performance improvement when testing with both trampolines and generic free structures.

https://github.com/vmarquez/scalaz/blob/free_jmhtesting/bench/src/main/scala/scalaz/FreeBench.scala 

Is a branch with jmh testing if anyone wants to replicate/tune.  (thanks to @julien-truffaut for the jmh templates) I've also created additional tests with caliper and found (on average) identical results.  After testing various numbers, 500 appeared to be the drop off point for any measurable improvement, though even much smaller sizes did result in measurable improvements.

### DIFF
--- a/core/src/main/scala/scalaz/Free.scala
+++ b/core/src/main/scala/scalaz/Free.scala
@@ -10,15 +10,12 @@ import std.tuple._
 //      "java.lang.Error: typeConstructor inapplicable for <none>"
 object Free extends FreeInstances with FreeFunctions {
 
-  var resumeCtr = 0
-
   /** Return from the computation with the given value. */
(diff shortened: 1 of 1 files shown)


aloiscochard:
Great stuff! :+1: 

Thanks for effort in doing all the benchmarks and finding the drop off point!

japgolly:
:+1:

PR #913, opened 2015-03-16
Added a parallel applicative instance for scalaz.concurrent.Future

etorreborre:


xuwei-k:
`futureParallelApplicativeInstance` is inaccessible (inside of `Nondeterminism[Future]`)

etorreborre:
Woops, thanks!

aloiscochard:
:+1: 

I even wish this was the default one, that's what I do in $work codebases.

jedws:
@aloiscochard this instance is not a Monad, for the same reason `Validation` does not have a Monad.

@etorreborre I am concerned that this instance exists for the `Future` type, I think it would be better to have a new type that this exists for, eg. `Future @@ Parallel`.

summoning @runarorama @tonymorris for opinions.

runarorama:
I agree with @jedws that this should be a newtype since you will inevitably want to embed monadic futures in applicative ones and vice versa, and you will want to be explicit about that. Exactly like `\/` and `Validation`.

etorreborre:
I added a Tag for the Applicative instance.

puffnfresh:
:+1:

aloiscochard:
@jedws oh right sorry, I'm fine with tag (I you would prefer being explicit when I want my future to be sequential - even if loosing the monad - but I understand the impact). I think a function `parTraverse` could be pretty useful but not sure where/if that would fit?

PR #914, opened 2015-03-17
fix license url for pomExtra

pocketberserker:
Current URL link (http://opensource.org/licenses/bsd-license.php) move "The BSD 2-Clause License" page.

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -130,7 +130,7 @@ object build extends Build {
         <licenses>
           <license>
             <name>BSD-style</name>
-            <url>http://www.opensource.org/licenses/bsd-license.php</url>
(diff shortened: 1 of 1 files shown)


jedws:
this does change the license from the 2 clause to the 3 clause BSD license, which seems fine to me as the extra clause reads:

    3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.

I think this needs a couple of +1s though… anyone?

relrod:
The 3-clause is what is in `LICENSE`, and I am fairly certain that what is in `LICENSE` is the only thing that matters. (The license at the page linked to could, in theory, change sometime in the future, but `LICENSE` won't change in a release once the release is put out). So +1 here.

PR #918, opened 2015-03-24
Add EitherT factory method that infers the outer type constructor

lmmworkv:
First PR, hope I've followed the correct conventions. Apologies if this is duplicating existing functionality; if so I wasn't able to find it.

Add an `EitherT` factory method that infers the outer type constructor. This makes it easier to use `EitherT` with monad types that have multiple parameters e.g. `State` - currently one has to use explicit type parameters (including a type lambda) when calling `EitherT.eitherT` (or `new EitherT`) with a `State`.

The use of `Functor` in the first `Unapply` is somewhat arbitrary/pragmatic, with the rationale that it's a very weak type and an `EitherT` for an `F[_]` that doesn't even have a `Functor` instance is somewhat useless. Very open to any better alternative here, or a way to avoid needing to specify a typeclass at all.

The `Unapply2` is unused but necessary to guide type inference. Again very open to a better way of doing this.

Thanks,
Michael

### DIFF
--- a/core/src/main/scala/scalaz/EitherT.scala
+++ b/core/src/main/scala/scalaz/EitherT.scala
@@ -304,6 +304,10 @@ sealed abstract class EitherTInstances extends EitherTInstances0 {
 trait EitherTFunctions {
   def eitherT[F[_], A, B](a: F[A \/ B]): EitherT[F, A, B] = EitherT[F, A, B](a)
 
+  def eitherTU[FAB, AB, A0, B0](fab: FAB)(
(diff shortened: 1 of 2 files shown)


xuwei-k:
Could you squash the commits?

lmmworkv:
Squashed

xuwei-k:
:+1:

aloiscochard:
:+1: 

Nitpick: The compilation tests (both yours and the pre-existing one)  could be defined in a function instead of being "just" a block of code in the initialization of the object.

That way, the code would not be executed at all runtime during the initialization of the object (which is the aim of such piece of code, iiuc, just to check if it compile).

lmmworkv:
Shall I move them then? Do you want that in a separate commit or squashed into this one?

Issue #911, opened 2015-03-04
Heap is not stack safe


julien-truffaut:
```scala
val as = (1 to 10000).toList
val q = as.foldLeft(Heap.Empty[Int])(_.+(_))
q.toStream.toList ==== as.sorted
```

causes

```
null
java.lang.StackOverflowError
	at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:196)
	at scalaz.TreeFunctions$$anon$3.subForest$lzycompute(Tree.scala:153)
	at scalaz.TreeFunctions$$anon$3.subForest(Tree.scala:153)
	at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:196)
	at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:196)
	at scalaz.TreeFunctions$$anon$3.subForest$lzycompute(Tree.scala:153)
	at scalaz.TreeFunctions$$anon$3.subForest(Tree.scala:153)
```

xuwei-k:
duplicate? https://github.com/scalaz/scalaz/issues/240

julien-truffaut:
@xuwei-k it seems so, closing it

Issue #240, opened 2013-01-03
Heap#toList StackOverflow


xuwei-k:
```
$ screpl org.scalaz%%scalaz-core%7.0.0-M7
[info] Set current project to default-cc9485 (in build file:/Users/user/.conscript/boot/ivy-console/)
Welcome to Scala version 2.9.2 (Java HotSpot(TM) 64-Bit Server VM, Java 1.7.0_09).
Type in expressions to have them evaluated.
Type :help for more information.

scala> import scalaz._,Scalaz._
import scalaz._
import Scalaz._

scala> Heap.fromData(1 to 10000 toList).toList
java.lang.StackOverflowError
        at scalaz.TreeFunctions$$anon$3.<init>(Tree.scala:138)
        at scalaz.TreeFunctions$class.node(Tree.scala:138)
        at scalaz.Tree$.node(Tree.scala:102)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
        at scalaz.TreeFunctions$$anon$3.subForest(Tree.scala:140)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
        at scalaz.TreeFunctions$$anon$3.subForest(Tree.scala:140)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
        at scalaz.TreeFunctions$$anon$3.subForest(Tree.scala:140)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
        at scalaz.TreeFunctions$$anon$3.subForest(Tree.scala:140)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
        at scalaz.Heap$$anonfun$insertWith$2$$anonfun$apply$12.apply(Heap.scala:181)
[...]

PR #920, opened 2015-03-26
Heap is now stack safe (and doesn't rely on stream)

vmarquez:
https://github.com/scalaz/scalaz/issues/911

https://github.com/scalaz/scalaz/issues/240

Both show problems with Heap.  One part of the problem is Tree itself is unsafe for many operations.  By moving to Cofree[IList,A] in place of Tree we're able to prevent SOs, because of it being trampolined.

### DIFF
--- a/core/src/main/scala/scalaz/Heap.scala
+++ b/core/src/main/scala/scalaz/Heap.scala
@@ -1,7 +1,7 @@
 package scalaz
 
 import std.tuple._
-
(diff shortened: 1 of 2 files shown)


xuwei-k:
I'm sorry. We will release 7.2.0 final soon. This change will conflict to new release and not have binary compatibility. Please resend to `series/7.3.x` branch if you are still interested in this. Thanks.

PR #922, opened 2015-03-30
Add withFilter to OptionT

eamelink:
To get rid of annoying compiler warning when making for-comprehensions. `EitherT` also aliases `filter` to `withFilter`.

### DIFF
--- a/core/src/main/scala/scalaz/OptionT.scala
+++ b/core/src/main/scala/scalaz/OptionT.scala
@@ -54,6 +54,9 @@ final case class OptionT[F[_], A](run: F[Option[A]]) {
 
   def filter(f: A => Boolean)(implicit F: Functor[F]): OptionT[F, A] = OptionT(F.map(self.run) { _ filter f })
 
+  /** Alias for filter */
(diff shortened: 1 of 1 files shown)


puffnfresh:
The warning is very useful, I prefer getting it and disabling compilation by turning warnings into errors. We've discussed before in https://github.com/scalaz/scalaz/pull/728

alias1:
Hit this myself today. Read through the previous discussion, which has points for and against that I don't have the required background knowledge to argue. So, a different take: if `withFilter` is a bad idea/won't be implemented/etc, then what is the suggested 'best case' workaround for using filters in for comprehensions with OptionT?

Contrived example:
```
for {
  one <- OptionT(Future(Option(1)))
  if one == 1
} yield { one }
```

jedws:
-1 I think we've come to the general conclusion that this is a scala wart and we shouldn't accomodate it. See #728 for examples on how this translates to a loss of type-safety.

puffnfresh:
@alias1 sorry I didn't see this, but here's a late answer:

```scala
for {
  one <- OptionT(Future(Some(1))).filter(_ == 1)
} yield one
```

Or:

```scala
for {
  one <- OptionT(Future(Some(1)))
  _ <- (one == 1).guard(())
} yield one
```

PR #923, opened 2015-03-30
Fix inconsistent LazyOptionT.orElse.

epost:
The `orElse` methods on `OptionT`, `MaybeT`, `EitherT` and `LazyEitherT` all take transformers as their argument, so let's bring LazyOptionT in line with those.


tonymorris:
:+1:

Issue #917, opened 2015-03-22
recover and recoverWith for \/


mhzajac:
It would be nice to have `recover` and `recoverWith` like methods for `\/` to be able to recover from certain error types with a `PartialFunction[A, B]` (similar to `Try` and `Future`). For example:

```
sealed trait Err
case object HorribleError extends Err
case object NotSoBadError extends Err

def somethingImportant[A](a: A): \/[Err, A] = ???

somethingImportant(param) recover { case NotSoBadError =>
   /* I can handle a `NotSoBadError` and still find an `A` to make things right. */
}
```

Something with signatures similar to:
```
def recover[BB >: B](pf: PartialFunction[A, BB]): (A \/ BB)
def recoverWith[AA >: A, BB >: B](pf: PartialFunction[AA, AA \/ BB]): (AA \/ BB)
```

PR #925, opened 2015-04-07
Add Disjunction#recover and Disjunction#recoverWith

lebek:
As suggested in #917

### DIFF
--- a/tests/src/test/scala/scalaz/DisjunctionTest.scala
+++ b/tests/src/test/scala/scalaz/DisjunctionTest.scala
@@ -32,4 +32,41 @@ object DisjunctionTest extends SpecLite {
     \/.fromTryCatchThrowable[Int, Foo](throw bar) must_=== \/.left(bar)
     \/.fromTryCatchThrowable[Int, Bar](throw foo).mustThrowA[Foo]
   }
+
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/scalaz/DisjunctionTest.scala
```diff
@@ -32,4 +32,41 @@ object DisjunctionTest extends SpecLite {
...
+  "recover" in {
+    sealed trait Foo
+    final class Bar extends Foo
+    final class Baz extends Foo
+    val bar = new Bar
+
+    implicit val equalFoo = Equal.equalA[Foo]
+    implicit val showFoo = Show.showA[Foo]
+
+    -\/[Foo](bar).recover({ case _: Bar => 1 }) must_=== \/-(1)
```
jedws: even though this is just a test, making Bar and Baz case objects would mean we don't have ugly type casing.

I'd hope we can eliminate type casing entirely, perhaps using wart-remover or the like.

    sealed trait Foo
    case object Bar extends Foo
    case object Baz extends Foo

    -\/[Foo](Bar).recover({ case Bar => 1 }) must_=== \/-(1)

PR #928, opened 2015-04-21
Include all modules for getting scalaz

martijnhoekstra:
Paper cut I encountered when getting started: "why doesn't import scalaz.concurrent._ work?"


S11001001:
I'd rather this part of the readme stay core-only.  The link is fine, and the longer dep list can go under #modularity.

PR #930, opened 2015-05-01
Added a findMapM method for Foldables

etorreborre:
We found that this method was useful in our code base when trying to find out an element matching a given criterion after being mapped to another value. We originally provided it on `List` but I thought that this could be useful for any `Foldable` given an interim transformation to an `EphemeralStream`.


tonymorris:
Looks good. I wonder if it is worth also having same that uses `OptionT`.

Issue #828, opened 2014-10-13
getGenericSuperclass fails on a type parametrized with Validation (MalformedParameterizedTypeException).


dpolshin:
hello,
i'm having issue with reflection failing on types parametrized with scalaz.Validation.
here's a code snippet that fails on 7.1.0 but works fine on 7.0.6.

build.sbt:
```scala
scalaVersion := "2.11.2"

sbtVersion := "0.13.6"

libraryDependencies ++= Seq(
  "org.scalaz" %% "scalaz-core" % "7.1.0"
)
```

Test.scala:
```scala
import _root_.scalaz.Validation

abstract class AbstractClass[T <: AnyRef]

class SomeClass extends AbstractClass[Validation[List[String], Object]]

object Test extends App {
  classOf[SomeClass].getGenericSuperclass
}
```

results in:
```
> run
[info] Running Test
[error] (run-main-6) java.lang.reflect.MalformedParameterizedTypeException
java.lang.reflect.MalformedParameterizedTypeException
        at sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl.validateConstructorArguments(ParameterizedTypeImpl.java:60)
        at sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl.<init>(ParameterizedTypeImpl.java:53)
        at sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl.make(ParameterizedTypeImpl.java:95)
        at sun.reflect.generics.factory.CoreReflectionFactory.makeParameterizedType(CoreReflectionFactory.java:105)
        at sun.reflect.generics.visitor.Reifier.visitClassTypeSignature(Reifier.java:140)
        at sun.reflect.generics.tree.ClassTypeSignature.accept(ClassTypeSignature.java:49)
        at sun.reflect.generics.visitor.Reifier.reifyTypeArguments(Reifier.java:68)
        at sun.reflect.generics.visitor.Reifier.visitClassTypeSignature(Reifier.java:138)
        at sun.reflect.generics.tree.ClassTypeSignature.accept(ClassTypeSignature.java:49)
        at sun.reflect.generics.repository.ClassRepository.getSuperclass(ClassRepository.java:84)
        at java.lang.Class.getGenericSuperclass(Class.java:696)
        at Test$.delayedEndpoint$Test$1(Test.scala:9)
        at Test$delayedInit$body.apply(Test.scala:8)
[...]

Issue #571, opened 2013-10-23
Scala type erasure fails for Category


shawjef3:
https://issues.scala-lang.org/browse/SI-7932

This affects at least the following.
scalaz 7.1.0-M3 scalaz/std/AnyValInstances$$anon$2.class method category

scalaz 7.0.{3,4} scalaz/std/AnyValInstances$$anon14.class method category

PR #932, opened 2015-05-11
no generic signatures for scala 2.10.x, see SI-7932, #571 and #828

Milyardo:
`-Yno-generic-signatures` is unnecessary for scala 2.11+ and breaks Java based reflection. Apply `-Yno-generic-signatures` only to versions of scala that need it.


S11001001:
Thanks for this.

PR #934, opened 2015-05-11
Some consistency changes in FreeAp

markus1189:
Noticed some style inconsistencies:

- use `ap(x.v(),x.k())` instead of  `ap(x.v() -> x.k())` (tuple syntax instead of `->`)
- in `monadic` replace the two type lambdas with the nicer kind-projector version

### DIFF
--- a/core/src/main/scala/scalaz/FreeAp.scala
+++ b/core/src/main/scala/scalaz/FreeAp.scala
@@ -64,7 +64,7 @@ sealed abstract class FreeAp[F[_],A] {
    * Embeds this program in the free monad on `F`.
    */
   def monadic(implicit F: Functor[F]): Free[F,A] =
-    foldMap[({type λ[α] = Free[F,α]})#λ](new (F ~> ({type λ[α] = Free[F,α]})#λ) {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/scalaz/FreeAp.scala
```diff
@@ -22,7 +22,7 @@ sealed abstract class FreeAp[F[_],A] {
   def para[B](pure: A => B, ap: λ[α => (F[α], FreeAp[F, α => A])] ~> λ[α => B]): B =
     this match {
       case Pure(x) => pure(x)
-      case x@Ap() => ap(x.v() -> x.k())
+      case x@Ap() => ap(x.v(), x.k())
```
jedws: this uses Scala's very dodgy auto-tupling, to do do this properly there needs to be a second set of parens around the tuple:

    ap((x.v(), x.k()))

The original version explicitly passed a tuple in, the new version is ambiguous whether it is passing a tuple parameter or two arguments.

https://issues.scala-lang.org/browse/SI-3583

markus1189:
Oops okay, I'l revert the change and leave the `->` :)

xuwei-k:
:+1:

Issue #936, opened 2015-05-19
Free's Monad[Free[S, ?]] instance requires a Functor[S], but flatMap doesn't, make them consistent


jedws:
the Monad instance for `Free[S, ?]` is defined as requiring a `Functor[S]`, but `flatMap` is not defined with the same constraint, as it uses `gosub` to achieve the same functionality. This means that for compilation, the `Functor[S]` constraint is unused.

    implicit def freeMonad[S[_]:Functor]: Monad[Free[S, ?]]

But, Free as a construction would usually use the Functor to implement this, for instance https://hackage.haskell.org/package/free-4.12.1/docs/src/Control-Monad-Free.html#line-188

There are good cases for both adding the constraint even though it is unused, and for removing it even though it is correct in theoretical terms.

runarorama:
We should have `Free` be the free monad generated by a type construction, not the free monad generated by a functor. These two are isomorphic via the yoneda lemma, but the former goes with the grain of Scala. At the moment `Free` is sort of a half-breed between them and we should fix it.

S11001001:
cf. #937

jedws:
@runarorama I think you said +1 to #937 but I'm not entirely sure :-)

runarorama:
@jedws This is my current thinking on this: https://github.com/scalaz/scalaz/pull/938

jedws:
So, @runarorama's #938 entirely subsumes this issue. Assuming that gets merged this is addressed.

PR #937, opened 2015-05-19
#936 remove unnecessary `Functor[S]` constraint from `Monad[Free[S, ?]]` supplier

jedws:
https://github.com/scalaz/scalaz/issues/936 

As @S11001001 pointed out, http://hackage.haskell.org/package/free-4.12.1/docs/src/Control-Monad-Free-Church.html#line-112 shows that the Functor constraint isn't necessary, and is merely an implementation detail


S11001001:
:+1:

runarorama:
Well... https://github.com/scalaz/scalaz/pull/938

jedws:
#938 is a much more comprehensive response to this… once it is merged this should be closed and abandoned.

PR #938, opened 2015-05-19
Simplify the representation of the Free monad.

runarorama:
Review by @pchiusano, @vmarquez, @divarvel, and/or @aloiscochard appreciated.


runarorama:
I am a little displeased that `x` and `x.flatMap(pure)` don't have the same in-memory representation or the same performance characteristics. A user cannot observe the difference other than with timing, but it seems less than ideal.

example/src/main/scala/scalaz/example/FreeCoyoUsage.scala
```diff
@@ -1,95 +0,0 @@
-package scalaz.example
```
tpolecat: Would be nice to see this reworked rather than deleted. People seemed to find it helpful.

jedws:
lgtm

core/src/main/scala/scalaz/Free.scala
```diff
@@ -41,12 +34,19 @@ object Free extends FreeInstances with FreeFunctions {
...
   type Sink[A, B] = Free[(=> A) => ?, B]
 
-  /** A free monad over the free functor generated by `S` */
-  type FreeC[S[_], A] = Free[Coyoneda[S, ?], A]
+  /** Suspends a value within a functor in a single step. Monadic unit for a higher-order monad. */
+  def liftF[S[_], A](value: S[A]): Free[S, A] =
+    Suspend(value)
+
+  /** Return the given value in the free monad. */
+  def point[S[_], A](value: => A): Free[S, A] = Return[S, A](value)
```
japgolly: Value is lazy here but eager in `Return`.

aloiscochard:
That is great!

Less is more, I think it makes thing much easier to understand to anyone diving in the code now that there is just a single implementation.

Plus, I was never very comfortable with that arbitrary `500` limit :-)

About the sub-optimal performance concerns of `x.flatMap(pure)`, well it's Scala... personally, I think the trade-off is worth it due to the big abstraction benefit we can see here.

LGTM

vmarquez:
Looks good.

PR #941, opened 2015-05-23
Optimize performance by avoiding lazy val

Atry:


xuwei-k:
Could you explain your intention? I disagree with this changes. I think these value should be _call by need_

Atry:
There is no reason to cache the `ifTrue` and `isFalse` parts. They usually don't evaluate again.

Atry:
The cache behavior yields more terrible case for `while`, https://github.com/scalaz/scalaz/issues/940

S11001001:
They do evaluate again often for List et al.

Atry:
@S11001001 No, they didn't.

```
scala> Monad[List].ifM(List(true, false, true, false), Monad[List].pure { println("ifTrue") }, Monad[List].pure { println("ifFalse") })
ifTrue
ifFalse
res3: List[Unit] = List((), (), (), ())
```

S11001001:
It shouldn't.

aloiscochard:
:-1: that is not the behavior I would expect.

PR #943, opened 2015-05-28
Remove unneccessary covariance requirement

greenrd:
This was overlooked when StateT was changed to no longer require a type
parameter with a covariant parameter.


xuwei-k:
:+1:

Issue #921, opened 2015-03-27
ListT violate the associative law


xuwei-k:
due to https://github.com/scalaz/scalaz/pull/166

- https://wiki.haskell.org/ListT_done_right
- https://wiki.haskell.org/ListT_done_right_alternative

> 3.4 Order of ListT []
This is a simple example that doesn't use IO, only pure ListT [].
v :: Int -> ListT [] Int
v 0 = ListT [[0, 1]]
v 1 = ListT [[0], [1]]
main = do
    print $ runListT $ ((v >=> v) >=> v) 0
    -- = [[0,1,0,0,1],[0,1,1,0,1],[0,1,0,0],[0,1,0,1],[0,1,1,0],[0,1,1,1]]
    print $ runListT $ (v >=> (v >=> v)) 0
    -- = [[0,1,0,0,1],[0,1,0,0],[0,1,0,1],[0,1,1,0,1],[0,1,1,0],[0,1,1,1]]

```scala
Welcome to Scala version 2.11.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_25).
Type in expressions to have them evaluated.
Type :help for more information.

scala> scalaz.BuildInfo.version
res0: String = 7.1.1

scala> import scalaz._, Scalaz._
import scalaz._
import Scalaz._

scala> val v = Kleisli[({type l[a] = ListT[List, a]})#l, Int, Int]{
     |   case 0 => ListT(List(List(0, 1)))
     |   case 1 => ListT(List(List(0), List(1)))
     | }
v: scalaz.Kleisli[[a]scalaz.ListT[[+A]List[A],a],Int,Int] = Kleisli(<function1>)

scala> ((v >=> v) >=> v) run 0
res1: scalaz.ListT[[+A]List[A],Int] = ListT(List(List(0, 1, 0, 0, 1), List(0, 1, 1, 0, 1), List(0, 1, 0, 0), List(0, 1, 0, 1), List(0, 1, 1, 0), List(0, 1, 1, 1)))

scala> (v >=> (v >=> v)) run 0
res2: scalaz.ListT[[+A]List[A],Int] = ListT(List(List(0, 1, 0, 0, 1), List(0, 1, 0, 0), List(0, 1, 0, 1), List(0, 1, 1, 0, 1), List(0, 1, 1, 0), List(0, 1, 1, 1)))

scala> res1 === res2
res3: Boolean = false
```

S11001001:
[...]

PR #947, opened 2015-06-02
Add support for find on ListT

justjoheinz:
Implement find for ListT. I wonder why it is not included, if there are reasons, please feel free to just ignore this PR.


xuwei-k:
I think there is no reason. BTW https://github.com/scalaz/scalaz/issues/921

PR #948, opened 2015-06-03
Mark all *Syntax values as transient lazy val to allow serializable instances.

pawel-n:
I need to be able to serialize my instances when using scalaz together with Apache Spark. However even when I mixin Serializable, e.g. "new Functor[T] with Serializable" the serialization fails because the field functorSyntax is not serializable. This patch marks all such fields as "@transient lazy val" to exclude them from serialization (they will be automatically reinitialized after deserialization).

### DIFF
--- a/core/src/main/scala/scalaz/Align.scala
+++ b/core/src/main/scala/scalaz/Align.scala
@@ -59,7 +59,7 @@ trait Align[F[_]] extends Functor[F] { self =>
   def alignLaw = new AlignLaw {}
 
   ////
-  val alignSyntax = new scalaz.syntax.AlignSyntax[F] { def F = Align.this }
+  @transient lazy val alignSyntax = new scalaz.syntax.AlignSyntax[F] { def F = Align.this }
 }
 
 object Align {
--- a/core/src/main/scala/scalaz/Applicative.scala
+++ b/core/src/main/scala/scalaz/Applicative.scala
@@ -110,7 +110,7 @@ trait Applicative[F[_]] extends Apply[F] { self =>
   def applicativeLaw = new ApplicativeLaw {}
 
   ////
-  val applicativeSyntax = new scalaz.syntax.ApplicativeSyntax[F] { def F = Applicative.this }
+  @transient lazy val applicativeSyntax = new scalaz.syntax.ApplicativeSyntax[F] { def F = Applicative.this }
 }
 
 object Applicative {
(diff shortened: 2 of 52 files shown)


jedws:
why do you want to be able to serialise typeclasses? There is no state in these things.

pawel-n:
Generally consider remotely calling a polymorphic function, e.g. given
```scala
def sort[A](xs: Seq[A])(implicit ord: Ordering[A])
```
if you want to call sort across JVM boundaries you have to serialize the arguments, including the ordering.

In my particular case I want to broadcast a value that contains typeclass instances to an Apache Spark cluster. An example class looks like that:
```scala
final case class FunctorMap[F[_], U[_], V[_]] (
  m: Model[U, V]
) (implicit
  f: Functor[F]
) extends Model[({type T[A] = F[U[A]]})#T, ({type T[A] = F[V[A]]})#T] {
  override type Type[A] = m.Type[A]

  override implicit val space = m.space

  override def apply[A](inst: Type[A])(input: F[U[A]])(implicit a: Analytic[A]): F[V[A]] =
    f.map(input)(m(inst)(_))
}
```
Here Model[A, B] is the hom-set (hom-type?) of a category of Scala types as objects and "functions with some parameter vector, differentiable over this parameter vector and polymorphic over the number field" as morphisms. FunctorMap lifts a scalaz functor and applies it to some model.

Theoretically if we serialize just the type F it should be possible to somehow reconstruct the instance Functor[F], assuming instance uniqueness, coherence, etc. I can imagine given a class and a type I could use some mad reflection to scan for all instances of a given class and then use a constraint solver to find the desired one - but I'm not insane enough to do that. 

Ideally I'd love if the instances provided by scalaz were serializable, but so far I'm ok with writing my own. I'm also not an expert on Scala so there might be some other solution I'm not aware of.

puffnfresh:
@pawel-n constraints on data structures is generally considered a bad idea. The implicit Functor should be a requirement on the methods instead. In your case I'd add an implicit Functor to `apply`.

I'm very happy to discuss why the position of the implicits matters.

pawel-n:
The problem is: apply is defined in the Model trait as
```scala
def apply[A](inst: Type[A])(input: In[A])(implicit a: Analytic[A]): Out[A]
```
so I cannot put additional constraints on it.

pawel-n:
The real issue is though that - no matter where the implicit is - I have to get both the apply function and the function instance on the second machine in order to call it. There is no way of getting around that unless we somehow conjure it up from nothing.

Issue #498, opened 2013-08-29
StateT should be trampolined


runarorama:
State and StateT are basically totally useless without Trampoline unless your JVM stack size is bigger than the number of steps in your State program.

Related: Issue #497

PR #949, opened 2015-06-05
Rework StateT to allow for trampolining

vmarquez:
A few folks have discovered that while it intuitively seems like creation of a StateT[Trampoline, A, B] would result in stack safety, it actually can still SOE on recursive flatMap calls. 
John De Goes pointed out that this could be avoided by refactoring StateT’s application function to be F[A => F[(A,B)]] rather than just A => F[(A,B)].  I implemented a toy version of this for my Lambda-Conf talk on stack safety in Scala and showed the way John imagined State does allow for stack safety.  It’s more or less: 

```scala
   case class StateT[F[_], S, A](sf: F[S => F[(S,A)]]) {
    def flatMap[B](f: A => StateT[F, S, B])(implicit M: Monad[F]): StateT[F, S, B] = 
      StateT[F, S, B](((s1: S) => { 
        sf.flatMap(sfa => {          //the key, important line here that f application happens inside the F[_]s bind and not before hand
            sfa(s1).flatMap(t =>  
                f(t._2).sf.flatMap(z => z(s1)) 
          )
        }) 
      }).point[F])
  }
```

Though ‘applying’ it gets to be a bit weird, hence why I now have a monad instance in the StateT trait. 

Another negative to this is it requires a lot more monads…everywhere.  Permeates through Hoist most obviously, and also some other method calls

We could make the argument that if we want to get fancy with state
then you can just use the free monad fold trick I’ve proposed, but the current StateT[Trampoline,A,B] *does* fail against the test case I provided (though it may need to be a bigger list).  Kleisli and RWST also
suffer from the same problem.  If this is a change we’re OK with I’d be happy to fix it on those classes too.   Feedback/review from @runarorama @tpolecat  @S11001001 and any others would be greatly appreciated.

### DIFF
--- a/core/src/main/scala/scalaz/StateT.scala
+++ b/core/src/main/scala/scalaz/StateT.scala
@@ -3,102 +3,109 @@ package scalaz
 import Id._
 
 trait IndexedStateT[F[_], -S1, S2, A] { self =>
-  /** Run and return the final value and state in the context of `F` */
-  def apply(initial: S1): F[(S2, A)]
+  
+  def getF[S <: S1]: (Monad[F] => F[S => F[(S2, A)]])
+ 
+  def apply(initial: S1)(implicit F: Monad[F]): F[(S2, A)] = 
+    F.join(F.map[S1 => F[(S2, A)], F[(S2, A)]](getF(F))(sf => sf(initial))) 
 
   /** An alias for `apply` */
-  def run(initial: S1): F[(S2, A)] = apply(initial)
+  def run(initial: S1)(implicit F: Monad[F]): F[(S2, A)] = apply(initial)
 
   /** Calls `run` using `Monoid[S].zero` as the initial state */
-  def runZero[S <: S1](implicit S: Monoid[S]): F[(S2, A)] =
+  def runZero[S <: S1](implicit S: Monoid[S], F: Monad[F]): F[(S2, A)] =
     run(S.zero)
 
   /** Run, discard the final state, and return the final value in the context of `F` */
-  def eval(initial: S1)(implicit F: Functor[F]): F[A] =
-    F.map(apply(initial))(_._2)
+  def eval(initial: S1)(implicit F: Monad[F]): F[A] =
+    F.bind[S1 => F[(S2, A)], A](getF(F))(sf => F.map(sf(initial))(_._2))
 
   /** Calls `eval` using `Monoid[S].zero` as the initial state */
-  def evalZero[S <: S1](implicit F: Functor[F], S: Monoid[S]): F[A] =
+  def evalZero[S <: S1](implicit F: Monad[F], S: Monoid[S]): F[A] =
     eval(S.zero)
 
   /** Run, discard the final value, and return the final state in the context of `F` */
-  def exec(initial: S1)(implicit F: Functor[F]): F[S2] =
-    F.map(apply(initial))(_._1)
+  def exec(initial: S1)(implicit F: Monad[F]): F[S2] =
+    F.map(run(initial))(_._1)
 
   /** Calls `exec` using `Monoid[S].zero` as the initial state */
-  def execZero[S <: S1](implicit F: Functor[F], S: Monoid[S]): F[S2] =
+  def execZero[S <: S1](implicit F: Monad[F], S: Monoid[S]): F[S2] =
     exec(S.zero)
 
-  def map[B](f: A => B)(implicit F: Functor[F]): IndexedStateT[F, S1, S2, B] = IndexedStateT(s => F.map(apply(s)) {
-    case (s1, a) => (s1, f(a))
-  })
-
-  def xmap[X1, X2](f: S2 => X1)(g: X2 => S1)(implicit F: Functor[F]): IndexedStateT[F, X2, X1, A] = IndexedStateT(s => F.map(apply(g(s))) {
-    case (s1, a) => (f(s1), a)
-  })
-
+  def map[B](f: A => B)(implicit F: Functor[F]): IndexedStateT[F, S1, S2, B] = 
+    mapsf((sf: (S1 => F[(S2, A)])) => (s: S1) => F.map(sf(s))(t => (t._1, f(t._2))))
+   
+  def xmap[X1, X2](f: S2 => X1)(g: X2 => S1)(implicit F: Functor[F]): IndexedStateT[F, X2, X1, A] = IndexedStateT.createState(
+    (m: Monad[F]) => (x: X2) => F.map(self(g(x))(m))(t => (f(t._1), t._2))
+  )
+    
   /** Map both the return value and final state using the given function. */
-  def mapK[G[_], B, S](f: F[(S2, A)] => G[(S, B)]): IndexedStateT[G, S1, S, B] =
-    IndexedStateT { s => f(apply(s)) }
+  def mapK[G[_], B, S](f: F[(S2, A)] => G[(S, B)])(implicit M: Monad[F]): IndexedStateT[G, S1, S, B] = IndexedStateT.createState(
+    (m: Monad[G]) => (s: S1) => f(apply(s)(M))
+  ) 
 
   import BijectionT._
   def bmap[X, S >: S2 <: S1](b: Bijection[S, X])(implicit F: Functor[F]): StateT[F, X, A] =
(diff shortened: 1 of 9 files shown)


xuwei-k:
Could you squash the commits?

japgolly:
If you squash the commits you may as well also remove the no-op changes to `PLens` & `VectorTest` :)

vmarquez:
@xuwei-k @japgolly squashed and reverted the noop commits, sorry for that!

vmarquez:
Addresses #498 too

bshelden:
+1.  I ran into this not too long ago.
I have a chunk of code working under RWST as my way of deferring side effects.  The problem is, I ran into a case where I couldn't avoid needing to traverse a large vector of incoming data.
I had originally used RWST since I wanted to be able to switch out for Trampoline, but was later shown that this leads to a stack overflow.

I solve this in a couple ways.  First, I just reverted to non-monadic style and dealt with the plumbing directly.  That worked well enough for the simple transformations, but I quickly realized it would lead to more pain and bugs than I wanted to eat, so I ended up dropping RWST and just writing my own monad backed by Trampoline.  Effective, but I've now re-written RWST myself for my own case.

core/src/main/scala/scalaz/MonadTrans.scala
```diff
@@ -18,7 +18,7 @@ object MonadTrans {
 }
 
 trait Hoist[F[_[_], _]] extends MonadTrans[F] {
-  def hoist[M[_]: Monad, N[_]](f: M ~> N): F[M, ?] ~> F[N, ?]
+  def hoist[M[_]: Monad, N[_]](f: M ~> N)(implicit G: Monad[N]): F[M, ?] ~> F[N, ?]
```
S11001001: I have to think about it some more, but a tentative :-1: to changing `Hoist`.

core/src/main/scala/scalaz/package.scala
```diff
@@ -138,18 +138,18 @@ package object scalaz {
...
-    def apply[S1, S2, A](f: S1 => (S2, A)): IndexedState[S1, S2, A] = new IndexedState[S1, S2, A] {
-      def apply(s: S1) = f(s)
-    }
+    def apply[S1, S2, A](f: S1 => (S2, A)): IndexedState[S1, S2, A] = IndexedStateT[Id, S1, S2, A](f)
   }
+
   object State extends StateFunctions {
     def apply[S, A](f: S => (S, A)): State[S, A] = new StateT[Id, S, A] {
-      def apply(s: S) = f(s)
+      implicit override def G: Monad[Id] = Monad[Id]
```
S11001001: I think referring to `G` here will crash.  Can you add a test to see?

vmarquez: Not really sure what you mean about it crashing, I've added another monad law test, can you let me know if you want another test?  G can be called/used from the StateT[Id] created with the apply...

Thanks for looking at this.

core/src/main/scala/scalaz/MonadTrans.scala
vmarquez: I originally didn't want to change it, but of the alternatives I do think this is the cleanest solution.   I don't agree we should leave StateT as is (being susceptible to stack overflows w\ a trampoline or Free).  Changing hoist doesn't just allow us to fix StateT, it would also allow us to do the same thing to Kleisli and RWST, which also suffer the same issue.  So I felt the good weighed the bad in this scenario.

xuwei-k: > I have to think about it some more, but a tentative :-1: to changing Hoist.

I agree with S11001001.

vmarquez:
@S11001001  @xuwei-k I've removed the changes to hoist along with the requirement to pass around the monad instance for the F.

xuwei-k:
Looks like this needs to be rebased.

vmarquez:
@xuwei-k  rebased and squashed.  If you think this looks OK, let me know if you want me to merge and fix the two conflicts.

vmarquez:
ping @xuwei-k @S11001001 if this is good I'll fix the conflicts, I'm a little worried if it sits for too long it'll become difficult to merge.  Let me know if you have more feedback.

xuwei-k:
I don't want to review until you rebase and fix the conflicts.

core/src/main/scala/scalaz/PLens.scala
```diff
@@ -116,7 +116,8 @@ sealed abstract class PLensFamily[A1, A2, B1, B2] {
     State(s => (s, get(s)))
 
   def %=[A >: A2 <: A1](f: B1 => B2): PState[A, B2] =
-    State(a => run(a) match {
+    State(a => 
+      run(a) match {
```
xuwei-k: unnecessary changes

vmarquez:
@xuwei-k @japgolly @S11001001 I've taken care of the squashing and merge issues, should be clean for review now.

xuwei-k:
:+1:

core/src/main/scala/scalaz/StateT.scala
```diff
@@ -1,104 +1,110 @@
 package scalaz
 
 import Id._
+import scalaz.syntax.monad._
```
tpolecat: The contributor guide asks that you not use syntax in the implementation.
> Do not use `scalaz.syntax._` to implement `scalaz._`. Instead use the type classes directly.

vmarquez: Ok.  I will go ahead and change it, then squash.

tpolecat:
:+1: pending syntax thing above

vmarquez:
@tpolecat took out the syntax import and went to the typeclass methods directly, thanks for the feedback 
cc @xuwei-k

puffnfresh:
The new constraints are slightly worry. Do we lose anything important?

vmarquez:
@puffnfresh well there are certainly things that are applicative but not Monads, however, I think the number of things you could do with the as an F inside a StateT fall into the realm of not interesting.  Counter examples welcome and I/we can try to see if we can work around it.

core/src/main/scala/scalaz/StateT.scala
```diff
@@ -3,102 +3,109 @@ package scalaz
 import Id._
 
 trait IndexedStateT[F[_], -S1, S2, A] { self =>
-  /** Run and return the final value and state in the context of `F` */
-  def apply(initial: S1): F[(S2, A)]
+  
+  def getF[S <: S1]: (Monad[F] => F[S => F[(S2, A)]])
+ 
+  def apply(initial: S1)(implicit F: Monad[F]): F[(S2, A)] =
```
puffnfresh: I think this can be Bind instead of Monad. Am I right?

puffnfresh:
@vmarquez Applicatives compose so you shouldn't really need to use StateT with just an Applicative. Can we relax from Monad to Bind at least?

vmarquez:
@puffnfresh hmm there are a few places we might be able to 'delay' a monad constraint, but eventually we need a monad I think.

https://github.com/scalaz/scalaz/pull/949/files#diff-feaa5593d003806fe7b23a70927a91c7R108

runarorama:
:+1:

xuwei-k:
merged https://github.com/scalaz/scalaz/commit/a07dc366363a9b3ac311b35abd6a2eb1029b9c99

TomasMikula:
Related: #1180: Alternative `StateT` representation that admits stack-safe operations for tail-recursive (`BindRec`) monads.

vmarquez:
I prefer not having trampoline baked in because I often don't need stack safety, when I do I can use a stack safe F

TomasMikula:
That is generally my sentiment, too. However, the current implementation seems to create enough function objects that it might not be faster than explicit trampolining. Only benchmarks would show, though.

vmarquez:
The other downside is if you put a Task inside the StateT (which is very often in my use cases), you're paying for twice as many thunks.

TomasMikula:
Is there a difference between one huge function composed of many small functions (like in the current implementation or in the straightforward implementation of `StateT`) and representing those compositions more explicitly (my PR)? I don't think there really are more objects allocated.

TomasMikula:
@vmarquez see [benchmark results](https://github.com/scalaz/scalaz/pull/1180#issuecomment-223852333).

PR #952, opened 2015-06-10
IList groupBy uses NonEmptyList instead of IList

adelbertc:


adelbertc:
API-breaking change, not really a drop-in for `scala.collection.immutable.List` by using `NonEmptyList` as opposed to `IList` - could keep the old one and have a `groupBy1` as well.

vmarquez:
My two cents are that as a library user,  I'm fine with breaking changes that can be fixed with a single method call (NEL.list -> IList)

tpolecat:
Looks good to me, only question is whether we want a deprecation cycle with the old one.

tpolecat:
Any reason this isn't on `Foldable`?

PR #954, opened 2015-06-12
fix binary incompatibility with recent scalacheck

refried:
Gen.value was deprecated in Oct 2013 and deleted in May 2014;  newer versions of scalacheck can slip into the build due to transitive dependencies, even if users choose the older version that 7.1.2 passes under.

Assuming this change doesn't break the build, I hope you will release this as 7.1.3, soonish.


xuwei-k:
:+1:

xuwei-k:
Thanks. I'll release 7.1.3 within a few days.

PR #959, opened 2015-06-24
Add method Order#sort(x, y).

TomasMikula:


gpampara:
:-1:

tonymorris:
@gpampara why?

TomasMikula:
The motivation for this method is to reduce the use of `if`/`else` in the user code, because it is prone to errors by accidentally swapping the `if` and `else` branches. This is due to the fact that a boolean test (e.g. `Order#lessThan(x, y)`) is "forgetful": the result (a boolean value) bears no connection to the tested values, and thus in order to do anything useful with the result of the test, you have to reach back to the tested values in the `if` and `else` branches, and this is very easy to mismatch.

Therefore, I think it is good practice to try to push `if`/`else` down to the most basic operations (such as the implementation of the proposed `sort` method) and compose higher level code from these basic operations (without additional `if`s if possible).

S11001001:
I like it. :+1:

gpampara:
Upon looking at this again, I'm not sure what I was thinking.... Sorry for the noise! :+1:

PR #962, opened 2015-06-25
EitherT.fromDisjunction

adelbertc:


xuwei-k:
:+1:

vmarquez:
This is similar to https://github.com/scalaz/scalaz/pull/849

I think both of them are good ideas, just thought I'd point it out we may not both 'styles'.

PR #967, opened 2015-07-01
Add ReaderT value alias

adelbertc:


xuwei-k:
:+1:

core/src/main/scala/scalaz/package.scala
```diff
@@ -114,6 +114,7 @@ package object scalaz {
   type |>=|[G[_], F[_]] = MonadPartialOrder[G, F]
 
   type ReaderT[F[_], E, A] = Kleisli[F, E, A]
+  val ReaderT = Kleisli
```
jedws: should add one for Reader as well

adelbertc: Hmm we have this https://github.com/scalaz/scalaz/blob/b55570be26de7051d9ed8904d7effeae34ed52cf/core/src/main/scala/scalaz/package.scala#L124-126

or did you have something else in mind?

jedws: yes, ignore me sorry

PR #978, opened 2015-07-28
Add bi-conditional <-->

cvogt:


relrod:
Looks reasonable to me. Maybe add the unicode variant ⇔ too, since the others have unicode variants?

cvogt:
don't see one for `-->` and I couldn't find tests for the unicode variants for other. I'd suggest to wait until someone asks for it.

PR #980, opened 2015-07-29
Add filter on Task to improve usability in for-comprehension

jedesah:
- Add both filter and withFilter to avoid the compiler issuing an unnecessary warning
- Add test case demonstrating how Task can be used in a for-comprehension with pattern matching thx to the withFilter function. Also asserts that the behavior is as expected.

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Task.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Task.scala
@@ -5,6 +5,7 @@ import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
 
 import scalaz._
 import scalaz.Tags.Parallel
+import scalaz.concurrent.Task.FilterException
 import scalaz.syntax.monad._
 import scalaz.std.list._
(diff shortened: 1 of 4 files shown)


tpolecat:
I'm uncomfortable with this. `Task` has no sensible empty value. Using a separate binding for destructuring is safe, unlike this.

jedesah:
@tpolecat Could you elaborate on why it is unsafe?

> Task has no sensible empty value

Are you referring to the use of `FilterException`. Maybe `MatchException` would be more appropriate?

vmarquez:
@jedesah Here's a link to the discussion where @puffnfresh shows the problem:

https://github.com/scalaz/scalaz/pull/728#issuecomment-45779363

My suggestion was having a scalaz.unsafe that adds the withFilter to appropriate monads so people can opt-in since it's a common request, but it would also be obvious it potentially loses type safety.

puffnfresh:
`withFilter` is really bad :-1:

jedesah:
Ok, I see this is a controversial subject. I'll pass on this for now.

PR #969, opened 2015-07-08
Add working UnapplyProduct

travisbrown:
This uses a trick for approximating the effect of multiple parameter lists (a trick I learned [from Miles](https://gist.github.com/milessabin/cadd73b7756fe4097ca0), of course) to provide `UnapplyProduct` instances. This means that the following finally works:

```scala
import scalaz._, Scalaz._

(1, 'a).bitraverseU(_.set("saw int" :: Nil), _.set("saw sym" :: Nil))
```

I've added a compilation test on the model of the other `Unapply` types, but no tests for syntax.

PR #981, opened 2015-07-30
add bisequenceU that uses the now fixed bitraverseU

rjsvaljean:
Since @travisbrown  fixed `bitraverseU` in https://github.com/scalaz/scalaz/pull/969 , it's useful to have the sequence version using the type inference provided by UnapplyProduct.


xuwei-k:
:+1:

PR #987, opened 2015-08-20
Add freeMonadC

Atry:


tpolecat:
While you're at it you might want to add an unapply for that shape, otherwise the instance ends up being of limited use. I did this in doobie but never got around to making a PR. Code [here](https://github.com/tpolecat/doobie/blob/master/core/src/main/scala/doobie/imports.scala#L127-L148) if you want to copy/paste, assuming it's sensible. @S11001001 can you glance at this?

Atry:
The `freeMonadC` in this PR only supports `FreeC` and `Coyoneda`, other `Functor`s are not supported. Therefore I did not use `Unapply`.

S11001001:
@tpolecat's comment still holds in that case.  You can test by creating a val whose type is a fully expanded version of a `FreeC` on, say, `Maybe`, and try calling `ToMonadOpsUnapply` on it.

jedws:
what is the status of this PR?

Atry:
Since `FreeC` is deleted in 7.2.0, I guess we could close this PR.

PR #988, opened 2015-08-20
Added multiply1 on Semigroup with efficient default implementation

jbgi:
and used it in default implementation of Monoid[F].multiply.

Straight port of times1p from the haskell semigroups package.

### DIFF
--- a/core/src/main/scala/scalaz/Semigroup.scala
+++ b/core/src/main/scala/scalaz/Semigroup.scala
@@ -22,6 +22,26 @@ trait Semigroup[F]  { self =>
   def append(f1: F, f2: => F): F
 
   // derived functions
+
+  /**
+   * For `n = 0`, `value`
+   * For `n = 1`, `append(value, value)`
+   * For `n = 2`, `append(append(value, value), value)`
+   *
+   * The default definition uses peasant multiplication, exploiting associativity to only
+   * require `O(log n)` uses of [[append]]
+   */
+  def multiply1(value: F, n: Int): F = {
+    @scala.annotation.tailrec
+    def go(x: F, y: Int, z: F): F = y match {
+      case y if (y & 1) == 0 => go(append(x, x), y >>> 1, z)
+      case y if (y == 1)     => append(x, z)
+      case _                 => go(append(x, x), (y - 1) >>>  1, append(x, z))
+    }
+    if (n <= 0) value else go(value, n, value)
+  }
+
+
   protected[this] trait SemigroupCompose extends Compose[λ[(α, β) => F]] {
     def compose[A, B, C](f: F, g: F) = append(f, g)
   }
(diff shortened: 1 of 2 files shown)


tonymorris:
:+1:

S11001001:
C'est super cool!

:+1:

core/src/main/scala/scalaz/Semigroup.scala
```diff
@@ -22,6 +22,33 @@ trait Semigroup[F]  { self =>
...
+   * For `n = 1`, `append(value, value)`
+   * For `n = 2`, `append(append(value, value), value)`
+   *
+   * The default definition uses peasant multiplication, exploiting associativity to only
+   * require `O(log n)` uses of [[append]]
+   */
+  def multiply1(value: F, n: Int): F = {
+    @scala.annotation.tailrec
+    def f(x: F, y: Int): F = y match {
+      case y if (y % 2 == 0) => f(append(x, x), y / 2)
```
japgolly: I don't know if Scalac/JVM will do this automatically but it may be faster to change `y % 2` to `y & 1`, and `y / 2` to `y >>> 1`.

jbgi: As this has an impact on readability, I'd like another opinion on those optimizations before I make the changes... someone?

jbgi: I just did some unscientific micro-benchmarking: using bitwise operations indeed make it almost twice as fast on cold start, however once hotspot (jdk8) is warm, there is no significant performance difference. But I guess it is worth it (the optimization may not be triggered in all cases).

xuwei-k:
https://travis-ci.org/scalaz/scalaz/jobs/76625925#L318

```
[error] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/Semigroup.scala:37: overloaded method value & with alternatives:
[error]   (x: Long)Long <and>
[error]   (x: Int)Int <and>
[error]   (x: Char)Int <and>
[error]   (x: Short)Int <and>
[error]   (x: Byte)Int
[error]  cannot be applied to (Boolean)
[error]       case y if (y & 1 == 0) => f(append(x, x), y >>> 1)
[error]                    ^
[error] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/Semigroup.scala:43: overloaded method value & with alternatives:
[error]   (x: Long)Long <and>
[error]   (x: Int)Int <and>
[error]   (x: Char)Int <and>
[error]   (x: Short)Int <and>
[error]   (x: Byte)Int
[error]  cannot be applied to (Boolean)
[error]       case y if (y & 1 == 0) => g(append(x, x), y >>> 1, z)
[error]                    ^
[error] two errors found
```

xuwei-k:
:+1:  
could you squash commits?

PR #992, opened 2015-08-25
Add a simplified BaseMonadError

Atry:


Atry:
I am not quite understand why `F` for `MonadError` takes two type parameter.

jedws:
@Atry we should make MonadError unary-kinded, I have been trying to do so but am coming up against some difficulties with things like `Monad[OptionT[String \/ ?, ?]]` that work in 7.2 branch previously

jedws:
closing this in favour of #1049

PR #995, opened 2015-08-31
remove Functions traits

xuwei-k:
I think there is no reason Functions traits does exists.
We can become easily keep binary compatibility if Functions traits remove.

### DIFF
--- a/core/src/main/scala/scalaz/Adjunction.scala
+++ b/core/src/main/scala/scalaz/Adjunction.scala
@@ -70,10 +70,7 @@ abstract class Adjunction[F[_], G[_]](implicit val F: Functor[F], val G: Functor
   }
 }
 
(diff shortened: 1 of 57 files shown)


aloiscochard:
:+1: agree. To be honest, I never understood what was the point of such split.

Maybe someone else have an idea?

S11001001:
I think the point was so that these could be mixed into "all functions" or "all imports" objects, and things of that like.

If we're not doing that, then I guess there's no point in having them.

xuwei-k:
"all imports" and "all functions" objects(e.g. `scalaz.std.AllInstances`, `scalaz.Scalaz`) still exists.
I have only removed ScalazDataTypeFunctions(e.g. MaybeFunctions, EitherTFunctions)

PR #996, opened 2015-09-03
Add `orZero` and `~` to `OptionT`, `LazyOptionT`, and `MaybeT`

blouerat:
`getOrElse` but with `Monoid.zero` as the default value


xuwei-k:
:+1:

PR #999, opened 2015-09-06
Add isLeft, isMiddle and isRight for Either3

aoiroaoino:


core/src/main/scala/scalaz/Either3.scala
```diff
@@ -25,6 +25,10 @@ sealed abstract class Either3[+A, +B, +C] extends Product with Serializable {
   def leftOr[Z](z: => Z)(f: A => Z)   = fold(f, _ => z, _ => z)
   def middleOr[Z](z: => Z)(f: B => Z) = fold(_ => z, f, _ => z)
   def rightOr[Z](z: => Z)(f: C => Z)  = fold(_ => z, _ => z, f)
+
+  def isLeft:   Boolean = leftOr(false)(_ => true)
+  def isMiddle: Boolean = middleOr(false)(_ => true)
+  def isRight:  Boolean = rightOr(false)(_ => true)
```
xuwei-k: Could you use pattern match instead of xxxOr method?

aoiroaoino: I see ;-)

congrats 7.2 final release! :tada:
I will resend a PullReq to the series/7.3.x branch :smile:

PR #993, opened 2015-08-26
Fixed drawTree stack overflow and memory usage problems

edmundnoble:
Fixes issue 673, `Tree.drawTree` stack overflow for wide trees. After trampolining `drawTree`, it started to throw `OutOfMemoryError` on large trees. I rewrote it to use `Vector[StringBuilder]` to represent the lines of the tree representation instead of `Stream[String]`, and had to represent the lines backwards for efficient prepending. Now, it works for trees up to 6000 members before it runs out of memory on my laptop. I have a local test, but it's a bit slow.

PR #1000, opened 2015-09-07
Add a test for issue #673

chris-martin:
This test is currently failing.


xuwei-k:
I think this is a different issue. #993 fixed `drawTree`. your test does not use `drawTree`.

chris-martin:
Okay, should I open a new issue for `flatten`?

chris-martin:
Replaced by issue #1001 and PR #1002.

Issue #1001, opened 2015-09-07
Tree.flatten stack overflow for wide trees


chris-martin:
This causes a stack overflow:

```scala
Tree.node(0, Stream.from(1).take(100000).map(i => Tree.leaf(i)))
  .flatten.force
```

Seen in 7.1.1 and 7.2.x.

xuwei-k:
workaround (but [another test](https://github.com/scalaz/scalaz/commit/86540ea9d2aede90d7d5f4bbbf4ca635fddd30b9) fail)

```scala
def flattenTree[A](tree: Tree[A]): Stream[A] = {
  def squish(t: Tree[A], xs: => Stream[A]): Stream[A] =
    Stream.cons(t.rootLabel, Foldable[Stream].foldRight(t.subForest, xs)(squish))

  squish(tree, Stream.Empty)
}
```

PR #1002, opened 2015-09-07
Add a test for issue #1001

chris-martin:
This test is currently failing.


jedws:
Is there a fix for this? A failing test without a fix isn't much use as a PR unless accompanied by a fix. Perhaps just add the tests to the original issue?

PR #1003, opened 2015-09-07
Add doc links between Lens/PLens types/aliases

chris-martin:
Trying to improve discoverability a little in the API docs.


S11001001:
:+1:

Issue #997, opened 2015-09-04
Task.gatherUnordered is very slow


orionll:
```scala
val tasks = Vector.tabulate(1024)(Task(_))
val resul = Task.gatherUnordered(simulation).run
println(result)
```

This takes 6 seconds on my computer and 48 seconds for 1500 iterations. The same code with Nondeterminism[Task].gatherUnordered takes milliseconds.

cokleisli:
Some more background on this issue: http://stackoverflow.com/questions/32392147/slow-performance-when-using-scalaz-task

PR #1004, opened 2015-09-08
Fix for #997 (Use Task's gatherUnordered instead of derived method)

vmarquez:
When calling Task.gatherUnordered without exceptionCancels set, Task was then calling Nondeterminism[Task].reduceUnordered, which is the derived method from the parent instance.  This is non optimized for Task, and ended up being much slower than when calling Nondeterminism[Task].gatherUnordered. 

I've changed it to call the same method when you aren't trying to cancel running Tasks, as the behavior is the same but now calling either Task.gatherUnordered has the same performance as Nondeterminism[Task].gatherUnordered.


vmarquez:
fix for #997

vmarquez:
Any chance someone can review so it gets into 7.2?  @xuwei-k?

PR #1009, opened 2015-09-21
enable pattern matching on `TagOf` instances

refried:
This would help me scrap some boilerplate.  Any objections?


puffnfresh:
I like it! :+1: 

I think a Monocle Iso would probably be more useful in general but this is good.

S11001001:
:+1:

PR #1015, opened 2015-10-02
Make tree construction and pattern matching more consistent.

shawjef3:
Right now, you create trees with "def node", "def leaf", or the apply method of Tree. However, you pattern match, but do not create trees with Tree.Node. This is very inconsistent with how Scala construction and pattern matching usually work. Normally you have a capital letter object that you can use for both construction and pattern matching.

Changing Node to fit what I would normally expect from a Scala object is pretty easy; just add apply. Then as an analog of the leaf method, add a Leaf object with apply and unapply, so that pattern matching and construction is more like a normal Scala case class.

Tree.apply is completely weird. It actually is for creating a leaf, not a tree, and there's no corresponding unapply. I think that method should be deprecated or changed to create a node instead of a leaf, but that's another ticket.

Jeff

### DIFF
--- a/core/src/main/scala/scalaz/Tree.scala
+++ b/core/src/main/scala/scalaz/Tree.scala
@@ -116,7 +116,12 @@ object Tree extends TreeInstances with TreeFunctions {
    */
   object Node {
     def apply[A](root: => A, forest: => Stream[Tree[A]]): Tree[A] = {
-      node[A](root, forest)
+      new Tree[A] {
+        lazy val rootLabel = root
+        lazy val subForest = forest
+
+        override def toString = "<tree>"
+      }
     }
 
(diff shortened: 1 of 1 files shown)


S11001001:
:+1:

Also :+1: to deprecating Tree.apply and Tree.unapply and changing all their callers.

jedws:
+1 for the change and @shawjef3 please deprecate the Tree.apply/unapply as well.

xuwei-k:
:+1: 
however `scalaz-seven` is not main branch. 

https://github.com/scalaz/scalaz/commit/824b27797d26ba466edd7430748cc40e
> This branch has been discontinued and is kept for historical purposes. Development continues on the appropriate release branches:

please send to `series/7.2.x`

shawjef3:
Reopened as #1031.

PR #1021, opened 2015-10-14
Narrow the type of orElse on Either

jedesah:


S11001001:
You should be able to use an as-pattern (`@`) to avoid reconstructing the `\/-`.  See map, leftMap, others in same file for examples.

jedesah:
@S11001001 Thx for the suggestion!

S11001001:
:+1:

PR #1022, opened 2015-10-16
A balanced, streaming approach to IList traversal.

S11001001:
I've always thought the "trampoline every applicative" answer was no solution; it's tedious, and we can do better in traversal, because we can build balanced applicative trees instead of applicative linked lists, which is what happens for a basic fold, be it left or right.

I emphasize that I'm not concerned with stack-overflow while traversing (though I also manage to keep that solved while still traversing the list only once), my interest is in _also_ solving the problem of stack consumption when interpreting the resulting applicative.  For example, the basic function applicative.

``` scala
import scalaz._, syntax.applicative._,
       syntax.traverse._, std.function._

object Test {
  trait Connection
  object AConnection extends Connection

  type DB[+A] = Connection => A

  def throwAt1000(n: Int): DB[IList[Unit]] =
    ((IList.fill(n - 1)(().point[DB])) ++
       (((_:Connection) => sys error s"thrown at $n")
          :: IList.fill(1000 - n - 1)(().point[DB]))).sequence[DB, Unit]
}
```

That mostly gives a bunch of repeated frames from the `bind` implementation for `Function1`; this represents the stack consumption while interpreting `Function1` (i.e. calling the resulting action).  So compare old b88002d1c4c1be6ee8437ba487fe861f77aa3b51 to new 2dd1bc12e50da914dd9dc90e4bfab8c52804d08f, with a few arguments to `throwAt1000`.
- 0: old 3, new 31
- 200: old 401, new 26
- 500: old 1001, new 24
- 999: old >1021, new 18

I emphasize again that the stack consumption that is improved is that when the resulting `DB` is called.  The consumption in the traversal phase remains constant.

There are a few approaches to balanced traversal, but this one is interesting because of the minimum amount of intermediate state it requires at any time.  It doesn't create another copy of the list outside the applicative.  In fact, for extra efficiency, the `state` could be a mutable stack without breaking anything, because it isn't used persistently; as the maximum stack height is logarithmic in list length, there would be considerable consing avoidance.

I will post a diagram later that may help explain why `roll` works.  I am also making a PR for `Vector`, showing off one simpler balanced traversal that I think is well-suited to `Vector` but not as good as this one for `IList` or `List`.

If we like this, I'll port it to `List`; it should just be name substitution to implement.

### DIFF
--- a/core/src/main/scala/scalaz/IList.scala
+++ b/core/src/main/scala/scalaz/IList.scala
@@ -560,6 +560,8 @@ sealed abstract class IListInstances extends IListInstance0 {
 
       def traverseImpl[F[_], A, B](fa: IList[A])(f: A => F[B])(implicit F: Applicative[F]): F[IList[B]] = {
         type StackEnt = F[IList[B] => IList[B]]
+        // For a brief explanation of 'roll':
+        // https://github.com/scalaz/scalaz/pull/1022#issuecomment-148694245
         @tailrec
         def roll(count: Int, stack: IList[StackEnt]): IList[StackEnt] =
           if ((count & 1) == 0) {
--- a/core/src/main/scala/scalaz/IList.scala
+++ b/core/src/main/scala/scalaz/IList.scala
@@ -558,8 +558,22 @@ sealed abstract class IListInstances extends IListInstance0 {
       override def cojoin[A](a: IList[A]) =
         a.uncons(empty, (_, t) => a :: cojoin(t))
 
-      def traverseImpl[F[_], A, B](fa: IList[A])(f: A => F[B])(implicit F: Applicative[F]): F[IList[B]] =
-        fa.foldRight(F.point(IList.empty[B]))((a, fbs) => F.apply2(f(a), fbs)(_ :: _))
+      def traverseImpl[F[_], A, B](fa: IList[A])(f: A => F[B])(implicit F: Applicative[F]): F[IList[B]] = {
+        type StackEnt = F[IList[B] => IList[B]]
+        @tailrec
+        def roll(count: Int, stack: IList[StackEnt]): IList[StackEnt] =
+          if ((count & 1) == 0) {
+            val ICons(h, ICons(th, t)) = stack
+            roll(count >> 1, F.apply2(th, h)(_ compose _) :: t)
+          } else stack
+        fa.foldLeft((0, IList[StackEnt]())){
+          case ((count, stack), a) =>
+            (count + 1,
+             roll(count + 1, F.map(f(a))(a => (a :: (_:IList[B]))) :: stack))
+        }._2.foldLeft(F.point(IList[B]())){(acc, se) =>
+          F.apply2(se, acc)(_(_))
+        }
+      }
 
       def unzip[A, B](a: IList[(A, B)]): (IList[A], IList[B]) =
         a.unzip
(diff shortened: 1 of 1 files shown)


S11001001:
@dolio As traversal-efficiency maven, what do you think of this?

S11001001:
#1023 is a (significantly easier to follow) implementation for vector, that relies on random access.  We could use that if people don't like this; I think this approach is nicer for lists though.

The main weakness of this implementation is in tree imbalances.  That will be easier to explain with diagrams, though, in a later comment.  They are "balanced enough" already; they guarantee maximum depth from root to any element through the applicative tree, which is the main goal.

S11001001:
So, say we've iterated over seven elements of a list.  Here's the history of operations (will be clearer in a moment), and the current `stack` variable contents.

ilist-roll-diagram-7

You can read this from right to left; the top of the stack is a single F, the second element is a 2-F ap tree, the third is a 4-F ap tree.

Now, we add the eighth element from the list:

ilist-roll-diagram-8

We push the single F, then "8" triggers 3 ap joins from the top, as labelled in red.

Then at the end we just finish ap-ing whatever's on the stack, so we're left with a full ap tree.

dolio:
I like this idea. I guess it's really only fixing one combinator, but being able to traverse without having to trampoline something as simple as reader is nice.

The only thing that came to mind was that you could use something like skew binary numbering to avoid looping in `roll`, but I'm not sure that would actually improve performance (since we're not doing anything incremental), and might just be more difficult to implement.

S11001001:
12d441c reduces the `bind` depth in the tests stated above by an additional 2 frames.  Not a huge win, I just forgot to use `ap` directly where it was reasonable.

S11001001:
This tree-building approach is also described more fully in section 9.2.1, "Binary Random-Access Lists", of *Purely Functional Data Structures*.  (Section 6.2.1 in the freely available PDF thesis.)

There are some additional structures in that chapter, as @dolio mentions, that may be relevant.

aloiscochard:
Thanks for the detailed explanation, I'm all for this :+1: ... and I'll be thinking about reusing that approach for a while, ty :)

edmundnoble:
Where are we on this? My initial benchmarks suggest this brings a significant (~20%) speed increase (depending on the applicative) to traversal. This strategy is equally valid for stdlib `List`, so I believe that should be added before this is merged.

vmarquez:
@S11001001 @edmundnoble Sorry for taking so long to reply to this. 

So, this makes sense why it'd be faster, but I did some benchmarking with JMH and it didn't appear to be so.   @edmundnoble Can you share your benchmarks?  It's possible that it depends on the Ap operation and just doing something simple shows a slowdown, compared to a more complicated Ap.   I do think this is a good strategy, just want to make sure it works in this scenario.

TomasMikula:
I actually wouldn't expect this to be faster than the current approach, but I think that the main strength to be considered is that the resulting applicative is composed in a balanced fashion.

Also, if I count correctly, this implementation uses _n_ `map`s and _n-1_ `apply2`s on the target applicative. I suspect the `map`s could perhaps be avoided (which may or may not improve performance).

fommil:
this is interesting, did it go anywhere?

emilypi:
@S11001001 should we close this or do you intend to see this through?

emilypi:
Closing for inactivity

PR #1023, opened 2015-10-16
Top-down divide-and-conquer-style traverse for Vector.

S11001001:
(Goals as explained in #1022.)

For `Vector`, random access means we can treat the whole structure as a tree by just pivoting at the middle.  Stack consumption while traversing is logarithmic in vector length, but the implementation uses even less intermediate state, the trees are more balanced, and some obvious microoptimizations (see comment in code) can yield quick constant-factor improvements.

The pivot could be chosen a little more intelligently, e.g. a 4-6 split is more efficient than a 5-5 split when a `case 2` exists, but I have not done this, for simplicity.

I also believe a `DVector`-style approach, with `compose` in the inductive case, as with #1022's rolling, might pay off versus simply appending the vectors.  Again, simplicity has stayed my hand.

### DIFF
--- a/core/src/main/scala/scalaz/std/Vector.scala
+++ b/core/src/main/scala/scalaz/std/Vector.scala
@@ -2,7 +2,7 @@ package scalaz
 package std
 
 import vector._
-import annotation.tailrec
+import annotation.{switch, tailrec}
 
 sealed trait VectorInstances0 {
   implicit def vectorEqual[A](implicit A0: Equal[A]): Equal[Vector[A]] = new VectorEqual[A] {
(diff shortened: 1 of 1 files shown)


S11001001:
There is also a "bottom-up" approach based on repeatedly traversing the Vector, combining adjacent pairs (or triples or quads for optimization) on each iteration, until shrinking the Vector to 1 element.  I think there is more value in improving #1022's "rolling" tree-building approach to be more balanced, though.

aloiscochard:
Not sure why this one is sleeping, but it looks great to me :+1:

S11001001:
If [SI-4442](https://issues.scala-lang.org/browse/SI-4442) is fixed, this should always beat #1265's `Vector` version. As it stands, I'm not sure; it may or may not be more efficient, depending on applicative choice. Certainly only this version fully avoids the need for trampolining; left-folding aps as in #1265 also introduces a dilemma for applicatives that don't perform well left-associated (e.g. `Free`, `Writer` of `List`).

Unlike #1022, this one satisfies @xuwei-k's laziness concerns mentioned in #1176. So, barring SI-4442-related concerns, it's probably the best choice.

/cc @edmundnoble

TomasMikula:
Another possible micro-optimization: Would it make sense to pull the `case 0` out of the loop and change the invariant to `0 <= s < e <= v.size`, i.e. strict inequality in the middle?

Also, since in any Vector of length length >1 there will be at least as many `case 2` as `case 1`, would it make a difference to move `case 2` up?

fommil:
This looks great! Any reason why it wasn't merged?

emilypi:
@S11001001 are you still looking after this PR, or should i close it?

TomasMikula:
Ah, this should have been merged long time ago. 😕

emilypi:
@TomasMikula yeah... 4 years may be a bit of a stretch to get it in now 😄

TomasMikula:
Revived in #2047. Everyone please go give it a 👍

PR #1026, opened 2015-10-17
Specialize various Foldable methods, and traverseS, for OneAnd.

S11001001:
Good `F`s for `OneAnd` tend to have specialized `Foldable` functions; e.g. `Vector`'s `length` or `toVector`.  We can keep the improved performance of those specializations when lifting into `OneAnd` by dispatching to them directly with `OneAnd`'s own specializations.


puffnfresh:
:+1:

xuwei-k:
:+1:

PR #1032, opened 2015-10-29
examples: Use kind-projector lambdas

markus1189:
In the examples, I found some type lambdas that are not yet written using kind-projector sugar, so I replaced them with the nicer sugar variant.


puffnfresh:
:+1:

PR #1033, opened 2015-10-30
Make tree construction and pattern matching more consistent

shawjef3:
Clone of #1015, with the addition of removing uses of leaf and node throughout the rest of the project.


xuwei-k:
What is the difference of #1031 ?

shawjef3:
There isn't any difference in the code.

PR #1035, opened 2015-11-02
FreeT and BindRec

puffnfresh:
BindRec type-class is based upon [Stack Safety for Free](http://functorial.com/stack-safety-for-free/index.pdf).


vmarquez:
Very awesome.  I think it's possible that we could elide the actual FreeT class itself if we add a method to Free that is similar to foldRun (though it would need to  deal with disjunctions and short circuiting so you don't necessarily have to fold the entire structure).  Then we just need a BindRec instance showing that the type is able to be 'folded' in a stack safe way.  LiftF exists meaning we can go from any monad M to a Free[M] rather than sticking it into a 'transformer'.

japgolly:
Very cool.

I noticed the `tailrecM` instances for `Free{,T}` and `Task` are not actually tail-recursive. Should they not be rewritten or removed, or is that actually ok?

puffnfresh:
@japgolly I conjecture that since they're in the Trampoline they're alright.

FreeT does look broken. I copied it from PureScript, which I think must also be broken.

puffnfresh:
@vmarquez I'm not sure how Free could be used as FreeT since the algebra and even the kinds are different:

```scala
FreeT :: (* -> *) -> (* -> *) -> * -> *
Free :: (* -> *) -> * -> *
```

puffnfresh:
@japgolly looking at FreeT again, `resume` is trampolined and so I think it's fine, too.

japgolly:
@puffnfresh Ah right. Ok cheers, that makes sense. It might be an idea to add a comment to those impls... although I see you've got stack-safety tests planned (awesome) so that should cover it.

paf31:
@puffnfresh Looks nice. One other thing you could possibly add is a variant of `run` for `Free` for any `BindRec`.

Also, the `List` instance looks a bit off since `go` doesn't seem to be called, but there is a sketch of a safe implementation of ["`ListT` done right"](https://wiki.haskell.org/ListT_done_right) at the end of the PDF.

Sorry if any of this is already covered elsewhere and I missed it.

jedws:
@paf31 tweaked the List and Vector implementations somewhat so as to keep a list of lists that are to be  appended. Also added IList and DList impl, with DList realising the underlying ILIst and then just using the IList impl under the covers

puffnfresh:
I'm pretty happy with this now. Some questions:

1. Does BindRec need some syntax?
2. Are there some more tests needed for Free, FreeT, Task, IO, etc?
3. Is there anything else we need to clean up?

core/src/main/scala/scalaz/std/Stream.scala
```diff
@@ -72,6 +72,15 @@ trait StreamInstances {
...
           f(\&/.Both(a.head, b.head)) #:: alignWith(f)(a.tail, b.tail)
+
+    def tailrecM[A, B](f: A => Stream[A \/ B])(a: A): Stream[B] = {
+      def go(abs: Stream[A \/ B]): Stream[B] =
+        abs match {
+          case \/-(b) #:: tail => b #:: go(tail)
+          case -\/(a) #:: tail => tailrecM(f)(a) #::: go(tail)
+        }
+      go(f(a))
+    }
```
xuwei-k: `BindRec[Stream]` is not available (should be change  the return type)

puffnfresh:
@xuwei-k both Stream and EphemeralStream have working instances! Verified by substituting Option for them in the FreeT tests :fireworks:

puffnfresh:
I'm answering my own questions:

1. No BindRec does not need syntax, things like FreeT will be the main uses
2. We have some useful tests and I've run example programs like the one below
3. I think we're good
4. I'll rebase our fixes together

```scala
import scalaz.{ Bind, FreeT, Functor }
import scalaz.effect.{ IO, SafeApp }

object Main extends SafeApp {
  sealed trait X[A]
  case class PurePrint[A](s: String, next: A) extends X[A]
  object X {
    implicit val functorX: Functor[X] = new Functor[X] {
      def map[A, B](fa: X[A])(f: A => B): X[B] =
        fa match {
          case PurePrint(s, next) => PurePrint(s, f(next))
        }
    }
  }

  val program: FreeT[X, IO, Unit] =
    Bind[FreeT[X, IO, ?]].forever {
      FreeT.liftF[X, IO, Unit](PurePrint("Hello", ())).flatMap { _ =>
        FreeT.liftM(IO.putStrLn("World from IO"))
      }
    }

  override val runc: IO[Unit] =
    program.runM {
      case PurePrint(s, next) =>
        IO.putStrLn(s ++ " from pure").map(_ => next)
    }
}
```

xuwei-k:
https://travis-ci.org/scalaz/scalaz/jobs/89564552#L440

```
[error] /home/travis/build/scalaz/scalaz/tests/src/test/scala/scalaz/iteratee/EnumeratorTTest.scala:85: ambiguous implicit values:
[error]  both method optionTBindRec in class OptionTInstances0 of type [F[_]](implicit F0: scalaz.Monad[F], implicit B0: scalaz.BindRec[F])scalaz.BindRec[[β]scalaz.OptionT[F,β]]
[error]  and method optionTMonadPlus in class OptionTInstances0 of type [F[_]](implicit F0: scalaz.Monad[F])scalaz.MonadPlus[[β]scalaz.OptionT[F,β]]
[error]  match expected type scalaz.Bind[OIO]
[error]     (consume[Int, OIO, List] &= enum.map(_ * 2)).run.run.unsafePerformIO() must_===(Some(List(2, 4, 6)))
[error]                              ^
```

puffnfresh:
@xuwei-k thanks, saw that straight after I pushed. Fixed!

puffnfresh:
Rebased, tests pass, examples work, I think we're good! :dog:

core/src/main/scala/scalaz/Cokleisli.scala
```diff
@@ -45,6 +45,9 @@ sealed abstract class CokleisliInstances0 {
     new CokleisliProfunctor[F] {
       def F = implicitly
     }
+
+  implicit def cokleisliBindRec[F[_], R]: BindRec[Cokleisli[F, R, ?]] =
+    new CokleisliBindRec[F, R] {}
 }
```
xuwei-k: `cokleisliBindRec[F[_], R]: BindRec[Cokleisli[F, R, ?]]` and `cokleisliMonad[F[_], R]: Monad[Cokleisli[F, R, ?]]` are same constraints.

should we change `cokleisliInstance[F[_], R]: BindRec[Cokleisli[F, R, ?]] with Monad[Cokleisli[F, R, ?]] =` ?

puffnfresh: @xuwei-k yes, pushed an amendment!

core/src/main/scala/scalaz/NonEmptyList.scala
```diff
@@ -131,7 +131,7 @@ object NonEmptyList extends NonEmptyListInstances {
   def apply[A](h: A, t: A*): NonEmptyList[A] =
     nels(h, t: _*)
 
-  def unapplySeq[A](v: NonEmptyList[A]): Option[(A, IList[A])] =
+  def unapply[A](v: NonEmptyList[A]): Option[(A, IList[A])] =
```
xuwei-k: What's the reason for this changes?

jedws: `unapplySeq` requires that the second argument be a `Seq`, which `IList` is not. The old one was totally unusable. I had a choice between changing to unapply and using the `List` form for pattern matching. This is cheaper and actually usable.

xuwei-k: I got it :+1:

tests/src/test/scala/scalaz/std/ListTest.scala
```diff
@@ -159,4 +160,22 @@ object ListTest extends SpecLite {
   }
 
   checkAll(FoldableTests.anyAndAllLazy[List])
+
+
+  object instances {
+    def equal[A: Equal] = Equal[List[A]]
+    def order[A: Order] = Order[List[A]]
+    def semigroup[A: Semigroup] = Monoid[List[A]]
```
xuwei-k: `def monoid[A]  = Monoid[List[A]]` ?

https://github.com/scalaz/scalaz/blob/240946b90/core/src/main/scala/scalaz/std/List.scala#L106

tests/src/test/scala/scalaz/std/StreamTest.scala
```diff
@@ -96,4 +97,18 @@ object StreamTest extends SpecLite {
   "filter" ! forAll {
     (xs: Stream[Int], p: Int => Boolean) => MonadPlus[Stream].filter(xs)(p) must_=== xs.filter(p)
   }
+
+  object instances {
+    def equal[A: Equal] = Equal[Stream[A]]
+    def order[A: Order] = Order[Stream[A]]
+    def semigroup[A: Semigroup] = Monoid[Stream[A]]
```
xuwei-k: `def monoid[A] = Monoid[Stream[A]]`

tests/src/test/scala/scalaz/IListTest.scala
```diff
@@ -427,4 +428,16 @@ object IListTest extends SpecLite {
 
   checkAll(FoldableTests.anyAndAllLazy[IList])
 
+  object instances {
+    def equal[A: Equal] = Equal[IList[A]]
+    def order[A: Order] = Order[IList[A]]
+    def monoid[A: Semigroup] = Monoid[IList[A]]
```
xuwei-k: `def monoid[A] = Monoid[IList[A]]`

tests/src/test/scala/scalaz/DListTest.scala
```diff
@@ -30,4 +32,13 @@ object DListTest extends SpecLite {
     
   }
 
+  object instances {
+    def equal[A: Equal] = Equal[DList[A]]
+    def monoid[A: Semigroup] = Monoid[DList[A]]
```
xuwei-k: `def monoid[A] = Monoid[DList[A]]`

xuwei-k:
I appreciate your awesome works! LGTM except some trivial comments :+1:

puffnfresh:
@xuwei-k fixed those!

PR #1036, opened 2015-11-04
Changed keySet from Set to ISet in ==>>

joshcough:
This is a bit more in the FP spirit.

I updated all the tests, and they all pass.

### DIFF
--- a/core/src/main/scala/scalaz/Map.scala
+++ b/core/src/main/scala/scalaz/Map.scala
@@ -220,8 +220,8 @@ sealed abstract class ==>>[A, B] {
   def keys: List[A] =
     foldrWithKey(List.empty[A])((x, _, xs) => x :: xs)
 
-  def keySet =
-    Set(keys: _*)
+  def keySet(implicit o: Order[A]): ISet[A] =
+    ISet.fromList(keys)
 
   def toList =
     toAscList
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/Map.scala
```diff
@@ -220,8 +220,8 @@ sealed abstract class ==>>[A, B] {
   def keys: List[A] =
     foldrWithKey(List.empty[A])((x, _, xs) => x :: xs)
 
-  def keySet =
-    Set(keys: _*)
+  def keySet(implicit o: Order[A]): ISet[A] =
+    ISet.fromList(keys)
```
xuwei-k: Maybe `foldrWithKey(ISet.empty[A])((x, _, xs) => xs insert x)` is more efficient

puffnfresh: @xuwei-k :+1:

S11001001:
This can be implemented without the `Order` constraint.  You can use `ISet`'s `Tip` and `Bin` data constructors, and match on the same of `==>>` to produce a matching set structure.  That way, there is no need to compare keys or rebalance the tree as it is constructed.

joshcough:
Thanks for the comments. I will make the changes on Saturday.

joshcough:
@S11001001 done. Thanks for the suggestion.

core/src/main/scala/scalaz/Map.scala
joshcough: @xuwei-k i went with @S11001001 suggestion, since  it eliminated the `Order` constraint.

S11001001:
:+1:

PR #1039, opened 2015-11-06
Fix EitherT.orElse

### LINKED ISSUE #1037: EitherT orElse broken [7.1.4]
Here is a testcase:

```scala
import scalaz._, Scalaz._
type Program[A] = EitherT[scalaz.effect.IO, Unit, A]
def prints(s: String) = EitherT ( effect.IO(println(s).right[Unit]) )
```
then:
```
scala> prints("heyo").run.unsafePerformIO
heyo
res3: scalaz.\/[Unit,Unit] = \/-(())
```

However...

```
scala> (prints("heyo") orElse prints("erm")).run.unsafePerformIO
heyo
heyo
res10: scalaz.\/[Unit,Unit] = \/-(())
```

And the implementation:

```scala
/** Return this if it is a right, otherwise, return the given value. Alias for `|||` */
  def orElse(x: => EitherT[F, A, B])(implicit F: Bind[F]): EitherT[F, A, B] = {
    val g = run
    EitherT(F.bind(g) {
      case -\/(_) => x.run
      case \/-(_) => g
    })
  }
```

The result from the bind, if successful, should be wrapped as a constant `\/-`, but instead the code inside the bind returns `this`, essentially, having the same effect as e.g. `prints("heyo").replicateM_(2)`.

The code should be:

```scala
/** Return this if it is a right, otherwise, return the given value. Alias for `|||` */
  def orElse(x: => EitherT[F, A, B])(implicit F: Monad[F]): EitherT[F, A, B] = {
    val g = run
    EitherT(F.bind(g) {
      case    -\/(_)  => x.run
      case r@(\/-(_)) => F.point(r)
    })
  }
```

_Edit:_ this actually needs a `Monad[F]` rather than a `Bind[F]`, and some extra parenthesis around `\/-(_)`.

dansanduleac:
Fixes #1037  and adds a test for it as well.

### DIFF
--- a/core/src/main/scala/scalaz/EitherT.scala
+++ b/core/src/main/scala/scalaz/EitherT.scala
@@ -144,16 +144,16 @@ final case class EitherT[F[_], A, B](run: F[A \/ B]) {
     F.map(run)(_ valueOr x)
 
   /** Return this if it is a right, otherwise, return the given value. Alias for `|||` */
-  def orElse(x: => EitherT[F, A, B])(implicit F: Bind[F]): EitherT[F, A, B] = {
+  def orElse(x: => EitherT[F, A, B])(implicit F: Monad[F]): EitherT[F, A, B] = {
     val g = run
     EitherT(F.bind(g) {
-      case -\/(_) => x.run
-      case \/-(_) => g
+      case    -\/(_)  => x.run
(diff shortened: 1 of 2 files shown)


jedws:
+1

PR #1040, opened 2015-11-08
deprecate all the unsafe methods on Future/Task

jedws:
and have them call newly named `unsafePerform*` variants


puffnfresh:
Is it always unsafe to run a Task?

jedws:
@puffnfresh wdym? `1.point[Task].unsafePerformSync` is safe, is that what you mean?

tpolecat:
In practice `.run` is unsafe in exactly the same way `.unsafePerformIO` is unsafe. My impr.ession is that `Task` has subsumed `IO` for most users

jedws:
@tpolecat exactly, but it is called `.run` which is the same as the nice and safe `run` on many of the functional data structures (eg.`Kleisli`).

vmarquez:
I like this change, but I do think methods like attemptRun and runAsync should somehow be marked as *more* unsafe than than the bounded (time wise) ones.  Thoughts @jedws  ?

jedws:
@vmarquez I'm not sure that "more unsafe" is much use to signify, or how you'd go about signifying it (`reallyUnsafe*` ?). With this change, we are simply trying to distinguish between the things that release the rockets and the pure API.

Perhaps another PR could add more documentation or something, but a

wedens:
:+1:

xuwei-k:
What is the status of this? I don't have a strong opinion this, but it's about time I want to release `7.2.0-RC1` and `7.2.0 final`.

tpolecat:
:+1: I would like to get this in for 7.2 if we can.

jedws:
@xuwei-k I've been waiting for others to opine and maybe merge, I really want to see it in 7.2

Issue #1041, opened 2015-11-08
mixup in 7.1.x `NonEmptyList#distinct`


refried:
There was a mixup in this cherry-pick 7a9f66c7dbc6438f34ed8c2cac61d1217c9110e6 because the new method takes an `Order[A]` but doesn't use it. (It delegates to `scala.collection.immutable.List#distinct`, not `IList`).

Will submit a PR.

refried:
(kind of a big bug!  have to use .toIList everywhere to avoid the buggy method on NonEmptyList)

PR #1042, opened 2015-11-08
fix NonEmptyList.distinct to use its implicit Order

refried:
per #1041

### DIFF
--- a/core/src/main/scala/scalaz/NonEmptyList.scala
+++ b/core/src/main/scala/scalaz/NonEmptyList.scala
@@ -1,5 +1,7 @@
 package scalaz
 
+import scala.annotation.tailrec
+
 /** A singly-linked list that is guaranteed to be non-empty. */
 final class NonEmptyList[+A] private[scalaz](val head: A, val tail: List[A]) {
   import NonEmptyList._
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/NonEmptyList.scala
```diff
@@ -42,10 +44,20 @@ final class NonEmptyList[+A] private[scalaz](val head: A, val tail: List[A]) {
...
 
-  def distinct[AA >: A](implicit A: Order[AA]): NonEmptyList[A] =
-    (list.distinct: @unchecked) match {
-      case x :: xs => nel(x, xs)
-    }
+  def distinct[AA>:A](implicit A: Order[AA]): NonEmptyList[AA] = {
+    @tailrec def loop(src: List[A], seen: ISet[AA], acc: NonEmptyList[AA]): NonEmptyList[AA] =
+      src match {
+        case h :: t =>
+          if(seen.notMember(h)){
```
jedws: really minor, but spaces before and after braces. We should incorporate an autoformatter…

xuwei-k:
Nice catch! thanks :+1:

core/src/main/scala/scalaz/NonEmptyList.scala
refried: Oops, sorry, will fix.  That was accidentally copied from 5b53c2239e3d139d7804fd09145a5070f0873e2b though.

PR #1043, opened 2015-11-08
add [A:Order]distinct and [A:Equal]distinctE to Foldable/Foldable1

refried:
`IList[A].distinct` and `NonEmptyList[A].distinct` currently require `Order[A]`, even though they could complete with just `Equal[A]`, although the latter would need *O(n^2)* runtime instead of *O(n log n)*.

This patch adds both methods to `Foldable` and `Foldable1`, returning `IList[A]` and `NonEmptyList[A]` respectively.

What do folks think?

### DIFF
--- a/core/src/main/scala/scalaz/Foldable.scala
+++ b/core/src/main/scala/scalaz/Foldable.scala
@@ -255,6 +255,24 @@ trait Foldable[F[_]]  { self =>
       }
     })._1
 
+  /** ``O(n log n)`` complexity */
+  def distinct[A](fa: F[A])(implicit A: Order[A]): IList[A] =
+    foldLeft(fa, (ISet.empty[A],IList.empty[A])) {
+      case ((seen, acc), a) =>
+        if (seen.notMember(a))
+          (seen.insert(a), a :: acc)
+        else (seen, acc)
+    }._2.reverse
+
+  /** ``O(n^2^)`` complexity */
+  def distinctE[A](fa: F[A])(implicit A: Equal[A]): IList[A] =
+    foldLeft(fa, IList.empty[A]) {
+      case (seen, a) =>
+        if (!IList.instances.element(seen,a))
+          a :: seen
+        else seen
+    }.reverse
+
   def collapse[X[_], A](x: F[A])(implicit A: ApplicativePlus[X]): X[A] =
     foldRight(x, A.empty[A])((a, b) => A.plus(A.point(a), b))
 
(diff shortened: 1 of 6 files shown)


xuwei-k:
This is not binary compatible. please send to `series/7.2.x` branch.

refried:
@xuwei-k Wait, why is it not binary compatible?

refried:
(There are no deleted or changed lines.)

xuwei-k:
https://travis-ci.org/scalaz/scalaz/jobs/89893998#L6391

```
[info] scalaz-core: found 4 potential binary incompatibilities
[error]  * method distinct1(java.lang.Object,scalaz.Order)scalaz.NonEmptyList in trait scalaz.Foldable1 does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("scalaz.Foldable1.distinct1")
[error]  * method distinctE1(java.lang.Object,scalaz.Equal)scalaz.NonEmptyList in trait scalaz.Foldable1 does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("scalaz.Foldable1.distinctE1")
[error]  * method distinct(java.lang.Object,scalaz.Order)scalaz.IList in trait scalaz.Foldable does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("scalaz.Foldable.distinct")
[error]  * method distinctE(java.lang.Object,scalaz.Equal)scalaz.IList in trait scalaz.Foldable does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("scalaz.Foldable.distinctE")
```

refried:
Those are "potential" binary incompatibilities, not actual binary incompatibilities, aren't they?
Isn't that error just saying that a method was added?  That doesn't create a binary incompatibility, does it it?

refried:
@xuwei-k Why do they call that a binary incompatibility?

jedws:
@refried the problem is that a `Foldable` instance compiled against 7.1.4 won't have that method, so will fail with some kind of `MissingMethodError`. Everything that declares `Foldable` instances would need to be recompiled.

xuwei-k:
https://github.com/xuwei-k/binary-incompatibility-sample

refried:
@xuwei-k Thanks for creating that example for me.  I almost still didn’t understand, but I think I do now:

The instance compiled against library 0.1 doesn’t provide the new method, but the application compiled against library 0.2 expects it to be to be there.

I guess that code in Scala traits gets copied into the class file of the instances, whereas for Java 8 interface default methods, it lives in the class file for the interface, so this scenario shouldn’t introduce BC errors once Scala incorporates a similar model; right?

Thanks.

refried:
Thanks @jedws, just saw your comment too.

refried:
P.S. Tldr: Adding methods to classes is ok, but to traits isn’t. (Right?)

refried:
https://github.com/scala/scala-dev/issues/35

xuwei-k:
> P.S. Tldr: Adding methods to classes is ok, but to traits isn’t. (Right?)

Yes, that's right.

PR #1044, opened 2015-11-08
add [A:Order]distinct and [A:Equal]distinctE to Foldable/Foldable1

refried:
how about this one :-)


xuwei-k:
:+1:

refried:
Do you think we should remove/deprecate the duplicate implementations from `IList` and `NonEmptyList`, since they don’t add any implementation-specific performance improvements? Or what's the general rule of thumb on that?

PR #1045, opened 2015-11-08
BindRec should override `forever`

puffnfresh:
Instead of having its own `foreverRec` method.

Pointed out by @tonymorris.


jedws:
+1

PR #1046, opened 2015-11-09
added BindRec instances for LazyEither, LazyEitherT and LazyOptionT

jedws:
and add tests

### DIFF
--- a/core/src/main/scala/scalaz/LazyEither.scala
+++ b/core/src/main/scala/scalaz/LazyEither.scala
@@ -160,8 +160,8 @@ object LazyEither extends LazyEitherInstances {
 
 // TODO more instances
 sealed abstract class LazyEitherInstances {
-  implicit def lazyEitherInstance[E] =
-    new Traverse[LazyEither[E, ?]] with Monad[LazyEither[E, ?]] with Cozip[LazyEither[E, ?]] with Optional[LazyEither[E, ?]] with MonadError[LazyEither, E] {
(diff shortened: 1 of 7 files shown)


core/src/main/scala/scalaz/LazyEither.scala
```diff
@@ -160,8 +160,8 @@ object LazyEither extends LazyEitherInstances {
 
 // TODO more instances
 sealed abstract class LazyEitherInstances {
-  implicit def lazyEitherInstance[E] =
-    new Traverse[LazyEither[E, ?]] with Monad[LazyEither[E, ?]] with Cozip[LazyEither[E, ?]] with Optional[LazyEither[E, ?]] with MonadError[LazyEither, E] {
+  implicit def lazyEitherInstance[E]: Traverse[LazyEither[E, ?]] with Monad[LazyEither[E, ?]] with BindRec[LazyEither[E, ?]] with Cozip[LazyEither[E, ?]] with Optional[LazyEither[E, ?]] with MonadError[LazyEither, E] =
```
puffnfresh: :+1: to the explicit type there, too!

core/src/main/scala/scalaz/LazyEitherT.scala
```diff
@@ -189,6 +189,12 @@ sealed abstract class LazyEitherTInstances1 {
       def iso = LazyEitherT.lazyEitherTLeftProjectionEIso2[F, L]
     }
 
+  implicit def lazyEitherTBindRec[F[_], L](implicit F0: Monad[F], B0: BindRec[F]): BindRec[LazyEitherT[F, L, ?]] =
```
puffnfresh: Seems like this will be ambiguous with the next implicit when looking for:

```
Bind[LazyEitherT[F, ?, ?]]
```

Given `F: BindRec`.

jedws: good catch, thanks

tonymorris:
:+1:

puffnfresh:
:+1:

PR #1047, opened 2015-11-11
MonadError that takes a unary-kinded F[_] (rather than F[_, _] 

jedws:
This is the first part of a series that will do the same for all MTL type-classes.


tpolecat:
Could @xuwei-k and/or @S11001001 verify the placement of the methods in 7ec5aae262c456dfeae29f4f8932681496fe6be4 … works but the implicit priority may not be ideal.

jedws:
this is subsumed by #1049 which makes all the MTL type classes take a unary-kinded F[_]

PR #1048, opened 2015-11-11
Refactor RWST to allow for trampolining

vmarquez:
This is the same change as 

https://github.com/scalaz/scalaz/pull/949

in that it allows for stack safety as long as any underlying F is trampolined or otherwise a free monad.  This touches a lot less than IndexedStateT and so hopefully a lot easier to review.

### DIFF
--- a/core/src/main/scala/scalaz/ReaderWriterStateT.scala
+++ b/core/src/main/scala/scalaz/ReaderWriterStateT.scala
@@ -3,7 +3,11 @@ package scalaz
 /** A monad transformer stack yielding `(R, S1) => F[(W, A, S2)]`. */
 sealed abstract class IndexedReaderWriterStateT[F[_], -R, W, -S1, S2, A] {
   self =>
-  def run(r: R, s: S1): F[(W, A, S2)]
+  
+  def getF[S <: S1, RR <: R]: Monad[F] => F[(RR, S) => F[(W, A, S2)]]
+  
+  def run(r: R, s: S1)(implicit F: Monad[F]): F[(W, A, S2)] = 
+    F.join( F.map[(R, S1) => F[(W, A, S2)], F[(W, A, S2)]](getF(F))(f => f(r, s)) )
 
   /** Discards the writer component. */
   def state(r: R)(implicit F: Monad[F]): IndexedStateT[F, S1, S2, A] =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/ReaderWriterStateT.scala
```diff
@@ -3,7 +3,10 @@ package scalaz
 /** A monad transformer stack yielding `(R, S1) => F[(W, A, S2)]`. */
 sealed abstract class IndexedReaderWriterStateT[F[_], -R, W, -S1, S2, A] {
   self =>
-  def run(r: R, s: S1): F[(W, A, S2)]
+  
+  def getF[S <: S1, RR <: R]: Monad[F] => F[(RR, S) => F[(W, A, S2)]]
+  
+  def run(r: R, s: S1)(implicit F: Monad[F]): F[(W, A, S2)] = F.join( F.map[(R, S1) => F[(W, A, S2)], F[(W, A, S2)]](getF(F))(f => f(r, s)) )
```
jedws: minor point, but please put sig on first line and impl on second. It is quite hard to see that this is now not abstract.

jedws:
is this the same aim as BindRec/FreeT? I'm struggling to understand a lot of this, and think it needs further analysis to check whether the impl needs so many more constraints and whether it isn't actually made redundant with the BindRec/FreeT work.

tests/src/test/scala/scalaz/ReaderWriterStateTTest.scala
```diff
@@ -27,12 +27,19 @@ object ReaderWriterStateTTest extends SpecLite {
...
+      .foldLeft(ReaderWriterStateT[Trampoline, Unit, String, Int, Int]((_, i: Int) => Trampoline.done(("", i, i))))( (a, b) => a.flatMap(_ => b))
+    10000 must_=== result.run((),0).run._3
+  }
+ 
   object instances {
     def functor[F[_]: Functor, R, W, S] = Functor[RWST[F, R, W, S, ?]]
     def plus[F[_]: Plus, R, W, S1, S2] = Plus[IRWST[F, R, W, S1, S2, ?]]
     def plusEmpty[F[_]: PlusEmpty, R, W, S1, S2] = PlusEmpty[IRWST[F, R, W, S1, S2, ?]]
-    def bindRec[F[_]: BindRec : Applicative, R, W: Semigroup, S] = BindRec[RWST[F, R, W, S, ?]]
+    def bindRec[F[_]: BindRec : Monad, R, W: Semigroup, S] = BindRec[RWST[F, R, W, S, ?]]
```
jedws: is it necessary to change this? could it be done with just existing BindRec? It seems like there is maybe a problem with BindRec that it couldn't be done with that and needs an explicit `Trampoline` but I'm probably missing something. cc @puffnfresh

core/src/main/scala/scalaz/ReaderWriterStateT.scala
vmarquez: Sure, will do.

vmarquez:
@jedws 'made redundant' sort of depends on if we are factoring in convenience.  FreeT is sort of superfluous if we decide it's ok for everyone to lift their transformer stack with liftT and then fold the resulting free structure, but I imagine it's not near as convenient at times.  

My PR is just a giving a guarantee that if you have say a IRWST[Trampoline, A, A], you get stack safety as is.  No lifting or using FreeT necessary.  A lot of folks have this intuition already, so I'd say the current implementation isn't as correct as it could be.  You could substitute Trampoline for Task or another trampoline based construct and have it work too with my change.  This is the same as #949 where I talk about 'how' my fix works in detail.

jdegoes:
`BindRec` is *not* a substitute for trampolining and this work (making the default transformers trampoline-able) is useful — even essential in making Scalaz stack-safe by default.

puffnfresh:
@jdegoes BindRec is exactly a substitute, but:

1. Stack safety without using BindRec is useful
2. Stack safety on transformers when the inner Monad is not BindRec is useful
3. This change makes a lot of RWST's methods more constrained, limiting the ways it can be used
4. In practice the limitations probably won't be hit

Therefore I'm on the :+1: side, but only by a little bit.

puffnfresh:
@vmarquez it's possible to use `BindRec[RWST[Trampoline, X, Y, ?]]` - you don't have to go via FreeT to get stack safety.

jdegoes:
@puffnfresh This paradigm fits awkwardly into the usual way of developing with monad transformers, and I also agree about (4).

core/src/main/scala/scalaz/ReaderWriterStateT.scala
```diff
@@ -12,49 +16,46 @@ sealed abstract class IndexedReaderWriterStateT[F[_], -R, W, -S1, S2, A] {
...
+      })
 }
 
 object IndexedReaderWriterStateT extends ReaderWriterStateTInstances with ReaderWriterStateTFunctions {
   def apply[F[_], R, W, S1, S2, A](f: (R, S1) => F[(W, A, S2)]): IndexedReaderWriterStateT[F, R, W, S1, S2, A] = new IndexedReaderWriterStateT[F, R, W, S1, S2, A] {
-    def run(r: R, s: S1): F[(W, A, S2)] = f(r, s)
+    override def getF[S <: S1, RR <: R]: Monad[F] => F[(RR, S) => F[(W, A, S2)]] = (F: Monad[F]) => F.point((r: R, s: S) => f(r, s))
+  }
+
+  def apply[F[_], R, W, S1, S2, A](f: Monad[F] => (R, S1) => F[(W,A, S2)]): IndexedReaderWriterStateT[F, R, W, S1, S2, A] = new IndexedReaderWriterStateT[F, R, W, S1, S2, A] {
```
xuwei-k: I think we should avoid overloading.

jedws: agreed. good catch @xuwei-k

vmarquez:
I agree @xuwei-k and @jedws. I have fixed the overloading issue and rebased.

xuwei-k:
:+1:

PR #1058, opened 2015-11-24
FreeT hoist, interpret

xuwei-k:
https://github.com/paf31/purescript-freet/blob/v0.3.1/src/Control/Monad/Free/Trans.purs#L86-L97

### DIFF
--- a/core/src/main/scala/scalaz/FreeT.scala
+++ b/core/src/main/scala/scalaz/FreeT.scala
@@ -60,6 +60,46 @@ sealed abstract class FreeT[S[_], M[_], A] {
       case a => gosub(a)(f)
     }
 
+  /**
+   * Changes the underlying `Monad` for this `FreeT`, ie.
+   * turning this `FreeT[S, M, A]` into a `FreeT[S, N, A]`
+   * given Functors for `S` and `N`
+   */
+  def hoistN[N[_]](mn: M ~> N)(implicit S: Functor[S], N: Functor[N]): FreeT[S, N, A] =
+    this match {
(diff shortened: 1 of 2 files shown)


puffnfresh:
Brilliant :+1:

core/src/main/scala/scalaz/FreeT.scala
```diff
@@ -39,14 +50,50 @@ sealed abstract class FreeT[S[_], M[_], A] {
...
   final def flatMap[B](f: A => FreeT[S, M, B]): FreeT[S, M, B] =
     this match {
-      case Gosub(a, k) => Gosub(a, (x: Any) => Gosub(k(x), f))
-      case a => Gosub(a, f)
+      case g @ Gosub() => gosub(g.a)(x => gosub(g.f(x))(f))
+      case a => gosub(a)(f)
+    }
+
+  /** Same as `hoistM` but different constraints */
+  def hoistN[N[_]](mn: M ~> N)(implicit S: Functor[S], N: Functor[N]): FreeT[S, N, A] =
```
jedws: Can we improve this doc? The purescript doc says `Change the underlying `Monad` for a `FreeT` action.` and the signature here is not as obvious as it is an instance method. Perhaps something like:

    /**
     * Changes the `Monad` for this `FreeT`, ie. turning
     * this `FreeT[S, M, A]` into a `FreeT[S, N, A]`
     * given Functors for `S` and `N`
     */

xuwei-k: fixed

PR #1060, opened 2015-11-26
Add generic zipWithIndex to Traverse syntax

puffnfresh:


tonymorris:
Maybe name it something else so as not to clash with stdlib? Torn.

puffnfresh:
I'm mostly concerned with structures which do not provide it, although this might provide it better.

jedws:
won't structures that do provide it already directly simply use the one they provide? Ie. you couldn't use this syntax to call the `Traverse` provided `zipWithIndex`, but would you want to?

puffnfresh:
@jedws yes, and so you get ridiculous shit like this:

```
scala> (Some(1): Option[Int]).zipWithIndex
res0: Iterable[(Int, Int)] = List((1,0))
```

Instead of something working.

I think I'm on the side of changing the name.

puffnfresh:
Also, it looks like I've flipped `zipWithIndex` when compared to the convention in stdlib.

tonymorris:
Yeah, the problem is when the stdlib takes the name and does something silly with it. There is a history of this. I don't know if `zipWithIndex` already does something silly, or plans on it for the future, but I know where my bets are.

It's the real reason that unicode identifier names are a good idea; name things exactly as others would not.

xuwei-k:
:+1: except discussion about name.

puffnfresh:
Re-targeting at 7.3

PR #1067, opened 2015-12-01
fix foldMap stack safety

xuwei-k:
https://github.com/non/cats/pull/702

/cc @runarorama


xuwei-k:
Oh, maybe this is incorrect https://github.com/non/cats/issues/712

jedws:
can you add a test for that?

xuwei-k:
I think we should use [this technique](https://github.com/scalaz/scalaz/commit/d76c426b2b921129693a1) again.

xuwei-k:
I will send another pull request.

PR #1071, opened 2015-12-03
Make Applicative instance for Future sequential

markus1189:
DISCLAIMER: I was unsure if I should do this pr at all, because I profit from the parallel applicative instance of Future currently, but I think it should at least be mentioned ;)

The issue is that in the definition of `ap` for the standard `scala.concurrent.Future` it uses the `Future.zip` method which runs in parallel (introduced here 4fc139114f38cb408e56e9cdb00d0307e2b089d4).  This clashes with the rule that the `ap` definition should be equivalent to the definition which uses `flatMap` if there also is a `Monad` instance. 

Maybe you are all aware of this fact and decided that it is nevertheless a beneficial evil, so take it or leave it ;)

-- 24pullrequests
[](http://24pullrequests.com/)

### DIFF
--- a/core/src/main/scala/scalaz/std/Future.scala
+++ b/core/src/main/scala/scalaz/std/Future.scala
@@ -65,10 +65,6 @@ private class FutureInstance(implicit ec: ExecutionContext) extends Nondetermini
   override def gather[A](fs: Seq[Future[A]]): Future[List[A]] =
     Future.sequence(fs.toList)
 
-  // override for actual parallel execution
-  override def ap[A, B](fa: => Future[A])(fab: => Future[A => B]) =
-    fab zip fa map { case (fa, a) => fa(a) }
-
   def attempt[A](f: Future[A]): Future[Throwable \/ A] =
     f.map(\/.right).recover { case e => -\/(e) }
 


puffnfresh:
Didn't know zip runs in parallel. Thanks! :+1:

puffnfresh:
Travis is giving an error for this, though:

```
[info] ! scalaz.std.FutureTest.monad error.monad.bind.ap consistent with bind: Falsified after 8 passed tests.
java.lang.AssertionError: -\/("SomeFailure(2)") !== -\/("SomeFailure(1)")
	at scalaz.SpecLite.fail(SpecLite.scala:56)
	at scalaz.SpecLite$AnyOps.must_$eq$eq$eq(SpecLite.scala:63)
	at scalaz.std.FutureTest$$anonfun$1.apply$mcV$sp(FutureTest.scala:62)
	at scalaz.std.FutureTest$$anonfun$1.apply(FutureTest.scala:57)
	at scalaz.std.FutureTest$$anonfun$1.apply(FutureTest.scala:57)
```

puffnfresh:
Looks like just the order is wrong. Flip the statements in the for-comprehension and it should be good to go.

xuwei-k:
>  Flip the statements in the for-comprehension and it should be good to go.

or just remove `override def ap` ?

puffnfresh:
@xuwei-k even better :+1:

markus1189:
Okay I removed the override completely, unfortunately I can't test it locally because I somehow can't even compile due to errors in `scalaz/std/Tuple.scala` relating to `Tuple<n>Monad`...  Do I miss something?

aloiscochard:
I'm totally for this change (but it should be well documented in the release note), I have workaround in work codebase due to the current implemnentation... I would be happy to remove it.

:+1: thanks

S11001001:
:-1:

This is unobservable unless you're doing side-effects.

If we make every parallel computation monad side-effect-friendly, there's no space left for people who aren't doing side-effects.

Before making any further changes in this area, I'd like to see an enumeration of Future-like things and which you think ought to be for side effects. "All of them" is not a good answer.

aloiscochard:
@S11001001 good point :-/

I do personally use `Future` for holding side effects, I do not use `Task` nor `IO` in our work codebase.

But... I'm probably a bad guy, I never got really convince about current `Task`/`IO` implementation, I change my upvote to downvote then... as I realize how surprising this can be (for people using `IO` and who use `Future` precisely to specify they want parallelism).

This is surely something we should think hard for a potential next major version, I would be really keen on a matterhorn like system that would make it clear where effects should be.

tpolecat:
I enjoy pointing out that `scala.concurrent.Future` has **nothing at all** to say about side-effects. If you are side-effecting then you lose RT immediately, so I think scalaz's treatment is only reasonable if we assume `Future` is only being used for pure computations. My preference would probably be to slice it off entirely and not provide any instances at all, but I guess that's another discussion.

Both `IO` and `Task` provide explicit effect-capturing constructors that do not memoize their results, so clearly their design and intent is sequential by default.

puffnfresh:
@S11001001 I think we agree that I should have the "same" program when I refactor from Monad to Applicative, and I think Futures running in parallel leaves me with a different program.

I don't care what happens when side-effects are involved.

tpolecat:
Confusion. I don't understand how there is an observable difference between parallel and sequential for pure computations.

jedws:
@tpolecat I don't think saying that `Future` is not about side-effects, while `Task` is, is really a consistent position. `Task` is really just a `Future[Throwable \/ ?]`, and it has been co-opted for the purpose of encapsulating side-effects for not necessarily very good reasons.

tpolecat:
@jedws I thought we were talking about stdlib `Future` ... at this point I kind of see scala**z** `Future` as an implementation detail for `Task`. I don't get the sense that many people use it directly.

puffnfresh:
@tpolecat the observed performance will be different. I don't know want performance of my programs to significantly change when I refactor. For example, I sometimes give in and use for-comprehensions in Scala due to inference problems with Applicatives. I don't want my programs to go from parallel execution to sequential by accident, I want to be explicit, therefore I want some more types.

A few teams within Atlassian use the underlying Future in Task, quite often. I've opened #1086 to talk about the Task problem, though.

tonymorris:
If the point from @puffnfresh is important [I'm still thinking about it], then is it worth having:

* A `Future` data type with a parallel `Applicative` instance, but without any `Monad` instance.

AND

* A (distinct) `Future` data type with both an `Applicative` and `Monad` instance, neither of which are parallel?

puffnfresh:
@tonymorris that's the approach taken by scalaz Future, which I think is a more useful approach:

https://github.com/scalaz/scalaz/blob/series/7.3.x/concurrent/src/main/scala/scalaz/concurrent/Future.scala#L383

And for clarification: I'm not saying that parallel execution of `Applicative[Future]` is bad because it creates side-effects (I don't think it does) - I think it's bad because I care about when I'm being parallel and not, and I want scalaz to care, too.

S11001001:
@tonymorris Yes, there should be both, and we should be clear about which is which, and indeed for any futureish thing which style it is meant for.

I disagree with @puffnfresh in that I don't think performance should be considered in expression equality; switching to applicative syntax, thereby declaring the independence of bindings, is a performance-improving technique broadly applicable to monads. Lots of equality substitutions have performance implications; I don't see this one as special.

S11001001:
...but I recognize why you might prefer being explicit, scalaz Future style. I just don't think it's always appropriate.

markus1189:
To chime in, I thought that `scalaz.Nondeterminism` is the encouraged approach if you want to use the `scala.concurrent.Future` with "runs-in-parallel" semantics. e.g. by using `mapBoth` or am I missing something?

I also agree with @puffnfresh that while having an instance f or Applicative that runs in parallel is not something inherently bad, I also don't want to be surprised by this when refactoring my code by replacing `for-comprehensions` with applicative syntax.

bmjames:
@markus1189 While `mapBoth`, on its own, may be fine for pure computations in `Future`, I don't think it's an acceptable solution to require a `Nondeterminism` instance if you only want deterministic parallel computation.

aloiscochard:
Just a quick note, to emphasis on @S11001001 point, I think that Haxl is a good example where we would like such optimization to happen seamlessly, I agree that in the case we are discussing here, I prefer precision and I would find such optimization as surprising.

It all depends of the domain imho.

S11001001:
@tonymorris Sorry, I didn't read your comment closely enough; I'm suggesting your first described type *should* have a `Monad` instance, for the reason described.

jedws:
@tpolecat in our production stack we have `EitherT` to capture errors, so we have no need to use `Task`, and have `scalaz.concurrent.Future` as the inner, effects capturing container. As @puffnfresh points out though, 'Future'/'Task' as an effects capturer leaves quite a bit to be desired.

puffnfresh:
@aloiscochard I think [Haxl is actually broken](https://chadaustin.me/2016/02/dropbox-hack-week-graphql-server-in-haskell/) because of this reason:

2016-02-09-140540_509x379_scrot

aloiscochard:
@puffnfresh I see, that is very interesting thanks for sharing, it's definitely broken to do it like that with `IO`, I agree.

That being said it could be totally sensible and desired for a very specific type to have such optimization behavior. I do not have a concrete real world example in mind, but this would not necessary shocks me. I can imagine domain where that would be sound and a sensible things to do.

S11001001:
To expand, as a rule, we do not accept changes in performance as a determinant of inequality.  This ship has sailed; refactoring by lawful equalities can change the performance of your program, and this has nothing to do with `Future`.

For example, given `A: *`, `xs: List[A]`, `xs.map(identity) = xs`, even though the performance of the latter is obviously better by `n` and some consing than that of the former.

For another example, while it is true that `m flatMap f flatMap g = m flatMap (a => f(a) flatMap g)`, it is common for the observed performance of the latter to be better. See @dolio's [post](https://plus.google.com/+DanDoel/posts/B1pSqNojj3k).

I do not see what should make `Future` an exception to this practice.

@markus1189 So it's too late; the performance characteristics *are* different for equal expressions in applicative and monad syntax. This is a good thing; otherwise, we would be forbidden by ap-consistency from ever producing a more efficient `ap` for monads taking advantage of the improved sharing and independence of `ap` vs `flatMap`.  Not to mention a more efficient `map`.

In one way, this is kind of the point: you can notice that an expression can be rewritten according to a lawful equality, and do so to gain something like improved performance, secure in the knowledge that the inputs-to-outputs of your program will not change as a result. Perhaps one day, if we are very good, we'll get ApplicativeDo, and then you'll only have to worry about accidentally introducing a spurious data dependency between lines in your `for` sequence.

markus1189:
> @markus1189 So it's too late

Okay thanks everybody for the insightful comments, I'll go ahead and close this.

LeifW:
To clarify: the commit mentioned didn't introduce parallelism; it was that way before. Stdlib Futures start executing when they're created. You see the _ prefixed intermediate values introduced to force the evaluation of the by-name Future values before putting them in the for-comprehension, in the old code.

LeifW:
Though, it seems that `// override for actual parallel execution` comment has been a lie this whole time. :(

PR #1074, opened 2015-12-05
Generalize forever to Apply

LeifW:
* Added `*>` and `<*` to the instance, instead of just in the syntax.
* Added forever to syntax also.

Example usage:
```scala
import scalaz._, Scalaz._, effect._, IO._

val a: IO[Unit] = putStrLn("Hello, world!")
val b: IO[Unit] = a.forever
b.unsafePerformIO // prints "Hello, world!" forever
```

### DIFF
--- a/core/src/main/scala/scalaz/syntax/ApplySyntax.scala
+++ b/core/src/main/scala/scalaz/syntax/ApplySyntax.scala
@@ -8,11 +8,11 @@ final class ApplyOps[F[_],A] private[syntax](val self: F[A])(implicit val F: App
   final def <*>[B](f: F[A => B]): F[B] = F.ap(self)(f)
   final def tuple[B](f: F[B]): F[(A,B)] = F.tuple2(self,f)
 
-  /** Combine `self` and `fb` according to `Apply[F]` with a function that discards the `A`s */
-  final def *>[B](fb: F[B]): F[B] = F.apply2(self,fb)((_,b) => b)
+  /** Combine `self` and `fb` according to `Apply[F]` and discard the `A`(s) */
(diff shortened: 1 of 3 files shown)


LeifW:
While doing this, I was a little curious why the functions on the instance take their arguments by-name, whereas in the syntax stuff seems by-value (except the `^`'s).

tonymorris:
:+1:

core/src/main/scala/scalaz/syntax/ApplySyntax.scala
```diff
@@ -31,7 +31,11 @@ final class ApplyOps[F[_],A] private[syntax](val self: F[A])(implicit val F: App
   /** Alias for `|@|` */
   final def ⊛[B](fb: F[B]) = |@|(fb)
 
-  ////
+  /**
```
tpolecat: The `////` is significant to the build ... I think `forever` need to go above it.

tpolecat:
:+1: after fixing the mechanical issue in `ApplySyntax`.

core/src/main/scala/scalaz/syntax/ApplySyntax.scala
```diff
@@ -31,7 +31,11 @@ final class ApplyOps[F[_],A] private[syntax](val self: F[A])(implicit val F: App
...
+   * Repeats this applicative action infinitely.
+   */
+  final def forever[B]: F[B] = F.forever(self)
+
```
puffnfresh: Just needs a `////` after here.

puffnfresh:
:+1:

xuwei-k:
Could you squash commits?

LeifW:
Squash'd

PR #1077, opened 2015-12-07
Change after function to use scheduler instead of Timer

nlim:
I was trying out the Task#retry functionality in Task and I found that it uses a scalaz.concurrent.Timer to implement its logic, and it appears scalaz.concurrent.Timer uses Thread.sleep.

Following the example of Future#timed, which uses the ScheduledExecutorService, we can avoid the use of Thread.sleep in the implementation of Task#retry


aloiscochard:
:+1:

vmarquez:
👍

PR #1080, opened 2016-01-03
fix digit append

mking:
I found a typo in the digit append implementation.

I updated the test to test this case.


tonymorris:
Cheers.

PR #1081, opened 2016-01-08
Add a leftMapNel for ValidationNel

YannMoisan:


aloiscochard:
I'm not sure about such a specialized function here, I mean where will we stop if we start adding such functions?

As a corollary, we add it here on `Validation`, but why not adding it on `Bifunctor` then?

I'm curious what others think, I feel like to achieve that we should rather have a newtype with a specialized Bifunctor instance.

xuwei-k:
> I'm not sure about such a specialized function here, I mean where will we stop if we start adding such functions?
> As a corollary, we add it here on Validation, but why not adding it on Bifunctor then?

I agree with aloiscochard too.

YannMoisan:
since the comment of @aloiscochard , the proposal is more generic (it works on any `Functor` instead of `NonEmptyList`).

Some of us should reimplement these combinators in their projects…

Do you close the PR because the code is not at the right place or just you don't want this kind of specific combinator in scalaz ?

tonymorris:
I think overly-specialised functions are a necessity for Scala, since generalisation often trades off against the utility of the type system. I am not sure if that is the case here, but I suspect so.

aloiscochard:
@YannMoisan it's totally reasonable to want to avoid duplicating combinators in every project!

I'm reopening the PR for discussion, but this time, now that I see the implementation with `Functor` I would like to ask:
Aren't we looking for a `ValidationT` here? /cc @tonymorris @S11001001 @puffnfresh

wedens:
I agree with @aloiscochard. There can be tons of such functions, and library should not be overloaded with very specialized functions.

xuwei-k:
> <del>Aren't we looking for a ValidationT here?</del> Shouldn't we favor creating transfomers for the Monad we want to compose?

- https://github.com/scalaz/scalaz/pull/67
- https://github.com/scalaz/scalaz/commit/9907b91048a270339

aloiscochard:
@xuwei-k indeed, thanks for the references.

@YannMoisan I would be curious to know, which Functor you usually lift into `Validation`? to see if they basically comes from a `Monad` or not and if we could transform for them to help you. (If you would like to discuss it in real-time feel free to pop in #scalaz on freenode! would be nice to understand a bit more the backgroun.)

YannMoisan:
Here are both use cases (it seems that they are very standard) : 
- for `nestedMap` :
I want to validate an optional field (from a CSV file), so it's quite straightforward to use a `Validation[A, Option[B]]`. And after that, I want to apply B => C

-for `leftNestedMap` 
I have a ValidationNel[E, A] and I want to enrich all errors (with a line number), so apply E => EE.

wedens:
https://hackage.haskell.org/package/bifunctors-5.1/docs/Data-Bifunctor-Biff.html

aloiscochard:
@YannMoisan thanks for the concrete example.

So here is the way to achieve it using (bi)functor composition (thanks @wedens for the pointer):

```
import scalaz._                                                                   
import scalaz.std.AllInstances._                                                  
val foo: Validation[List[String], Option[Int]] = ???                              
val functorR = Bifunctor[Validation].rightFunctor[List[String]].compose[Option]   
val functorL = Bifunctor[Validation].leftFunctor[Option[Int]].compose[List]       
functorR.map(foo)(_ + 2)                                                          
functorL.map(foo)(_ + ":1")                                                       
```

This is sad that scala type inference make things such verbose, but if we want to workaround the inference issue we should find about a relatively generic solution (thinks Unapply) instead of introducing highly specialized functions.

You have two solution now:
- In your project (or common), you keep a list of most used composition and refer to them directly (basically store those `functorR`/`functorL` in a shared place in your projects, that's what I usually do)
- Find a syntactic workaround (which make the inference works) that you might want to contribute to scalaz. That involve some research.

I close the issue for now, I hope you agree on the reasoning here. Inference can be very frustrating :-(

YannMoisan:
Thanks for such a complete answer and all your smart comments.
I have some questions :
- @aloiscochard nice trick with `BiFunctor`. But it seems that is not generic (the code that I proposed is).
- for right part, is there an advantage to use `BiFunctor`, `Functor` seems sufficient :
``` 
def functorR[A] = Functor[({type l[a] = Validation[A, a]})#l].compose[Option]
functorR.map(foo)(_ + 2)
```
- for `Unapply`, I'm still a scalaz noob (it's my first PR). So, i don't yet fully understand `Unapply` magic :)

wedens:
@YannMoisan functor composition is the most generic thing possible when you have "nested `map`"
`Functor` instance for `Validation` "works" on `Success` values, so yeah, `Functor` is sufficient

aloiscochard:
Hi @YannMoisan,

* 1. that is true, if we extract your code and try to write an abstract function for it, it will work purely on `Functor` while mine would have an extra `Bifunctor` constraint!

If you write a generic function you should **definitely** follow the approach you just shared (using type lambda to get the `Functor` directly and then compose it).

But I'm pretty sure that you would agree that even if this approach is more generic, it is a also a hell less friendly to write?

I thought intuitively, based on your initial example, that you were not writing generic function but you were "just" using it on concrete types. That is why I did show the Bifunctor approach.
In this case I find the `Bifunctor` nicer to use (syntax wise), it should be noted that this can be mitigated using the `kind-projector` plugin.

* 2. Unapply, that's quite a beast!

That's is a tricky thing to understand as it takes advantage of relatively advanced and obscure scala technics, I hope those articles will help:
* http://eed3si9n.com/learning-scalaz/Unapply.html
* http://typelevel.org/blog/2013/09/11/using-scalaz-Unapply.html

Good luck!

Feel free to let us know if something is not clear, or just pop in #scalaz on freenode.

PR #1082, opened 2016-01-08
lift a value in a Validation

YannMoisan:


wedens:
isn't it the same as `toSuccess` and `toSuccessNel` from `scalaz.syntax.std.option._`?

YannMoisan:
It's just a straightforward helper, to propose the same feature as `BooleanOps.either` but for `Validation`. wdyt ?

PR #1083, opened 2016-01-18
Add Bifunctor widen

edmundnoble:
Adds a function `widen` to take advantage of the natural covariance of `Bifunctor`. I was initially going to implement this for `EitherT` because that was my use-case, but this is probably more useful.


xuwei-k:
Could you squash commits?

edmundnoble:
Squashed.

xuwei-k:
:+1:

aloiscochard:
:+1: thanks

PR #1090, opened 2016-01-29
Make NonEmptyList#tails stack safe.

TomasMikula:
The original implementation was converted to a test of correctness.


xuwei-k:
:+1:

PR #1091, opened 2016-01-29
Add method NonEmptyList#inits.

TomasMikula:
`IList` has both `tails` and `inits`, so I think `NonEmptyList` should, too.

The implementation is in terms of `tails` and `reverse`, which is analogous to the implementation of `IList#inits`.


TomasMikula:
The [failed Travis build](https://travis-ci.org/scalaz/scalaz/jobs/105769242#L3370) apparently has nothing to do with this commit.

PR #1092, opened 2016-01-29
Add method IList#foreach.

TomasMikula:
Since `NonEmptyList` already has a `foreach` method, I thought `IList` should, too.

### DIFF
--- a/core/src/main/scala/scalaz/IList.scala
+++ b/core/src/main/scala/scalaz/IList.scala
@@ -240,6 +240,9 @@ sealed abstract class IList[A] extends Product with Serializable {
   def map[B](f: A => B): IList[B] =
     foldRight(IList.empty[B])(f(_) :: _)
 
+  def foreach(f: A => Unit): Unit =
+    toList.foreach(f)
+
   // private helper for mapAccumLeft/Right below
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/IList.scala
```diff
@@ -240,6 +240,9 @@ sealed abstract class IList[A] extends Product with Serializable {
   def map[B](f: A => B): IList[B] =
     foldRight(IList.empty[B])(f(_) :: _)
 
+  def foreach(f: A => Unit): Unit =
+    toList.foreach(f)
```
tpolecat: My instinct here would be to write out the tail-recursive method to avoid constructing the `List` (and avoid relying on other methods for evaluation order).

tpolecat:
Honestly I think I would prefer to just get rid of `foreach` on `NonEmptyList` … there are very few side-effects in scalaz and I am reluctant to add another.

core/src/main/scala/scalaz/IList.scala
TomasMikula: Makes sense. However, I was just following what [NonEmptyList.foreach](https://github.com/scalaz/scalaz/blob/de094534fc3cda329b05c0995106c440c26f979e/core/src/main/scala/scalaz/NonEmptyList.scala#L25-L28) is doing.

TomasMikula:
Closing this for the above reason (side-effecting).

PR #1094, opened 2016-02-03
Add a `filterLength` function

puffnfresh:
Counts how many matches of a predicate occur in a Foldable.

The same as:

    filterLength f = length . filter f

But this version is manually fused, for better efficiency in Scala.


runarorama:
:+1:

tonymorris:
:+1:

Issue #1098, opened 2016-02-11
IMap / ISet should allow lazy folding


dansanduleac:
Currently we have these constructs that drive traversal of an `IMap`:

```scala
def foldlWithKey[C](z: C)(f:  (C, A, B) => C): C
def foldrWithKey[C](z: C)(f: (A, B, C) => C): C
```

Such that any and all traversals of an `IMap` require converting the structure into a `List` or some other `O(n)` strict memory structure (I'm including `Stream` in here as well, since the entire structure of the stream still has to be created).

This is exemplified by some of the provided instances as well: `Show` and `Order` both require converting the entire structure into a `List` before doing anything else. This is clearly sub-optimal. 

__But__ with a simple change, it should be possible to fold lazily in either direction, and with that in place, replace any `toAscList` intermediate operations with something like `toEphemeralStream`. That change is:

```scala
// (exact same implementations, but I've made `z`, as well as the `C` parameter to `f`, lazy)
def foldlWithKey[C](z: => C)(f:  (=> C, A, B) => C): C
def foldrWithKey[C](z: => C)(f: (A, B, => C) => C): C
```

I've added these to a pimp, called them `foldrWithKey0` and equivalent, and ran some benchmarks:

```scala
import FixedIMap._

val testMap = IMap.fromList(1 to 1024 map (i => i -> i) toList)

  def count[T](act: (=> Unit) => T): T = {
    var called = 0
    val out = act(called+=1)
    println(called)
    out
  }

  // Default implementation prints 1024 :(
  count { act => testMap.foldrWithKey(EphemeralStream[(Int, Int)]()) ((a, b, c) => { act; (a -> b) ##:: c }) }
  // Our implementation prints 1 if we don't access any of the elements.
  val stm = count { act => testMap.foldrWithKey0(EphemeralStream[(Int, Int)]()) ((a, b, c) => { act; (a -> b) ##:: c }) }
  // Prints 501 -- we're only evaluating f for as long as we need to.
  count { act =>
    val foo = testMap.foldrWithKey0(EphemeralStream[(Int, Int)]()) ((a, b, c) => { act; (a -> b) ##:: c })
    require(foo.take(500).length == 500)
  }
```

Exactly the same can be done with an `ISet`.

PR #1099, opened 2016-02-12
Fix IMap/ISet to support lazy folding, and optimise intermediary operations

dansanduleac:
This addresses #1098.

By adding variants of the `foldLeft`/`foldRight` methods which are lazy, it's now possible to fold over these structures efficiently, without allocating an additional `O(n)`.

* A lot of intermediate operations, as well as typeclasses, were using `toList` and `fromList` behind the scenes, and such uses have been replaced with `toAscEphemeralStream` and `fromEphemeralStream`. 
* Similar to the above, for both `IMap` and `ISet`, hashCode now calls `MurmurHash3.orderedHash` on an iterator that traverses the stream, rather than `.toList.hashCode`.
* `Foldable` / `BiFoldable` instances now use the lazy variant of `foldRight` in each case, making all such operations efficient as well.

Necessary changes to `EphemeralStream`:
* Implement `Equal`, `Order` and `Show`, plus tests
* Add `zipLongest` method (plus tests), which is used in the implementation of `Equal` and `Order`.

_Bonus_: Fix `Foldable.foldMap` implementation for `EphemeralStream` to use `foldLeft` rather than `foldRight`, since the latter can easily blow the stack, for example if you do:
```scala
EphemeralStream.range(0, 1000000).foldMap(ISet.singleton)
```

### DIFF
--- a/core/src/main/scala/scalaz/EphemeralStream.scala
+++ b/core/src/main/scala/scalaz/EphemeralStream.scala
@@ -2,6 +2,8 @@ package scalaz
 
 import java.lang.ref.WeakReference
 
+import scala.annotation.tailrec
+
 /** Like [[scala.collection.immutable.Stream]], but doesn't save
   * computed values.  As such, it can be used to represent similar
   * things, but without the space leak problem frequently encountered
@@ -102,6 +104,14 @@ sealed abstract class EphemeralStream[A] {
       emptyEphemeralStream
     else
       cons((head(), b.head()), tail() zip b.tail())
+
+  def zipLongest[B](b: => EphemeralStream[B]): EphemeralStream[A \&/ B] =
+    if (isEmpty)
+      b.map(\&/.That(_))
+    else if (b.isEmpty)
+      map(\&/.This(_))
+    else
+      cons(\&/.Both(head(), b.head()), tail() zipLongest b.tail())
 
   def unzip[X, Y](implicit ev: A <:< (X, Y)): (EphemeralStream[X], EphemeralStream[Y]) =
     foldRight((emptyEphemeralStream[X], emptyEphemeralStream[Y]))(q => r =>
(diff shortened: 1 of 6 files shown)


xuwei-k:
please send to `series/7.3.x` branch

dansanduleac:
Thanks for your quick reply. 
May I ask, why is it not possible to merge in 7.1.x? It's not breaking binary compatibility as far as I can tell.
(I'm asking because I mostly use 7.1 because of scalaz-stream's current dependency on it, and would like to be use the changes in 7.1)

xuwei-k:
Here is `sbt mimaReportBinaryIssues` result

```
[error]  * method fromEphemeralStreamWithKey(scalaz.EphemeralStream,scala.Function3,scalaz.Order)scalaz.==>> in trait scalaz.MapFunctions does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("scalaz.MapFunctions.fromEphemeralStreamWithKey")
[error]  * method fromEphemeralStreamWith(scalaz.EphemeralStream,scala.Function2,scalaz.Order)scalaz.==>> in trait scalaz.MapFunctions does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("scalaz.MapFunctions.fromEphemeralStreamWith")
[error]  * method fromEphemeralStream(scalaz.EphemeralStream,scalaz.Order)scalaz.==>> in trait scalaz.MapFunctions does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("scalaz.MapFunctions.fromEphemeralStream")
[error]  * method fromEphemeralStream(scalaz.EphemeralStream,scalaz.Order)scalaz.ISet in trait scalaz.ISetFunctions does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("scalaz.ISetFunctions.fromEphemeralStream")
```


Furthermore, `MurmurHash3` does not exists in Scala 2.9

https://travis-ci.org/scalaz/scalaz/jobs/108780527#L454

```
[error] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/ISet.scala:572: object hashing is not a member of package util
[error]     scala.util.hashing.MurmurHash3.orderedHash(EphemeralStream.toIterable(stm).iterator)
[error]                ^
[warn] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/Map.scala:872: non variable type-argument A in type pattern scalaz.==>>[A,B] is unchecked since it is eliminated by erasure
[warn]       case that: ==>>[A, B] =>
[warn]                  ^
[error] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/Map.scala:880: object hashing is not a member of package util
[error]     scala.util.hashing.MurmurHash3.orderedHash(EphemeralStream.toIterable(stm).iterator)
[error]                ^
```

dansanduleac:
Thanks for the info! I admit I didn't consider the `MirmurHash3` thing as I forgot this is being compiled for 2.9 as well.

Are the mima errors a problem in this case though? Surely, within a minor version, new functions have been introduced in patch versions before, so long as the semantics or signature of existing functions don't change.

xuwei-k:
> so long as the semantics or signature of existing functions don't change.

Yes, but mima is right. see https://github.com/scalaz/scalaz/pull/1043 and https://github.com/xuwei-k/binary-incompatibility-sample

PR #1100, opened 2016-02-12
Fix IMap/ISet to support lazy folding, and optimise intermediary operations

dansanduleac:
This addresses #1098.

By adding variants of the `foldLeft`/`foldRight` methods which are lazy, it's now possible to fold over these structures efficiently, without allocating an additional `O(n)`.

* A lot of intermediate operations, as well as typeclasses, were using `toList` and `fromList` behind the scenes, and such uses have been replaced with `toAscEphemeralStream` and `fromEphemeralStream`. 
* Similar to the above, for both `IMap` and `ISet`, hashCode now calls `MurmurHash3.orderedHash` on an iterator that traverses the stream, rather than `.toList.hashCode`.
* `Foldable` / `BiFoldable` instances now use the lazy variant of `foldRight` in each case, making all such operations efficient as well.

Necessary changes to `EphemeralStream`:
* Implement `Equal`, `Order` and `Show`, plus tests
* Add `zipLongest` method (plus tests), which is used in the implementation of `Equal` and `Order`.


core/src/main/scala/scalaz/Map.scala
```diff
@@ -1160,6 +1172,16 @@ object ==>> extends MapInstances {
...
 
+  /* EphemeralStream operations */
+  final def fromEphemeralStream[A: Order, B](s: EphemeralStream[(A, B)]): A ==>> B =
+    s.foldLeft(empty[A, B]) { t => x => t.insert(x._1, x._2) }
+
+  final def fromEphemeralStreamWith[A: Order, B](s: EphemeralStream[(A, B)])(f: (B, B) => B): A ==>> B =
+    fromEphemeralStreamWithKey(s)((_, x, y) => f(x, y))
+
+  final def fromEphemeralStreamWithKey[A: Order, B](s: EphemeralStream[(A, B)])(f: (A, B, B) => B): A ==>> B =
+    s.foldLeft(empty[A, B])(a => c => a.insertWithKey(f, c._1, c._2))
```
S11001001: Why do the `fromFoldable*` variants not suffice for these three functions?

tests/src/test/scala/scalaz/EphemeralStreamTest.scala
```diff
@@ -10,16 +10,21 @@ object EphemeralStreamTest extends SpecLite {
...
   checkAll(zip.laws[EphemeralStream])
   checkAll(align.laws[EphemeralStream])
   checkAll(cobind.laws[EphemeralStream])
 
-  implicit def ephemeralStreamShow[A: Show]: Show[EphemeralStream[A]] =
+  def ephemeralStreamSlowShow[A: Show]: Show[EphemeralStream[A]] =
     Show[List[A]].contramap(_.toList)
 
+  "show" ! forAll{ e: EphemeralStream[Int] =>
+    implicitly[Show[EphemeralStream[Int]]].show(e) must_===(ephemeralStreamSlowShow[Int].show(e))
```
S11001001: Can be written `Show[EphemeralStream[Int]]` without the `implicitly[...]`.

core/src/main/scala/scalaz/EphemeralStream.scala
```diff
@@ -218,9 +228,44 @@ sealed abstract class EphemeralStreamInstances {
...
+      }
+      case other => _ => other   // short circuit.
+    }
+  }
+
+  implicit def ephemeralStreamShow[A: Show]: Show[EphemeralStream[A]] = Show.show { stm =>
+    val A = Show[A]
+
+    @tailrec
+    def commaSep(rest: EphemeralStream[A], acc: Cord): Cord =
```
S11001001: I think `Cord.mkCord` should suffice for handling the comma separation here.  (Maybe a `Foldable` variant should be added for that function...)

emilypi:
@dansanduleac is this still relevant for you?

dansanduleac:
@emilypi Not relevant to me anymore, shall I close?

PR #1101, opened 2016-02-14
Add mapF in analogy to flatMapF to Monad Transformers

justjoheinz:
Given map(A => B) and flatMapF(A => G[F[B]) we should have mapF(A => G[B])
I also fixed an anomaly in the flatMapF signature of OptionT compared to the other transformers.
I also added a show instance for MaybeT.
This would break existing code, please review carefully.

### DIFF
--- a/core/src/main/scala/scalaz/EitherT.scala
+++ b/core/src/main/scala/scalaz/EitherT.scala
@@ -71,6 +71,12 @@ final case class EitherT[F[_], A, B](run: F[A \/ B]) {
   def map[C](f: B => C)(implicit F: Functor[F]): EitherT[F, A, C] =
     EitherT(F.map(run)(_.map(f)))
 
+  /** Map on the right of this disjunction. */
+  def mapF[C](f: B => F[C])(implicit M: Monad[F]): EitherT[F, A, C] =
+    flatMapF {
+      f andThen (mb => M.map(mb)(b => \/-(b)))
+    }
+
   /** Traverse on the right of this disjunction. */
   def traverse[G[_], C](f: B => G[C])(implicit F: Traverse[F], G: Applicative[G]): G[EitherT[F, A, C]] =
     G.map(F.traverse(run)(o => Traverse[A \/ ?].traverse(o)(f)))(EitherT(_))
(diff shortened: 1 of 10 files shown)


jedws:
seems like there's an interface these could implement to guarantee consistency, something like

/**
     * Standard interface for monad transformers
     * 
     * F is what a transformer is parameterised by
     * M is the underlying type being transformed
     * T is the transformer type itself (F-bound)
     * A is the contained type
     */
    trait MonadTransformer[F[_], M[_], T[_[_], _], A] {
      def mapF[X](f: A => X): T[F, X]
      def flatMapF[X](f: A => F[M[X]])(implicit F: Monad[F]): T[F, X]
    }

I don't know if this is valuable or not.

justjoheinz:
I did not do something like that, since currently not all Monad Transformers implement something like flatMapF. I only added mapF (which would be `def mapF[X](f: A => F[X]): T[F, X]` for transformers which already provided flatMapF.

But describing these functions in a trait would have avoided the problem of the misaligned implemention of flatMapF in OptionT of course.

vmarquez:
:+1:

@justjoheinz if you can get another +1 I'll merge.  Thanks.

justjoheinz:
Anything I can do, do get this into master?

TomasMikula:
This method, unlike `flatMapF`, indeed seems to have a uniform implementation for any monad transformer:

```scala
trait MonadTrans[F[_[_], _]] {
  def liftM[G[_] : Monad, A](a: G[A]): F[G, A]

  implicit def apply[G[_] : Monad]: Monad[F[G, ?]]

  def mapF[G[_], A, B](fa: F[G, A])(f: A => G[B])(implicit M: Monad[G]): F[G, B] =
    Monad[F[G, ?]].bind(fa)(a => liftM(f(a)))
}
```

(Maybe a better name here would be `mapM` for consistency with `liftM`.)

justjoheinz:
Ok, I see that I can get this implemented. flatMapF should work as well, as the current diverging implementations are inconsistent.

TomasMikula:
@justjoheinz I don't see how you could implement `flatMapF` uniformly, without knowing the implementation details of individual data types. It seems to me that `flatMapF` is there for data types that are just wrappers around other data types, and `flatMapF(f)` is like `flatMap(f)`, except it allows `f` to return an unwrapped value. In a generic setting, it is impossible to tell what the unwrapped type for `T[F, A]` is (or that `T` is a wrapper type at all).

PR #1102, opened 2016-02-15
Add a Band type-class

puffnfresh:
Gives an optimised `multiply1` method.

---

TODO:
- [ ] Instances for tags (e.g. First and Last)
- [ ] Instances for Sets
- [ ] value-coinductive instances for Maps

### DIFF
--- a/core/src/main/scala/scalaz/Band.scala
+++ b/core/src/main/scala/scalaz/Band.scala
@@ -1,0 +1,42 @@
+package scalaz
+
+////
+/**
+ * [[scalaz.Semigroup]] which is also idempotent, i.e. appending a value with
+ * itself results in the same value.
+ *
+ * @see [[scalaz.Band.BandLaw]]
+ */
+////
+trait Band[F] extends Semigroup[F] { self =>
+  ////
+
+  // derived functions
+
+  /**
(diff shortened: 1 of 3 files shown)


tonymorris:
LGTM. Is there any prior work that this is based on?

puffnfresh:
@tonymorris a little. `Data.Semigroup` allows `stimes` to be overridden for "idempotent semigroups" (i.e. bands) - since idempotency is just a law, I thought it'd make sense as a refined type-class.

@ekmett might have a thought on this.

jbgi:
would instances for {```ISet```, ```union```/```intersection```} also be useful?

ekmett:
Makes sense to me.

ekmett:
@jbgi Indeed.

fthomas:
The [conjunction and disjunction `Monoid[Boolean]`s](https://github.com/scalaz/scalaz/blob/1ea53e0721280e290abc12259be3a45d2940c429/core/src/main/scala/scalaz/std/AnyVal.scala#L52-L62) also have an idempotent `append` and therefore could extend `Band`.

xuwei-k:
rebased and merged in https://github.com/scalaz/scalaz/commit/679095043944f27a8ef487a45302e5bbf55f02a2

Issue #1103, opened 2016-02-15
Faster EphemeralStream weakMemo with OCC


dansanduleac:
I've been running some benchmarks on `scala.Stream` vs `scalaz.EphemeralStream`.

Let's define a timing function:

``` scala
def timed(stuff: => Any, iterations: Int = 100): Unit = {
  val vs = for (_ <- 1 to iterations) yield { 
    System.gc // try to get rid of potential GC pauses
    var t = System.currentTimeMillis
    stuff
    System.currentTimeMillis - t
  }
  println(s"Average time after $iterations iterations: ${vs.sum / vs.length}ms")
}
```

Now, 

```
scala> timed(scalaz.EphemeralStream.range(0, 10000000).length)
Average time after 100 iterations: 731ms
scala> timed(Stream.range(0, 10000000).length)
Average time after 100 iterations: 418ms
```

The `EphemeralStream` traversal takes `75%` longer than the Stream traversal.

But if we were to change `EphemeralStream.weakMemo` to use an `AtomicReference` instead of a synchronized block, allowing for the head/tail to possibly be evaluated multiple times in parallel:

``` scala
def weakMemo[V](f: => V): () => V = {
    val ref: AtomicReference[WeakReference[V]] = new AtomicReference()
    () => {
      def genNew(old: WeakReference[V]) = {
        val x = f
        ref.compareAndSet(old, new WeakReference(x))
        x
      }
      val v = ref.get()
      if (v != null) {
        val crt = v.get()
        if (crt == null) {
          genNew(v)
        } else crt
      } else genNew(v)
    }
  }
```

```
scala> timed(scalaz.EphemeralStream.range(0, 10000000).length)
Average time after 100 iterations: 489ms
```

This is now only `17%` slower, yet still maintaining the benefit of `head` and `tail` being individually lazy (as opposed to, say, saving the `(head, tail)` tuple inside a `WeakReference` instead and continuing to use synchronization but for the entire tuple).

Would you say this is a worthwhile improvement?

runarorama:
:+1:

S11001001:
:+1:

S11001001:
Related: #1259

S11001001:
Fixed by #1348.

PR #1104, opened 2016-02-15
Improve EphemeralStream's weakMemo to be optimistic using CAS

dansanduleac:
- Implement the change outlined in #1103.
- Add a test which reads the same stream from 4 threads concurrently, and ensures that the values being read are correct and in the right order. This should check that contention over the memoized head/tail values does not break safety.

### DIFF
--- a/tests/src/test/scala/scalaz/EphemeralStreamTest.scala
+++ b/tests/src/test/scala/scalaz/EphemeralStreamTest.scala
@@ -132,4 +132,15 @@ object EphemeralStreamTest extends SpecLite {
       EphemeralStream.fromStream(Stream.iterate(0)(_ + 1).zipWithIndex.take(n))
     )
   }
+
+  "reading from a stream in parallel should be safe" in {
+    import scalaz.concurrent._
+    val limit = 10000000
+    val stm = EphemeralStream.range(1, limit + 1)
+    val nthreads = 4
+    // Ensure that we get back the right numbers, in the right order.
+    val tsks = Task.gatherUnordered( List.fill(nthreads)(Task(stm.foldLeft(0)(prev => n => { (prev + 1) must_=== n; n }))) )
+    // And that the result contains the last number for each of the threads.
+    tsks.unsafePerformSync must_=== List.fill(nthreads)(limit)
+  }
 }
--- a/core/src/main/scala/scalaz/EphemeralStream.scala
+++ b/core/src/main/scala/scalaz/EphemeralStream.scala
@@ -1,6 +1,7 @@
 package scalaz
 
 import java.lang.ref.WeakReference
+import java.util.concurrent.atomic.AtomicReference
 
 /** Like [[scala.collection.immutable.Stream]], but doesn't save
   * computed values.  As such, it can be used to represent similar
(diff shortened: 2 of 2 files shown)


dansanduleac:
@jbgi Interestingly when running 4 threads in parallel that consume the stream, the version you deleted seems *faster* than the non-looping one.
Turns out this is because evaluating `f` outside of `genNew` seems to be noticeably faster in itself. I suspect this is because in the former case (where it captures over the by-name parameter `f`), scala won't inline the function, for some reason.

I've fixed all functions to have the evaluated `x` passed into `genNew` rather than evaluating `f` within.

Then I ran some benchmarks using the `timed` function I outlined in #1103, but modified to also do 5 'warm-up' iterations first and ignore the time for them:

```scala
  def timed(stuff: => Any, iterations: Int = 100): Unit = {
    // warm up
    (1 to 5) foreach (_ => stuff)
    val vs = for (_ <- 1 to iterations) yield {
      System.gc // try to get rid of potential GC pauses
      val t = System.currentTimeMillis
      stuff
      System.currentTimeMillis - t
    }
    println(s"Average time after $iterations iterations: ${vs.sum / vs.length}ms")
  }

  def stuff = {
    val stm = scalaz.EphemeralStream.range(0, 10000000)
    val tsks = List.fill(4)(Task(stm.length))
    val t = Task.gatherUnordered(tsks)
    t.unsafePerformSync
  }

  timed(stuff, 100)
```

### 4 threads in parallel that consume the stream
Note: All versions are fixed so that genNew is `@inline` and takes the `x: V` as a parameter

Version | Average time after 100 iterations |
--------- | ------- |
First version that calls genNew tail-recursively | 973ms |
Version that does not loop | 975ms |
My version | 940-974ms | 

So about the same ball-park. They all seem to vary between 940 and 1000 ms on my machine, when not memory-constrained.
I'm not sure if this change necessarily increases performance or reduces contention, maybe you can find a use-case or a benchmark to illustrate this?

jbgi:
interesting.
As for your version, if you are not going to use the result of ```compareAndSet``` then. maybe, you might as well use ```set``` (not exactly the same, would need to benchmark).
And if you only use ```set``` then you don't need the ```AtomicReference``` wrapper at all: a ```@volatile var``` would be enough.

jbgi:
As for the use case of my version, you would need no trivial transformations of the stream for any improvement to be noticeable, I think. If you map over it with a ```sleep(1)``` identity function the benchmark would be somewhat realistic I think.

tests/src/test/scala/scalaz/EphemeralStreamTest.scala
```diff
@@ -132,4 +132,15 @@ object EphemeralStreamTest extends SpecLite {
       EphemeralStream.fromStream(Stream.iterate(0)(_ + 1).zipWithIndex.take(n))
     )
   }
+
+  "reading from a stream in parallel should be safe" in {
+    import scalaz.concurrent._
+    val limit = 10000000
+    val stm = EphemeralStream.range(1, limit + 1)
```
S11001001: Due to #1259, will need `.memoized` at the end of this line.  Else, `weakMemo` will not be used here.

S11001001:
Will close #1172, #1103.

jbgi:
@S11001001, @dansanduleac I just re-did some tests with what I consider to be a relatively realistic benchmark, and the following implementation of `genNew`
```scala
      @inline
      def genNew(x: V, old: WeakReference[V]): V = {
        if (ref.compareAndSet(old, new WeakReference(x)))
          x
        else {
          val crt = ref.get.get
          if (crt != null) crt
          else x
        }
      }
```
got the benchmark to run in 10ms on average vs. 15ms for the current patch implementation.
The benchmark being:
```scala
  timed(stuff, 10)

  def stuff = {
    import scalaz.concurrent._
    val stm = scalaz.EphemeralStream.range(1, 10000).map(i => (BigDecimal(Math.exp(Math.sqrt(i) / i)) * BigDecimal(Math.log(Math.sqrt(i) / i))).rounded).memoized
    val tsks = List.fill(4)(Task(stm.foldLeft[BigDecimal](BigDecimal(0.0))(_ => a => a)))
    val t = Task.gatherUnordered(tsks)
    t.unsafePerformSync
  }
```
So I believe this is worth it (will open a PR on the patch).

S11001001:
Fixed by #1348.

PR #1106, opened 2016-02-19
FreeApUsage example

beezee:
Illustrates using FreeAp, example is a simple exercise in parsing meaningfully typed values out of a Map[String, Any]


vmarquez:
Interesting. (about FreeAp).  Cool example though, 👍 from me for what it's worth.

jedws:
:+1:

PR #1107, opened 2016-02-20
Un-deprecating timed and retry as they do not side effect

vmarquez:
If you look at how 'unsafeRetryInternal' is implemented, it doesn't actually side effect.

https://github.com/scalaz/scalaz/blob/series/7.3.x/concurrent/src/main/scala/scalaz/concurrent/Task.scala#L272

The same can be said of unsafePerformTimed in Future (which Task delegates to for its timed functionality).  

https://github.com/scalaz/scalaz/blob/series/7.3.x/concurrent/src/main/scala/scalaz/concurrent/Future.scala#L248

It's very useful to be able to call retry on one part of a Task (and not others) before the 'end of the world' or even before passing back a Task from a function, so it is confusing to see it called 'unsafe' when it doesn't actually run anything.


vmarquez:
@jedws I think we talked about this in #scalaz so let me know your thoughts.  Thanks.

jbgi:
But those methods can observe lawful refactoring or any runtime conditions that induce performance modification, therefore violating rerenrential transparency. So, by a corollary argument of the one made by @S11001001 in https://github.com/scalaz/scalaz/pull/1071 those methods would be kept ```unsafe```.

vmarquez:
@jbgi how is it any different than calling async with a side effect inside? It's not. You can't say that's referentially transparent and say that a retry wouldn't be.

jedws:
I think I'm convinced, but don't think you can just remove the `unsafe*` as 7.2 clients will be using them. They need to be put back and deprecated unfortunately.

vmarquez:
@jedws I did this on the 7.3, so it's OK to make breaking changes, no?  I suppose we should have a separate PR for back porting to 7.2 replacements.  Let me know though, I'm happy to do both or either.

jedws:
@vmarquez preferable to not break though if we can avoid it, and have at least one deprecation cycle before removing it. From a source POV many will have warnings as errors and will go through the change as soon as they upgrade, but with deprecations at least you get a "use foo instead" message rather than no such method.

vmarquez:
Ok I've added them back in, and marked them as deprecated.  Let me know if you want any other changes/cleanup, happy to do so.  @jedws

jedws:
👍 

I think we want to remove the implicit schedulers in a separate PR

PR #1109, opened 2016-02-22
Foldable.splitBy and splitByRelation

jbgi:
```splitBy``` splits the elements into groups that produce the same result  by a function f:
```scala
scala> List("A","B","C", "AA","AB", "A", "ABC", "EFG") splitBy (_.length)
res1: scalaz.IList[(Int, scalaz.NonEmptyList[String])] =
   [(1,NonEmpty[A,B,C]),(2,NonEmpty[AA,AB]),(1,NonEmpty[A]),(3,NonEmpty[ABC,EFG])]
```
following @ritschwumm comment, also added ```splitByRelation``` that splits into groups of elements that are transitively dependant by a relation r:
```scala
scala> List("A","BA","BC", "AA","BC", "AC") splitByRelation ((s1, s2) => !s2.intersect(s1).isEmpty)
res6: scalaz.IList[scalaz.NonEmptyList[String]] = [NonEmpty[A,BA,BC],NonEmpty[AA],NonEmpty[BC,AC]]
```

### DIFF
--- a/core/src/main/scala/scalaz/Foldable.scala
+++ b/core/src/main/scala/scalaz/Foldable.scala
@@ -263,6 +263,17 @@ trait Foldable[F[_]]  { self =>
       }, Some(pa))
     })._1
 
+  /**
+    * Splits the elements into groups that produce the same result by a function f.
+    */
+  def splitBy[A, B: Equal](fa: F[A])(f: A => B): IList[(B, NonEmptyList[A])] =
+    foldRight(fa, IList[(B, NonEmptyList[A])]())((a, bas) => {
+      val fa = f(a)
+      bas match {
+        case INil() => IList.single((fa, NonEmptyList(a)))
+        case ICons((b, as), tail) => if (Equal[B].equal(fa, b)) ICons((b, a <:: as), tail) else ICons((fa, NonEmptyList(a)), bas)
+      }
+    })
 
   /**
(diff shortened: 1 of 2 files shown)


ritschwumm:
could be even more general by using a manually supplied equality relation instead of a implicit Equal instance, right?

jbgi:
@ritschwumm yes.
Should it be ```scala.Math.Equiv[B]``` or simply ```(B,B) => Boolean``` then?

jbgi:
on second thought, this would be a separate method, and cannot use```Equiv``` because the relation might not be transitive, so something like
```scala
def splitByRelation[A](fa: F[A])(relation: (A, A) => Boolean): IList[NonEmptyList[A]]
```
WDYT? (a better name?)

ritschwumm:
the second thought looks good to me. the name - well... i can't think of a better one, but that doesn't mean it's great.

vmarquez:
Looks good to me, before I merge though, @jbgi what do you think about returning an IMap instead of an IList?  It looks to me like (a better) groupBy.

http://www.scala-lang.org/files/archive/nightly/docs/library/index.html#scala.collection.immutable.List

jbgi:
@vmarquez well, in my use case I really need to keep the elements in the same order after the split. more like sliding windows than a groupBy.

PR #1113, opened 2016-03-09
More stack-safety for FreeT.

TomasMikula:
It is crucial that

    Gosub(a, f1) >>= f2

is not transformed _eagerly_ to

    Gosub(a, x => f1(x) >>= f2)

because then

    Gosub(a, f1) >>= f2 >>= f3 >>= f4 >>= ...

builds up `Gosub(a, g)` with `g` stack-unsafe.

On the other hand, if we keep it as

    Gosub(...Gosub(Gosub(Gosub(Gosub(a, f1), f2), f3), f4)...)

we then in `resume` compose functions from the right and each of the composed functions is stack safe (`f3`, `f4` are composed to `g3 = x => Gosub(f3(x), f4)`, then `f2`, `g3` to `g2 = x => Gosub(f2(x), g3)` and finally `f1`, `g2` to `g1 = x => Gosub(f1(x), g2)`. We see that stack-safety of each _g<sub>i</sub>_ depends only _f<sub>i</sub>_, not on _g<sub>i+1</sub>_, so the stack frames don't add up.)

As a consequence, for `FreeT[S, M, ?]` to be a `Functor`, `M` must be at least `Applicative`.

I checked that the two newly added test cases did not pass before.

### DIFF
--- a/core/src/main/scala/scalaz/FreeT.scala
+++ b/core/src/main/scala/scalaz/FreeT.scala
@@ -63,18 +63,15 @@ object FreeT extends FreeTInstances {
 }
 
 sealed abstract class FreeT[S[_], M[_], A] {
-  final def map[B](f: A => B)(implicit S: Functor[S], M: Functor[M]): FreeT[S, M, B] =
+  final def map[B](f: A => B)(implicit S: Functor[S], M: Applicative[M]): FreeT[S, M, B] =
     this match {
       case Suspend(m) => Suspend(M.map(m)(_.bimap(f, S.lift(_.map(f)))))
-      case g @ Gosub() => gosub(g.a)(g.f.andThen(_.map(f)))
+      case g @ Gosub() => gosub(g)(x => point(f(x)))
     }
 
   /** Binds the given continuation to the result of this computation. */
   final def flatMap[B](f: A => FreeT[S, M, B]): FreeT[S, M, B] =
-    this match {
-      case g @ Gosub() => gosub(g.a)(x => gosub(g.f(x))(f))
-      case a => gosub(a)(f)
-    }
+    gosub(this)(f)
 
   /**
    * Changes the underlying `Monad` for this `FreeT`, ie.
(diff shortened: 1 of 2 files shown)


xuwei-k:
Yeah, Maybe this is more stack-safe. but require more strong constraints( `Functor` => `Applicative` ).
I think this is a trade-off. Should we choose stack-safety or weak-constraints? or does anyone have a better solution?

aloiscochard:
Yeah it's definitely a trade-off.

I don't have a strong opinion here, but I'll definitely look if there is anything doable about it in my current experiment. I feel like it should be possible to get best of both worlds with an alternative encoding, but maybe I'm dreaming awake ;)

TomasMikula:
My thinking was that in order to interpret `FreeT[S, M, A]`, `M` will have to be a `Monad` anyway. The only case where you can escape without ever requiring `M` to be a `Monad` seems to be `hoistM`/`hoistN`, but in that case `N` takes on the "obligation" to be a monad (or be eventually transformed into a monad).

So suppose that is the case: we have `fa: FreeT[S, M, A]` and `M ~> N`, where `N` is a monad and `M` is not even an applicative. We cannot do `fa.map(f)`, because we don't have `Applicative[M]`. As a rescue, we can always use `Free[M, ?]` instead of `M`, i.e. `FreeT[S, Free[M, ?], A]`. Now we can do `map`s, and since we have `M ~> N` with `M` a functor and `N` a monad, we also have `Free[M, ?] ~> N`, so we keep the ability to `hoist` to `N`, too. So at least we don't lose any expressivity.

I actually find it more unfortunate that `map` [requires](https://github.com/scalaz/scalaz/blob/cf0282d8c3a5a26d4f7e7093d3b3fbe8720b9da6/core/src/main/scala/scalaz/FreeT.scala#L66) `Functor[S]`. This requirement could now (i.e. after requiring `Applicative[M]`) be dropped, if we just implement `map` in terms of `flatMap`. This would be convenient, because we could do `map`s and `flatMap`s on `FreeT[S, M, A]` without `Functor[S]`, and only later `interpretT` to `FreeT[Coyoneda[S, ?], M, A]`. (I could add that to this PR, or do that later in a separate one.)

TomasMikula:
#1115 shows that we can then lift the `S[_]: Functor` constraint.

TomasMikula:
And #1116 takes this even further after slightly changing the representation of `FreeT` (the `Suspend` case).

xuwei-k:
@puffnfresh @jedws Do you have any thoughts?

PR #1116, opened 2016-03-11
Relax more FreeT constraints

TomasMikula:
This is a follow up to #1113 and #1115.

Changing `FreeT[S, M, A]` encoding slightly and embracing that `M` will eventually need to be a monad anyway, so it is OK to require `Applicative[M]` anywhere, we are able to lift more `S[_]: Functor` constraints on `FreeT` operations.

The encoding and constraints on `S` are now more analogous to `Free`.

### DIFF
--- a/core/src/main/scala/scalaz/Free.scala
+++ b/core/src/main/scala/scalaz/Free.scala
@@ -325,7 +325,7 @@ sealed abstract class Free[S[_], A] {
   /** Extraction from `Free` as a comonad in the endofunctor category. */
   def extractF(implicit S: Monad[S]): S[A] = foldMap(NaturalTransformation.refl[S])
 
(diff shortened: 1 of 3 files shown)


TomasMikula:
I will also note that eager `map`s on `S` could have led to [building up](https://github.com/scalaz/scalaz/blob/cf50500e0a1ca357d85b545a6bbc0890ceebde4d/core/src/main/scala/scalaz/Coyoneda.scala#L36-L37) a non-stack-safe function inside `S`, in case `S =:= Coyoneda[S0, ?]`. After this is merged, on the other hand, it doesn't even seem worth it to add a test for it, because if the operations on `FreeT[S, M, A]` don't assume anything about `S`, they cannot do anything with the enclosed `S` instances, in particular they cannot break them by accumulating `map`s.

puffnfresh:
I think this is the right approach.

PR #1117, opened 2016-03-11
Don't require Functor[S] for MonadPlus[FreeT[S, M, ?]] and MonadError[FreeT[S, M, ?], E]

TomasMikula:
This concludes a series (#1115, #1116) of refactorings eliminating the need for `Functor[S]` instance for all operations except `resume`/`runM` and all typeclass instances except `Foldable`/`Traverse`.

### DIFF
--- a/core/src/main/scala/scalaz/FreeT.scala
+++ b/core/src/main/scala/scalaz/FreeT.scala
@@ -63,11 +63,8 @@ object FreeT extends FreeTInstances {
 }
 
 sealed abstract class FreeT[S[_], M[_], A] {
-  final def map[B](f: A => B)(implicit S: Functor[S], M: Applicative[M]): FreeT[S, M, B] =
-    this match {
-      case Suspend(m) => Suspend(M.map(m)(_.bimap(f, S.lift(_.map(f)))))
-      case g @ Gosub() => gosub(g)(x => point(f(x)))
-    }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/scalaz/FreeT.scala
```diff
@@ -23,7 +23,11 @@ object FreeT extends FreeTInstances {
   /** Return the given value in the free monad. */
   def point[S[_], M[_], A](value: A)(implicit M: Applicative[M]): FreeT[S, M, A] = Suspend(M.point(-\/(value)))
 
-  def suspend[S[_], M[_], A](a: M[A \/ S[FreeT[S, M, A]]]): FreeT[S, M, A] = Suspend(a)
+  def suspend[S[_], M[_], A](a: M[A \/ S[FreeT[S, M, A]]])(implicit M: Applicative[M]): FreeT[S, M, A] =
+    liftM(a).flatMap({
+      case -\/(a) => point(a)
+      case \/-(s) => roll(s)
+    })
```
TomasMikula: `suspend` with this signature doesn't seem like a very useful smart constructor for the user. Perhaps it was never meant to be public? It used to mimic (leak?) the representation of `Suspend`, which this commit changes anyway. Should `suspend`, with this signature, be removed?

PR #1119, opened 2016-03-11
Added print trace feature to ShowOps

JohnReedLOL:
Print trace feature includes a stack trace in the print so that the user can trace the print.

_______________________________________________________________________________

**Example 1:**

`if( reallyImportantThing.trace ) `

**Output:**

```
"true" in thread eventThread:
  at path.to.file.MyClassName(MyClassName.scala:49)
```

^ The IDE will create a link to the location. User can click on the console instead of navigating.

_______________________________________________________________________________

**Example 2:**

`if( reallyImportantThing.trace2 ) `

**Output:**

```
"true" in thread eventThread:
  at path.to.file.MyClassName(MyClassName.scala:49)
  at path.to.file.ParentClass(ParentClass.scala:32)
```

_______________________________________________________________________________

**Example 3:**

`if( reallyImportantThing.trace0 ) `

**Output:**

`"true" in thread eventThread:`

_______________________________________________________________________________

**Example 4:**

`if( reallyImportantThing.traceN(6) ) `

**Output:**

```
"true" in thread eventThread:
  at path.to.file.MyClassName(MyClassName.scala:49)
  at path.to.file.ParentClass(ParentClass.scala:32)
  at path.to.file.ParentClass2(ParentClass2.scala:25)
  at path.to.file.ParentClass3(ParentClass3.scala:37)
  at path.to.file.ParentClass4(ParentClass4.scala:115)
  at path.to.file.ParentClass5(ParentClass5.scala:12)
```

_______________________________________________________________________________

**Example 5:**

`if( reallyImportantThing.traceN(-1111) ) `

**Output:**

`"true" in thread eventThread:`

### DIFF
--- a/core/src/main/scala/scalaz/syntax/ShowSyntax.scala
+++ b/core/src/main/scala/scalaz/syntax/ShowSyntax.scala
@@ -8,6 +8,79 @@ final class ShowOps[F] private[syntax](val self: F)(implicit val F: Show[F]) ext
   final def shows: String = F.shows(self)
   final def print: Unit = Console.print(shows)
   final def println: Unit = Console.println(shows)
+
+  /**
+    * 1 internal call, 1 thread call, and ShowOps.traceInternal are the three levels of stack trace offset.
+    */
+  private val stackOffset = 3
+
+  /**
+    * Prints this (self) and returns this
+    * @return this (self)
+    */
+  final def trace0: F = traceInternal(0)
+  /**
+    * Prints this (self) with a 1 line stack trace for easy locate-ability and returns this
+    * @return this (self)
+    */
+  final def trace1: F = traceInternal(1)
+  /**
+    * Prints this (self) with a 1 line stack trace for easy locate-ability and returns this (self). Short for "trace1"
+    * @return this (self)
+    */
+  final def trace: F = traceInternal(1)
+  /**
+    * Prints this (self) with a 2 line stack trace for easy locate-ability and returns this
+    * @return this (self)
+    */
+  final def trace2: F = traceInternal(2)
+  /**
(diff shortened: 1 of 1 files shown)


xuwei-k:
I think scalaz should not have these functions, because there are side effects.

JohnReedLOL:
But scallaz already has "reallyImportantThing.print" and "reallyImportantThing.println", and these things  have side effects of printing to the terminal. Not to mention that the [Debug.Trace](https://hackage.haskell.org/package/base-4.8.2.0/docs/Debug-Trace.html) feature is from Haskell and in Haskell it is useful for debugging and does not in any way reduce the correctness of the program or the ability of the user to reason about its correctness.

JohnReedLOL:
If Haskell can have _Debug.Trace.trace_ function as well as _Debug.Trace.traceStack_ function , then Scala can have _Debug.Trace.trace_ functionality too.

JohnReedLOL:
Not to mention that nothing is being modified. The original data it not modified. The name of the thread is same. The trace is getting a trace, but it is not changing. 

Finally, trace is temporary use feature. It is inserted into code, used, and then removed.

tonymorris:
Hi John,
The reasons for `trace` in Haskell are a little different to Scala. For example, it is much easier in Scala to break an expression apart and add a `println` before carrying on the computation. It is clear what the evaluation order will be, even give this side-effect.

As to how you contribute, simply suggest an idea and as you can see, it will be scrutinised. Rinse and repeat [and join in the learning].

PS: I had no idea those functions were on `Id` and I think they are a bug to be marked **SEVERE, NEVER REPEAT THIS EXERCISE, DELETE IMMEDIATELY**.

PR #1120, opened 2016-03-12
Update ISet based on Haskell v0.5.7.1 updates

monkey-mas:
Our ISet is based on v0.5.0.0. We can introduce new functions.

ref.
https://github.com/haskell/containers/compare/v0.5.0.0...v0.5.7.1?diff=split&name=v0.5.7.1#diff-13
https://github.com/haskell/containers/compare/v0.5.0.0...v0.5.7.1?diff=split&name=v0.5.7.1#diff-a9b15bd3edd743a14b46e25762f4fe8cR1122
https://github.com/haskell/containers/compare/v0.5.0.0...v0.5.7.1?diff=split&name=v0.5.7.1#diff-a9b15bd3edd743a14b46e25762f4fe8cR1481

### DIFF
--- a/core/src/main/scala/scalaz/ISet.scala
+++ b/core/src/main/scala/scalaz/ISet.scala
@@ -5,8 +5,8 @@ import Ordering._
 import std.option._
 
 /**
- * @see [[http://hackage.haskell.org/package/containers-0.5.0.0/docs/Data-Set.html]]
- * @see [[https://github.com/haskell/containers/blob/v0.5.0.0/Data/Set/Base.hs]]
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/ISet.scala
```diff
@@ -329,6 +329,54 @@ sealed abstract class ISet[A] {
         }
     }
 
+  final def splitRoot(): List[ISet[A]] =
```
xuwei-k: I think `final def splitRoot: List[ISet[A]]` is better

xuwei-k:
:+1:

monkey-mas:
Thanks for your review! :)

Issue #1123, opened 2016-03-15
Trampolined StreamT should not stack-overflow on foldLeft


TomasMikula:
`StreamT#foldLeft` overflows the stack, even when used on a `Trampoline`d stream. This is due to taking arguments by-name. I propose to fix this by changing the signature from

```scala
def foldLeft[B](z: => B)(f: (=> B, => A) => B)(implicit M: Monad[M]): M[B]
```

to

```scala
def foldLeft[B](z: B)(f: (B, A) => B)(implicit M: Monad[M]): M[B]
```

TomasMikula:
A failing test case is in #1124.

TomasMikula:
Implementation of the proposed fix is in #1125.

TomasMikula:
Fixed by #1125.

PR #1124, opened 2016-03-15
Test of StreamT#foldLeft stack safety

TomasMikula:
This PR is not meant to be merged as is. It provides a failing test case for #1123.


TomasMikula:
Merged as part of #1125.

PR #1126, opened 2016-03-15
Add StreamT#foldRightM, which is able to terminate on an infinite stream.

TomasMikula:
`StreamT#foldRight` always evaluates the whole stream. This version is able to terminate early.


xuwei-k:
There is a conflict, please rebase.

TomasMikula:
Thanks for pointing that out. Rebased.

xuwei-k:
:+1:

Issue #1122, opened 2016-03-15
Stop fucking breaking shit


tonymorris:
This is an issue for deleting all the completely useless broken shit in Scalaz. Most of it was committed between 2012 and 2014. Hurry up and delete it. It's urgent.

puffnfresh:
Some things on my list:

- [ ] Task (#1086)
- [ ] TaskApp
- [ ] Run
- [ ] Timer
- [ ] Strategy
- [ ] Nondeterminism
- [ ] Catchable

tonymorris:
I have looked at `Task`, `Strategy`, `Nondeterminism` and `Catchable` and I agree on all those.

aloiscochard:
:+1: you can remove all of those I won't mind at all :-)

makes sense

shajra:
I mentioned this in IRC.  If we go to an effect type in @aloiscochard's scato project, we might have yanked users from Task (7.2) back to IO (7.3) and then to something new again in Scalaz 8.  It might be just enough churn to lose projects like http4s.

So I've been lobbying to keep Task there until it's properly replaced.  @puffnfresh has some reservations about the implementation of the effect type in Scato (side-effects and casting), but since they don't leak, I'm not as concerned.  Perhaps it can be mended internally.  But @aloiscochard has been doing some good work trying to keep the implementation fast, which is one reason I'd prefer it be kept on the table as an option.

vmarquez:
Please do not yank Task until we have a suitable replacement.  It is very widely used.  

Nondeterminism is a useful, lawful type class that is buried in a lot of crap right now.  You  may have to squint to see it.  Don't throw the baby out with the bathwater though,  let's just clean that up.   

Why is catchable a problem?  It is lawful and I've made use of it before.  Is it redundant?

tpolecat:
@vmarquez `Catchable` is trying to abstract over data types that can observe nontermination, which is bogus ... it's a side-effect so `IO` should be the only data type that has anything to say about it. But I do agree that you need it if you have more than one `IO` (as `Task` has become) ... in fact you need something even more bogus (i.e., doobie's `Capture`).

Anyway I agree that we should get rid of this stuff, but we do need something like @puffnfresh's WIP on `forkIO` and so on, to support things people are now doing with `Nondeterminism`.

tpolecat:
Also is `Arrow` actually broken, or just not useful? Seems like it doesn't quite fit with the other things on the list. Might be worth moving to another issue.

[...]

PR #1128, opened 2016-03-15
revert deletion in favor of deprecation

tpolecat:
This reverts 8fd1b3b and adds deprecation notices. I think it's important not to break source compatibility without a deprecation cycle.

Many scalaz users (including me) need to interact with side-effect APIs and this method ends up being useful, but I'm fine removing it as long as there's some warning. Others may want to object generally.

### DIFF
--- a/core/src/main/scala/scalaz/syntax/IdOps.scala
+++ b/core/src/main/scala/scalaz/syntax/IdOps.scala
@@ -16,6 +16,24 @@ final class IdOps[A](val self: A) extends AnyVal {
   final def ▹[B](f: A => B): B =
     f(self)
 
+  /**Applies `self` to the provide function for its side effect, and returns `self`. The Kestrel combinator.
+   * Mostly for use with dodgy libraries that give you values that need additional initialization or
+   * mutation before they're valid to use.
+   *
+   * The name `tap` comes from the Ruby method: [[http://ruby-doc.org/core-2.0.0/Object.html#method-i-tap]]
+   * which allows you to "tap into" a method call chain, in order to perform operations on intermediate
+   * results within the chain.  `unsafe` because it enables side effects.
+   */
+  final def unsafeTap(f: A => Any): A = {
+    f(self); self
+  }
+
+  /** Alias for `unsafeTap`. */
+  final def <|(f: A => Any): A = unsafeTap(f)
+
+  /** Alias for `unsafeTap`. */
+  final def ◃(f: A => Any): A = unsafeTap(f)
+
   final def squared: (A, A) =
     (self, self)
 
--- a/core/src/main/scala/scalaz/syntax/IdOps.scala
+++ b/core/src/main/scala/scalaz/syntax/IdOps.scala
@@ -24,14 +24,17 @@ final class IdOps[A](val self: A) extends AnyVal {
    * which allows you to "tap into" a method call chain, in order to perform operations on intermediate
    * results within the chain.  `unsafe` because it enables side effects.
    */
+  @deprecated("will be removed in 7.4", "7.3")
   final def unsafeTap(f: A => Any): A = {
     f(self); self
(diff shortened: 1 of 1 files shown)


tonymorris:
I have heard it said that it is useful. After much discussion, no such demonstration of usefulness was forthcoming. What is this useful thing?

jedws:
:+1:

tonymorris:
Also, I think "deprecation cycles" are extremely silly, more so than the silly method itself.

tpolecat:
There are 365 instances in our codebase at work, almost all of a form like this, when interacting with old Java APIs that use construct-then-initialize:

```scala
val inst = new InstGmosNorth <| (_.setIssPort(IssPort.SIDE_LOOKING))
```

It's just syntax but it avoids an intermediate binding that might be a source of bugs.

```scala
val inst = {
  val tmp = new InstGmosNorth
  tmp.setIssPort(IssPort.SIDE_LOOKING)
  tmp
}
```

tonymorris:
If you go back to something like scalaz 2 or 3 or $WAYBACK, you'll find this method, and I am 100% certain I did not use `Any => Unit` because there is no way that signature belongs in working software, which has been a longstanding goal of mine.

Can we at least partially unfuck it with an appropriate type?

tonymorris:
er, I mean `A => Any`.

jedws:
presumably:

    implicit class IdOps[A](a: A) { def <|[B](f: A => B): A = { f(a); a } }

is preferred

tonymorris:
By the way, I do not agree that this function has improved a mitigation against a source of bugs. I think it has increased the probability of bugs by providing tools to masquerade as a function. That is, it is not just useless, but worse than the air it occupies. It is itself a bug, by this definition.

puffnfresh:
I think not having this function is better than having it and the deprecation allows bugs.

:-1:

bshelden:
Given the nature of this issue, I would argue this should be reverted, then re-applied as part of #1122.
#1122 should end up as the next breaking version of scalaz.
:+1:
That said, given the current version is in a very broken state, effort should be focused onto #1122 and this version left to rot.

nuttycom:
I don't think this requires a deprecation cycle; when I originally wrote this I had in mind something morally equivalent to `trace` and as a consequence removing it shouldn't break any meaningful code. This is really something that has no place in scalaz; if someone wants to use it they can do it in their own codebase.

coltfred:
:+1: - Giving users at least 1 version where it says "This was a horrible mistake" isn't too much to ask. Scalaz is used by thousands of projects and has historically done a great job of dealing with compatibility (thanks @xuwei-k).

wedens:
I use `<|` a lot in code that uses java apis. so, I'm in favour of deprecation.

aloiscochard:
I'm totally for the disappearance of such a thing, OTOH I can see that folks sadly rely on it today even though I don't.

So I'm fine with deprecation approach, maybe we should list somewhere the mistake we do not want to repeat for an hypothetical version 8 ?

S11001001:
I favor deprecation in 7.2.x, but deletion in 7.3.x, as opposed to this PR.  A 7.2.x release with the deprecation warning serves the desire for vocal notice to downstream; mitigation is sufficiently simple that the migration benefit outweighs the cost of keeping it longer.

:-1: as it stands (but see above)

vmarquez:
I agree with @S11001001. It makes the most sense, let's deprecate in the 7.2 release cycle and just remove it in 7.3.

xuwei-k:
>  let's deprecate in the 7.2 release cycle and just remove it in 7.3.

+1

coltfred:
I think @S11001001's idea is a good one. Does it address your concerns @tpolecat and @wedens?

tpolecat:
Closed in favor of #1132.

PR #1129, opened 2016-03-15
Add StreamT#asStream: conversion of StreamT[Id, A] to a lazy Stream[A].

TomasMikula:
`StreamT[M, A]#toStream` requires evaluation of the whole stream, either eagerly, or later (depending on the monad `M`). As a consequence, it cannot be used on an infinite `StreamT`.

This PR adds a method `asStream`, which works only if `M =:= Id`, but the resulting `Stream` is lazy.


xuwei-k:
:+1:

TomasMikula:
Thanks for merging! As a possible follow-up, I opened #1134.

PR #1130, opened 2016-03-16
Add stack-safe versions of StreamT[M, A] operations

TomasMikula:
`StreamT[M, A]` operations can be written in a stack-safe fashion when `M` is a tail-recursive monad (i.e. when `M[_]: BindRec`), without the need for trampolining the stream. This PR adds these stack-safe versions.

~~With each test I also added a negative test for the original operation that stack-overflows.~~ I deleted those negative tests, because I was not able to test for stack overflow in a platform-independent manner (`StackOverflowError` on JVM vs. `RangeError` in JS).

### DIFF
--- a/core/src/main/scala/scalaz/StreamT.scala
+++ b/core/src/main/scala/scalaz/StreamT.scala
@@ -15,16 +15,31 @@ sealed class StreamT[M[_], A](val step: M[StreamT.Step[A, StreamT[M, A]]]) {
        , skip = s => s.uncons
        , done = M.point(None)
        ))
+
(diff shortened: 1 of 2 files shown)


TomasMikula:
The JS tests failed with

> Exception: java.lang.AssertionError: wrong exception thrown, expected: class java.lang.StackOverflowError got: RangeError: Maximum call stack size exceeded

I guess I should just remove the negative tests?

TomasMikula:
Just a ping that this one is now merge-able.

xuwei-k:
https://github.com/xuwei-k/scalaz/commit/fa87dc6e79c9b84a77601f15b7435799bf8c08d0

TomasMikula:
Good point! I added your commit to this PR.

PR #1137, opened 2016-03-30
add StrictTree

shawjef3:
This is a strict version of Tree. Maybe some people would find it useful.

It's largely a copy of Tree, which I altered for maximum strictness.

### DIFF
--- a/core/src/main/scala/scalaz/StrictTree.scala
+++ b/core/src/main/scala/scalaz/StrictTree.scala
@@ -107,7 +107,7 @@ case class StrictTree[A](
   }
 
   /** Binds the given function across all the subtrees of this tree. */
-  def cobind[B](f: StrictTree[A] => B): StrictTree[B] = unfoldTree(this)(t => (f(t), () => t.subForest))
+  def cobind[B](f: StrictTree[A] => B): StrictTree[B] = unfoldTree(this)(t => (f(t), t.subForest))
 
   def foldNode[Z](f: A => Vector[StrictTree[A]] => Z): Z =
     f(rootLabel)(subForest)
@@ -241,7 +241,6 @@ sealed abstract class StrictTreeInstances {
    */
 }
 
-
 object StrictTree extends StrictTreeInstances {
   /**
    * Node represents a tree node that may have children.
@@ -276,12 +275,12 @@ object StrictTree extends StrictTreeInstances {
     }
   }
 
-  def unfoldForest[A, B](s: Vector[A])(f: A => (B, () => Vector[A])): Vector[StrictTree[B]] =
+  def unfoldForest[A, B](s: Vector[A])(f: A => (B, Vector[A])): Vector[StrictTree[B]] =
     s.map(unfoldTree(_)(f))
 
-  def unfoldTree[A, B](v: A)(f: A => (B, () => Vector[A])): StrictTree[B] =
+  def unfoldTree[A, B](v: A)(f: A => (B, Vector[A])): StrictTree[B] =
     f(v) match {
-      case (a, bs) => Node(a, unfoldForest(bs.apply())(f))
+      case (a, bs) => Node(a, unfoldForest(bs)(f))
     }
 
   //Only used for .equals.
--- a/core/src/main/scala/scalaz/StrictTree.scala
+++ b/core/src/main/scala/scalaz/StrictTree.scala
@@ -1,0 +1,464 @@
+package scalaz
+
+import scala.collection.mutable
+import std.vector.{vectorInstance, vectorMonoid}
+
+/**
+  *
+  * @param rootLabel The label at the root of this tree.
+  * @param subForest The child nodes of this tree.
+  * @tparam A
+  */
+case class StrictTree[A](
+  rootLabel: A,
+  subForest: Vector[StrictTree[A]]
+) {
+
+  import StrictTree._
+
+  /**
+    * Run a bottom-up algorithm.
+    *
+    * This is the framework for several stackless methods, such as map.
+    *
+    * @param reduce is a function from a label and its mapped children to the new result.
+    */
+  private[scalaz] def runBottomUp[B](
+    reduce: A => mutable.Buffer[B] => B
+  ): B = {
+    val root = BottomUpStackElem[A, B](None, this)
+    val stack = mutable.Stack[BottomUpStackElem[A, B]](root)
+
+    while (stack.nonEmpty) {
+      val here = stack.elems.head
+      if (here.hasNext) {
+        val child = here.next()
+        val nextStackElem = BottomUpStackElem[A, B](Some(here), child)
+        stack.push(nextStackElem)
+      } else {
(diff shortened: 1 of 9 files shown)


shawjef3:
Something that needs improving is the flatten method. The naive way, which is something like rootlabel +: subforest.flatMap(_.flatten), causes a stack overflow.

jbgi:
Would additional strict data constructors for Tree be a lighter solution?

core/src/main/scala/scalaz/StrictTree.scala
```diff
@@ -0,0 +1,208 @@
...
+    def alignWith[A, B, C](f: (\&/[A, B]) ⇒ C): (StrictTree[A], StrictTree[B]) => StrictTree[C] = {
+      def align(ta: StrictTree[A], tb: StrictTree[B]): StrictTree[C] =
+        StrictTree.Node(f(\&/(ta.rootLabel, tb.rootLabel)), Align[Vector].alignWith[StrictTree[A], StrictTree[B], StrictTree[C]]({
+          case \&/.This(sta) ⇒ sta map {a ⇒ f(\&/.This(a))}
+          case \&/.That(stb) ⇒ stb map {b ⇒ f(\&/.That(b))}
+          case \&/(sta, stb) ⇒ align(sta, stb)
+        })(ta.subForest, tb.subForest))
+      align _
+    }
+    def zip[A, B](a: StrictTree[A], b: StrictTree[B]): StrictTree[(A, B)] = {
```
xuwei-k: unused method?

shawjef3:
@jbgi I don't think strict constructors for Tree would be enough. Tree makes heavy use of lazy val, call by name, and Stream that would make it somewhat lazy even if the constructor wasn't..

tests/jvm/src/test/scala/scalaz/StrictTreeTestJVM.scala
```diff
@@ -0,0 +1,21 @@
+package scalaz
+
+import scalaz.scalacheck.ScalazArbitrary._
+import StrictTree._
+import org.scalacheck.Gen
+import org.scalacheck.Prop.forAll
+
+object StrictTreeTestJVM extends SpecLite {
+
+  "ScalazArbitrary.treeGenSized" ! forAll(Gen.choose(1, 200)){ size =>
```
xuwei-k: s/treeGenSized/strictTreeGenSized

core/src/main/scala/scalaz/StrictTree.scala
```diff
@@ -0,0 +1,208 @@
...
+        A.order(x.rootLabel, y.rootLabel) match {
+          case Ordering.EQ =>
+            Order[Vector[StrictTree[A]]].order(x.subForest, y.subForest)
+          case x => x
+        }
+    }
+
+  /* TODO
+  def applic[A, B](f: StrictTree[A => B]) = a => StrictTree.node((f.rootLabel)(a.rootLabel), implicitly[Applic[newtypes.ZipVector]].applic(f.subForest.map(applic[A, B](_)).?)(a.subForest ?).value)
+   */
```
aloiscochard: @shawjef3 do you plan to implement it as part of this PR, or maybe just stale dead code?

shawjef3: @aloiscochard It's a TODO for Tree, so I copied it to StrictTree. You'd have to ask @retronym what the status is.

aloiscochard:
@jbgi and even if we get the same semantic with strict constructor, the performance would be probably much different simply because of the lazy val (run-time check at read).

@shawjef3 Great stuff, thanks for the contribution!

just curious what was the original motivation, was it performance?

shawjef3:
@aloiscochard Yes, basically the motivation is performance.

shawjef3:
Something I'm leaving for a later PR is Tree#toStrictTree. I'm having a heck of a time implementing it without causing a stack overflow. Something about how I'm using Trampoline is incorrect.

shawjef3:
I added Tree#toStrictTree, and made sure that it works with deeply nested trees.

StrictTree#toTree had a performance problem with deeply nested trees due to a huge number of calls to Vector#flatten, so I changed it to an imperative implementation. Likewise for StrictTree#flatten.

I found that the Equals instance for Tree does not work for deeply nested trees, which causes some problems with some tests. I changed it to a trampolined implementation. Likewise for StrictTree's Equals.

shawjef3:
A lot of methods in StrictTree have problems with stack overflows for deep trees. I'm working on new implementations, but they can wait for a later release.

The only other thing I've noticed is that my implementation of loc was premature. It needs to be removed. People can still use StrictTree#toTree#loc. StrictTree#loc should be implemented later to return a strict version of a TreeLoc. I'll upload a new change in a few minutes

shawjef3:
I replaced several trampolined implementations with procedural implementations, yielding large speed increases.

Is there a way to squash this pull request without a force push?

shawjef3:
I think I'm done with this for now. traverse1 and cobind still can cause stack overflow exceptions, but they're going to be harder to convert to stackless implementations.

xuwei-k:
> Is there a way to squash this pull request without a force push?

There is no way. Please rebase and squash commits.

xuwei-k:
I think `BottomUpStackElem ` and `ZipStackElem` should define in companion object, not `class StrictTree`.

core/src/main/scala/scalaz/StrictTree.scala
```diff
@@ -0,0 +1,449 @@
...
+  /**
+    * This implementation is 24x faster than the trampolined implementation for StrictTreeTestJVM's hashCode test.
+    *
+    * @return
+    */
+  override def hashCode(): Int = {
+    runBottomUp(hashCodeCombiner)
+  }
+
+  override def equals(obj: scala.Any): Boolean = {
```
xuwei-k: unnecessary override?

shawjef3: The reason for this override is to avoid stack overflow on recursive calls to equals.

shawjef3:
@xuwei-k I moved the classes as you requested.

shawjef3:
I also moved AlignStackElem,

shawjef3:
What is keeping this from being accepted?

aloiscochard:
:+1:

shawjef3:
@aloiscochard @xuwei-k mind if I merge https://github.com/shawjef3/scalaz/commit/e9c833911dadffee5b0067adfb60a5ff448add66 into this pull request? Its purpose is to make StrictTree.unfoldTree and unfoldForest more strict. StrictTree doesn't benefit from the children being created by continuations.

aloiscochard:
@shawjef3 makes sense :+1:

TomasMikula:
Has anyone else experienced tests taking too long or even `OutOfMemoryError` during tests after this was merged?

vmarquez:
val size = 1000000



Might be a little overkill.

Issue #1105, opened 2016-02-18
Deadlock on 1-core virtual machine


raimohanska:
I got a nasty deadlock situation when using http4s client on a 1-core virtual machine. The problem is related to using then number of available processors as the size of the default thread pool in [Strategy](https://github.com/scalaz/scalaz/blob/series/7.3.x/concurrent/src/main/scala/scalaz/concurrent/Strategy.scala#L39).

The problem manifests as an application that has just stopped doing stuff. There are no error/warn messages anywhere, nor you can see a deadlock in a thread dump. I took me at least 8 hours to the get the bottom of this :(

I'll be happy to submit a PR for upping this default to at least 4 threads even if there's just a single CPU core available.

puffnfresh:
This code:

1. Uses implicits for configuration
2. Uses side-effects to create a default configuration
3. Is typed as `() => A`

There's not a lot going on for it. I've said this a few times, but I would like to get rid of `Strategy` and `Task`.

raimohanska:
I cannot debate on scalaz architecture yet, as I'm not fully familiar with it

But wouldn't it anyway be an improvement to use at least 2 threads instead of 1 in a single-core case? This is a practical thing that'll cause huge amounts of pain to newbies who cannot possibly guess that the default configuration is prone to deadlocks on a 1-core machine. Deadlocks that'll never occur on the newbie's 8-core laptop.

raimohanska:
I cannot see any possible downsides in defaulting to a threadpool with more than 1 thread.

jedws:
The problem is not so much that the defaults are wrong, but that defaulting
at all is the wrong thing to be doing.

pchlupacek:
@jedws in 7.2 is still the default strategy available? I think it shall be dropped from any next major release. In meantime i think min 4 threads would be completely sensible fix to me.

vmarquez:
I agree with getting rid of the defaults.

raimohanska:
I too agree with getting rid of the default. Submitted a PR for more convenient defaults to avoid nasty deadlocks on 1-core machines anyway. #1139

S11001001:
Fixed by #1139.

PR #1139, opened 2016-04-06
Default executor uses at least 4 threads

raimohanska:
Related to https://github.com/scalaz/scalaz/issues/1105, here's a PR for making the default executor thread pool size at least 4, to avoid nasty deadlocks on 1-core machines.

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Strategy.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Strategy.scala
@@ -36,7 +36,7 @@ trait Strategys extends StrategysLow {
    * where N is equal to the number of available processors.
    */
   val DefaultExecutorService: ExecutorService = {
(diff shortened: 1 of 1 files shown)


puffnfresh:
I think a static 4 would be more useful than using `availableProcessors` at all.

raimohanska:
@puffnfresh might be so. Shall I change it to 4 to get this merged in soon? :)

runarorama:
Looks good to merge as is.

aloiscochard:
Is there a specific reason to use `4` and not `2`?

aloiscochard:
@pchlupacek I see that you proposed `4` in the other issue, maybe you could motivate why 4 instead of 2?

pchlupacek:
@aloiscochard I think 2 are just not enough. Easily you can run off resources. Hardly you can with 4 and more.

Issue #615, opened 2013-12-27
Stream traverse foldMap regression?


wookietreiber:
I think, I have stumbled across a regression concerning Stream/traverse/foldmap. See the following diff:

https://github.com/wookietreiber/scala-hash/commit/85b8ae2cac75d208eac19a1af0ee38f6a5e2828c

The simple foldLeft (new) works as one would expect, but the foldMap variant (old) seems to run forever (I killed the sbt test process after a few minutes) ...

BTW, I bumped scalaz version a few commits before that, see https://github.com/wookietreiber/scala-hash/commit/b363abaeac3b62ec5dabb6dfe99b3d469cf45465, it was 7.0.3 where everything still worked fine with foldMap. To narrow it down, I also checked 7.0.4 and the problem appears with this version, too, so the regression must have been between 7.0.3 and 7.0.4. I hope this helps.

edmundnoble:
Looks to me like we just need to add `foldMapLeft` to `Foldable`. For consistency I think we should also have a monoidal `fold` which associates to the left, but I'm not sure what to call it :wink: (`mfoldl`?). There are a lot of places in Scalaz (for example `Cord.toVector`, some parts of `Tree`) which could benefit from this, because their monoids are certainly not lazy (and this option should also be in `Reducer`, under `foldReduceLeft`). I'll get on making a PR.

PR #1143, opened 2016-04-13
Stream foldMap implementation should not overflow the stack.

GodPlaysChess:
Former `foldMap` implementation blowed the stack because of the `foldRight`. Actually I am thinking, that `foldRight` itself should be implemented in a safe way: 
`override def foldRight[A, B](fa: Stream[A], z: => B)(f: (A, => B) => B): B =   this.foldLeft(fa.reverse, z)((b, a) => f(a, b))` , but then I hit failing test:
scalaz.std.StreamTest.:foldRight evaluates lazily: Exception raised on property evaluation.
Do you think the `foldRight` on Stream should really evaluate lazily?


xuwei-k:
see https://github.com/scalaz/scalaz/issues/615

Issue #1144, opened 2016-04-14
Lazy vals in type classes lock on the whole instance


runarorama:
E.g. in `Bind.scala`:

``` scala
override def ap[A, B](fa: => F[A])(f: => F[A => B]): F[B] = {
  lazy val fa0 = fa
  bind(f)(map(fa0))
}
```

While `fa` is being evaluated, the evaluating thread occupies the monitor for `this` (the `Bind` instance), and other threads entering `ap` must wait. This is especially egregious for monads like `Future` and `Task` where you might want to have multiple threads calling `ap` at the same time.

Possible fix:

``` scala
override def ap[A, B](fa: => F[A])(f: => F[A => B]): F[B] = {
  val monitor = new Object { lazy val fa0 = fa }
  bind(f)(map(monitor.fa0))
}
```

In this specific case, I'm not sure it's even necessary to cache the `fa` value since `map` is strict. There are other instances of this though. For example, `Monad.whileM`.

xuwei-k:
use `scalaz.Need` instead of `new Object{` ?

runarorama:
Not possible in every case, but yeah, I'm currently going through and replacing `lazy val` with `Need`.

jedws:
I think the title should be changed, the problem is the lazy val, not the by name argument.

aloiscochard:
Nice catch @runarorama, and what a terrible scala gotcha :-(

runarorama:
@jedws The title says "by need", not "by name".

alexandru:
Hi, I noticed this thread on Twitter, sorry for adding my opinion, couldn't help it :-P

Personally I already knew how `lazy val` behaves, because its behavior can't come for free and I checked out the generated bytecode a long time ago. Checking out the generated bytecode is a healthy habit, with the `javap` utility being available and easy to use. And due to the `lazy val` constraints, the current implementation is very reasonable:

1. `lazy val` has guaranteed idempotency in the presence of multi-threading and effects, or in other words the lazy expression will be evaluated only once. And in order to guarantee idempotency in this case, you can't escape the use of a lock. It's impossible. In other instances, like when you've got an idempotent `() => Unit`, you might escape with a simple `getAndSet`, but not this one.
[...]

PR #1145, opened 2016-04-15
Replacing lazy vals with `Need`

runarorama:
Addresses #1144

### DIFF
--- a/core/src/main/scala/scalaz/Name.scala
+++ b/core/src/main/scala/scalaz/Name.scala
@@ -52,13 +52,31 @@ object Name {
   }
 }
 
-object Need {
+/** Synchronized, thread-safe version of `Need` */
+object ThreadSafeNeed {
   def apply[A](a: => A): Need[A] = {
     new Need[A] {
-      private[this] lazy val value0: A = a
-      def value = value0
+      lazy val value: A = a
     }
   }
+}
+
+object Need {
+  def apply[A](can: => A): Need[A] = {
+    new Need[A] {
+      private[this] var a: A = _
+      private[this] var haz: Boolean = _
+      def value = {
+        if (!haz) {
+          a = can
+          haz = true
+        }
+
+        a
+      }
+    }
+  }
+
   def unapply[A](x: Need[A]): Option[A] = Some(x.value)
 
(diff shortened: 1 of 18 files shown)


tests/jvm/src/test/scala/scalaz/NeedTestJVM.scala
```diff
@@ -5,9 +5,9 @@ object NeedTestJVM extends SpecLite {
   "Need" should {
     "clear the Function0 reference" in {
       @volatile var flag = false
-      val method = Need.getClass.getMethod("apply", classOf[Function0[_]])
+      val method = ThreadSafeNeed.getClass.getMethod("apply", classOf[Function0[_]])
```
runarorama: Note that this test does not pass with `Need` instead of `ThreadSafeNeed`

core/src/main/scala/scalaz/Name.scala
```diff
@@ -52,13 +52,31 @@ object Name {
...
   def apply[A](a: => A): Need[A] = {
     new Need[A] {
-      private[this] lazy val value0: A = a
-      def value = value0
+      lazy val value: A = a
+    }
+  }
+}
+
+object Need {
```
runarorama: Changed the implementation of `Need` to not use a lazy val. `ThreadSafeNeed` uses a `lazy val` to synchronize the memo.

tonymorris:
:+1:

core/src/main/scala/scalaz/Name.scala
```diff
@@ -52,13 +52,31 @@ object Name {
...
 }
 
-object Need {
+/** Synchronized, thread-safe version of `Need` */
+object ThreadSafeNeed {
   def apply[A](a: => A): Need[A] = {
     new Need[A] {
-      private[this] lazy val value0: A = a
-      def value = value0
+      lazy val value: A = a
```
xuwei-k: I think `private[this]` is important if we want clear reference of the thunk `a: => A`. see https://github.com/scalaz/scalaz/pull/422.
Also, There is another way https://github.com/xuwei-k/scalaz/commit/e52825eb3d

runarorama: This test passes, without the `private [this]`:

https://github.com/scalaz/scalaz/pull/1145/files#diff-2edaebab281ed9a80e0bfd3ea3d54db8R5

aloiscochard:
:+1:

xuwei-k:
:+1:

PR #1151, opened 2016-04-22
change EphemeralStream.foldLeft to use a tail recursive loop

purefn:
Prior to this, running the code below resulted in a runtime exception saying that the head of an empty stream was being accessed

```
val es = EphemeralStream.unfold(0)(n => if (n < 5000) Some((List.fill(10000)(500L), n+1)) else None)
es.foldLeft( ().point[IO])(io => e => io >> IO.putStrLn(e.length.shows)).unsafePerformIO
```

Note, the above still doesn't work properly. It results in an OOME, as does 

```
es.traverse_(e => IO.putStrLn(e.length.shows))
```

which is what I'm really trying to get working properly. 

Also of note is that

```
es.foldLeft( ().point[IO])(io => e => io *> IO.putStrLn(e.length.shows)).unsafePerformIO
```

DOES work. But, if you inline `*>`

```
es.foldLeft( ().point[IO])(io => e => Apply[IO].apply2(io, IO.putStrLn(e.length.shows))((_, _) => ())).unsafePerformIO
```

it does NOT work, resulting in an OOME again.

I'm still trying to puzzle this out and get `traverse_` working as expected. I will be eternally grateful to anyone that can provide some insight here.

### DIFF
--- a/core/src/main/scala/scalaz/EphemeralStream.scala
+++ b/core/src/main/scala/scalaz/EphemeralStream.scala
@@ -38,13 +38,11 @@ sealed abstract class EphemeralStream[A] {
     if (isEmpty) z else f(head())(tail().foldRight(z)(f))
 
   def foldLeft[B](z: => B)(f: (=> B) => (=> A) => B): B = {
-    var t = this
-    var acc = z
-    while (!t.isEmpty) {
-      acc = f(acc)(t.head())
-      t = t.tail()
-    }
-    acc
+    @annotation.tailrec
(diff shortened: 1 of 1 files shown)


xuwei-k:
LGTM. Could you add some tests?

xuwei-k:
added test https://github.com/scalaz/scalaz/commit/bc410b3e176f19c5a0db0c

purefn:
@xuwei-k sorry I didn't get to writing the test this weekend. Thanks for doing it!

PR #1153, opened 2016-04-26
IndexedStateT is a lawful Profunctor!

vmarquez:


TomasMikula:
This PR gives a Profunctor in `[S1, A]`. Isn't `IndexedStateT` also a Profunctor in `[S1, S2]`? Is `[S1, A]` version more useful than `[S1, S2]`? I can see that `[S1, A]` is compatible with the default `Functor` instance. OTOH, `[S1, S2]` seems more "symmetric".

S11001001:
Here I think the `Functor` consistency is more important than the symmetry of the states version, especially in terms of usability.

PR #1155, opened 2016-05-02
Add FreeT#foldMap.

TomasMikula:
This method is analogous to `Free#foldMapRec`.

As opposed to `FreeT#runM`, it doesn't require `Functor[S]`.

### DIFF
--- a/core/src/main/scala/scalaz/FreeT.scala
+++ b/core/src/main/scala/scalaz/FreeT.scala
@@ -107,6 +107,29 @@ sealed abstract class FreeT[S[_], M[_], A] {
   @deprecated("Alias for `interpret`", "7.3")
   def interpretT[T[_]](st: S ~> T)(implicit M: Functor[M]): FreeT[T, M, A] = interpret(st)
 
+  /**
+    * Runs to completion, mapping the suspension with the given transformation
+    * at each step and accumulating into the monad `M`.
+    */
+  def foldMap(f: S ~> M)(implicit M0: BindRec[M], M1: Applicative[M]): M[A] = {
+    def go(ft: FreeT[S, M, A]): M[FreeT[S, M, A] \/ A] =
+      ft match {
+        case Suspend(ma) => M0.bind(ma) {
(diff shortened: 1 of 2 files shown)


TomasMikula:
There's a spurious build error. Can someone please restart the build?

core/src/main/scala/scalaz/FreeT.scala
```diff
@@ -107,6 +107,29 @@ sealed abstract class FreeT[S[_], M[_], A] {
   @deprecated("Alias for `interpret`", "7.3")
   def interpretT[T[_]](st: S ~> T)(implicit M: Functor[M]): FreeT[T, M, A] = interpret(st)
 
+  /**
+    * Runs to completion, mapping the suspension with the given transformation
+    * at each step and accumulating into the monad `M`.
```
jedws: `M` is not a Monad, probably should just say "accumulating into `M`"

TomasMikula: `Bind(Rec)` and `Applicative` give a monad, so it is not completely wrong.

I was wondering, maybe it's worth it to introduce `MonadRec` type class to `scalaz` instead of requiring `BindRec` + `Applicative`?

TomasMikula:
Any reviews on this one? Perhaps @puffnfresh?

`foldMap` is probably the most useful method for interpreting free monads, so I would like to get this in.

core/src/main/scala/scalaz/FreeT.scala
TomasMikula: @jedws was that a satisfactory answer?

xuwei-k:
👍

core/src/main/scala/scalaz/FreeT.scala
puffnfresh: @TomasMikula a MonadRec could be useful but I think the effort might not be worth it.

puffnfresh:
I like it :+1:

PR #1160, opened 2016-05-16
Add Traverse type class instance for Free.Source

Atry:


Atry:
Hi @xuwei-k ,

The type class itself works for me. Unfortunately, when I use this type class, I experienced some unapply error:

> Implicit not found: scalaz.Unapply[scalaz.Foldable, scalaz.Free[[β](String, β),Unit]]. Unable to unapply type `scalaz.Free[[β](String, β),Unit]` into a type constructor of kind `M[_]` that is classified by the type class `scalaz.Foldable`. Check that the type class is defined by compiling `implicitly[scalaz.Foldable[type constructor]]` and review the implicits in object Unapply, which only cover common type 'shapes.'

Sorry I am not familiar with the unapply syntax. I wonder if you could add some unapply instances corresponds to this `Traverse` instance?

Atry:
Ping @xuwei-k

vmarquez:
this looks 👍 from me.  @xuwei-k what do you think?

PR #1162, opened 2016-05-16
TheseT monad transformer (plus IO instances and tests)

oxbowlakes:
TheseT is a monad transformer for `\&/` (pronounced *these*). 

  - `A \&/ B` represents having *either* an `A`, *or* a `B`, or *both*
  - the monad transformer `TheseT[M, A, B]` is essentially an ops class for a `M[A \&/ B]`. TheseT is both a monad and it supports many of the operations one finds on `\&/`


oxbowlakes:
Not sure what this failure is - it doesn't look obvious that it's actually anything to do with the PR (appears to be in MiMa and something to do with the JS backend)

scalacheck-binding/src/main/scala/scalaz/scalacheck/ScalazArbitrary.scala
```diff
@@ -256,6 +256,9 @@ object ScalazArbitrary {
   implicit def eitherTArb[F[+_], A, B](implicit A: Arbitrary[F[A \/ B]]): Arbitrary[EitherT[F, A, B]] =
       Functor[Arbitrary].map(A)(EitherT[F, A, B](_))
 
+  implicit def theseTArb[F[+_], A, B](implicit A: Arbitrary[F[A \&/ B]]): Arbitrary[TheseT[F, A, B]] =
```
xuwei-k: `implicit def theseTArb[F[_], A, B]`

core/src/main/scala/scalaz/TheseT.scala
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  implicit def theseTSemigroup[F[_]: Apply, A: Semigroup, B: Semigroup]: Semigroup[TheseT[F, A, B]] = new Semigroup[TheseT[F, A, B]] {
+    override def append(f1: TheseT[F, A, B], f2: => TheseT[F, A, B]) = TheseT(Apply[F].apply2(f1.run, f2.run)(_ append _))
+  }
+
+}
+sealed abstract class TheseTInstances extends TheseTInstances0
+
+trait TheseTFunctions {
```
xuwei-k: please remove `TheseTFunctions` trait. just use companion object https://github.com/scalaz/scalaz/pull/995

core/src/main/scala/scalaz/TheseT.scala
```diff
@@ -0,0 +1,148 @@
...
+
+sealed abstract class TheseTInstances1 {
+  implicit def TheseTInstance1[F[_]: Traverse, L]: Traverse[TheseT[F, L, ?]]
+  = new Traverse[TheseT[F, L, ?]] {
+    override def traverseImpl[G[_], A, B](fa: TheseT[F, L, A])(f: (A) => G[B])(implicit A: Applicative[G]) =
+      fa.traverse(f)
+
+  }
+
+  /* cannot use kind projector */
```
xuwei-k:
```scala
implicit def TheseTHoist[A: Semigroup]: Hoist[TheseT[?[_], A, ?]] = new Hoist[TheseT[?[_], A, ?]] {
```

https://github.com/non/kind-projector/commit/cb61341983263ea8bee

oxbowlakes: That doesn't compile

    info] Compiling 1 Scala source to C:\Users\cmarsha\Work\temp\scalaz\core\target\scala-2.10\classes...
    error] C:\Users\cmarsha\Work\temp\scalaz\core\src\main\scala\scalaz\TheseT.scala:101: not found: type ?
    error]   implicit def TheseTHoist[A: Semigroup] = new Hoist[TheseT[?[_], A, ?]] {
    error]                                                             ^
    error] [X_kp2]scalaz.TheseT[<error>,A,X_kp2] takes one type parameter, expected: two
    error] [X_kp2]scalaz.TheseT[<error>,A,X_kp2] takes one type parameter, expected: two
    error] three errors found
    error] (core/compile:compile) Compilation failed

scalacheck-binding/src/main/scala/scalaz/scalacheck/ScalazArbitrary.scala
oxbowlakes: I have made this change

core/src/main/scala/scalaz/TheseT.scala
oxbowlakes: I have made this change as requested

core/src/main/scala/scalaz/TheseT.scala
xuwei-k: because your branch is too old. please rebase the upstream

oxbowlakes: I have rebased and pushed.

vmarquez:
👍🏼

core/src/main/scala/scalaz/TheseT.scala
```diff
@@ -0,0 +1,146 @@
...
+  = G.map(F.traverse(run)(o => Traverse[AA \&/ ?].traverse(o)(g)))(TheseT(_))
+
+
+  def bitraverse[G[_], C, D](f: A => G[C], g: B => G[D])(implicit F: Traverse[F], G: Applicative[G]): G[TheseT[F, C, D]]
+  = G.map(F.traverse(run)(Bitraverse[\&/].bitraverseF(f, g)))(TheseT(_: F[C \&/ D]))
+
+  def &&&[AA >: A, C](t: TheseT[F, AA, C])(implicit M: Semigroup[AA], F: Apply[F]): TheseT[F, AA, (B, C)]
+  = TheseT(F.apply2(run, t.run)(_ &&& _))
+
+  def ===[AA >: A, BB >: B](x: TheseT[F, AA, BB])(implicit EA: Equal[AA], EB: Equal[BB], F: Apply[F]): F[Boolean]
```
xuwei-k: I think we should define `Equal[TheseT[F, A, B]` instance and then remove this method. WDYT?

tonymorris: I agree, otherwise :+1:

oxbowlakes: `EitherT` defines both `===` as an instance method on `EitherT` (along the lines of the above) and also a `Equal` instance (in `EitherTInstances`) **[1]**. This is presumably a mistake **[2]** and I took the view that the former was more useful because the most common value for `F` is presumably `IO` (or a stack with `IO` in it) and thus there will be no implicit `Equal[EitherT[F, A, B]]`

I do agree that the instance method `===` with a return type other than `Boolean` is a bit iffy; I was just trying to keep `TheseT` as close to `EitherT` as possible

**[1]** - 

    implicit def eitherTEqual[F[_], A, B](implicit F0: Equal[F[A \/ B]]): Equal[EitherT[F, A, B]] 
      = F0.contramap((_: EitherT[F, A, B]).run)

**[2]** - just because `et1 === et2` is never using the typeclass but the instance method

oxbowlakes: I did as you ask because I just can't take the silent treatment anymore

xuwei-k: sorry for the delay. https://github.com/scalaz/scalaz/pull/1162/commits/75935e138206a8fbf is LGTM

tests/src/test/scala/scalaz/TheseTTest.scala
```diff
@@ -0,0 +1,26 @@
...
+import std.AllInstances._
+import org.scalacheck.Prop.forAll
+
+
+object TheseTTest extends SpecLite {
+  type TheseTList[A, B] = TheseT[List, A, B]
+  type TheseTListInt[A] = TheseT[List, Int, A]
+  type TheseTOptionInt[A] = TheseT[Option, Int, A]
+
+  implicit def TheseTListIntEqual[A: Equal]: Equal[TheseTListInt[A]] = new Equal[TheseTListInt[A]] {
```
xuwei-k: I think this is unnecessary because now there is `implicit def theseTEqual` in `ThereT.scala`.

PR #1166, opened 2016-05-24
Add parseBigInt and parseBigDecimal to std.String

mossprescott:
We need these two types in quasar (see https://github.com/quasar-analytics/quasar/pull/1193).

Also switched them to not use try/catch directly, but could take or leave that change.


mossprescott:
Ok, so this failed the binary compatibility check because it adds two methods. Where does that leave me?

xuwei-k:
Can't keep binary compatibility. please submit to `series/7.3.x` branch

PR #1167, opened 2016-05-26
Add parseBigInt and parseBigDecimal to std.String

mossprescott:
We need these two types in quasar (see https://github.com/quasar-analytics/quasar/pull/1193).

Also switched them to not use try/catch directly, but could take or leave that change.

Resubmitted against series/7.3.x. See #1166.

### DIFF
--- a/core/src/main/scala/scalaz/std/String.scala
+++ b/core/src/main/scala/scalaz/std/String.scala
@@ -79,6 +79,18 @@ trait StringFunctions {
   } catch {
     case e: NumberFormatException => Failure(e)
   }
+
+  def parseBigInt(s:String): Validation[NumberFormatException, BigInt] = try {
+    Success(BigInt(s))
+  } catch {
+    case e: NumberFormatException => Failure(e)
+  }
+
+  def parseBigDecimal(s:String): Validation[NumberFormatException, BigDecimal] = try {
+    Success(BigDecimal(s))
+  } catch {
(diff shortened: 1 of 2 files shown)


puffnfresh:
:+1:

core/src/main/scala/scalaz/std/String.scala
```diff
@@ -38,47 +38,32 @@ trait StringFunctions {
 
   // Parsing functions.
 
-  def parseBoolean(s:String): Validation[IllegalArgumentException, Boolean] = try {
-    Success(s.toBoolean)
-  } catch {
-    case e: IllegalArgumentException => Failure(e)
-  }
+  def parseBoolean(s:String): Validation[IllegalArgumentException, Boolean] =
+    \/.fromTryCatchThrowable[Boolean, IllegalArgumentException](s.toBoolean).validation
```
xuwei-k: There is `fromTryCatchThrowable` in Validation as well.

```scala
Validation.fromTryCatchThrowable[Boolean, IllegalArgumentException](s.toBoolean)
```

https://github.com/scalaz/scalaz/blob/v7.3.0-M2/core/src/main/scala/scalaz/Validation.scala#L390

mossprescott:
Thanks, @xuwei-k!

I went ahead and squashed that fix because it seemed the obvious thing. If you'd prefer, I can squash the two that remain.

PR #1169, opened 2016-05-26
Lazy non-empty right fold for streams; more efficient left fold too

S11001001:
Lifting a `Semigroup` into `Option` to get a `Monoid` unfortunately makes an always-strict `mappend`, meaning a strict fold.  Here we avoid that for `Stream` and `EphemeralStream`'s `foldMapRight1Opt` by taking advantage of their structure.


xuwei-k:
👍

PR #1171, opened 2016-05-27
Improve IMap based on haskell/containers v0.5.7.1 updates

monkey-mas:
Our implementation is based on v0.4.0. Since haskell's Data.Map of v0.5.7.1 has a lot of improvement compared to that of v0.4.0, we can expect a better result by catching up the latest haskell's implementation.

### 1. Modify a re-balancing parameter `delta` from 4 to 3
As Milan Straka(a.k.a foxik from haskell/containers) and Kazu Yamamoto discussed [1, 2], we can expect insertion speed-up and deletion slow-down by changing the value. Actually only valid delta value with `ratio = 2` is 3 or 4 as proved by Yoichi Hirai and Kazuhiko Yamamoto, and Milan Straka[3, 4].
    One justification to set 3, I think, is we'd expect more insertion operations than deletion operations. Even though deletion is executed more than the number of times insertion is called, it doesn't really hurt us as it means we do deletion onto an empty map which indicates `O(1)`.
    [1] [Proposal: Further performance improvements of Data.Map by Kazu](https://mail.haskell.org/pipermail/libraries/2010-September/014254.html)
    [2] [Proposal: Further performance improvements of Data.Map by Milan](https://mail.haskell.org/pipermail/libraries/2010-September/014343.html)
   [3] [Balancing weight-balanced trees](https://yoichihirai.com/bst.pdf)
   [4] [Adams' Trees Revisited - Correct and Efficient Implementation](http://fox.ucw.cz/papers/bbtree/)

###  2. Apply the `correct` algorithm for key comparison
As Kazu Yamamoto pointed it out[1], we shouldn't use equality to check the validity of the size of subtrees. Milan told me the reason as below why equality was used when I fixed the old documentation of Data.Map. You can also find [the comment](https://github.com/haskell/containers/pull/191#issuecomment-199256260) from [my PR](https://github.com/haskell/containers/pull/191) :)
   - `The reason why (<=) had to be used before was that with delta=5 using (<) can be shown easily to not keep tree balance. Hence whoever implemented it, tried (<=), and the QuickCheck could not find a problem -- which happened several years later in [#4242](https://ghc.haskell.org/trac/ghc/ticket/4242).`

This change actually leads us to slow down deletion operation speed like you can find the benchmark result of [this commit](https://github.com/monkey-mas/scalaz/commit/da1912304799d44b3be12e7165bbfd7ec0387c52). I, however, think this is acceptable as the updated implementation follows the correct (`f-balance`) algorithm as mentioned by Kazu[1].

### 3. Improve functionality including tree re-balancing
As you can see from the benchmark results below, most of the functionality was improved.

###  4. Deprecate some functions
I deprecated some functions that used to be public functions and should've been treated as `private` functions. I did this as some functions are exposed to users even though they shouldn't. I talked about it with @xuwei-k, and this might not be a good idea as this breaks binary-compatibility. we might need discussion?

###  5. Add new functions
I added some new functions introduced (and "property-based" tests as well ;) )

## Benchmark
To justify these changes with regard to the performance, benchmark(average execution time) was measured using [OpenJDK: jmh](http://openjdk.java.net/projects/code-tools/jmh/) 

### Setup information
```
OS: Ubuntu 14.0.4 (with Linux kernel version 3.13.0-86-generic)
CPU: Intel(R) Core(TM) i7-2600 CPU @ 3.40GHz
VM version: JDK 1.8.0_77, VM 25.77-b03
JVM heap size: 8G(8192M)
sbt-plugin: [sbt-jmh](https://github.com/ktoso/sbt-jmh) (0.2.6(auto plugin) / OpenJDK:jmh version 1.11.3)
Warm-up trial: 30 times
Actual trial: 50 times
```

### Benchmark strategy
`insert`:
  - `Asc` means that we insert elements starting from 1 to one million.
  - `Dsc` means that we insert elements starting from one million to 1.
  - `Rand` means that we insert elements chosen from a pseudo-randomly shuffled list with one million elements.

`delete`:
  - `Asc` means that we delete elements starting from 1 to one million from a pseudo-randomly shuffled list with one million elements.
  - `Dsc` means that we delete elements starting from one million to 1 a pseudo-randomly shuffled list with one million elements.
  - `Rand` means that we delete elements chosen from a pseudo-randomly shuffled list with one million elements.

`union`:
  - `E_O` means that we do operations using two maps, one of which contains even elements from 2 to one million and the other of which has odd elements from 1 to one million.
  - `Rand` means that we do operations using two maps, one of which contains a pseudo-randomly shuffled list(say `a`) with 500,000 elements chosen from 1 to one million and the other of which(say `b`) has elements that `a` does not have.
  - `Rand2` means that we union b and a unlike we do `union a and b` in `Rand` benchmark.

`difference / intersection`:
  - `E_O` means that we do operations using two maps, one of which contains even elements from 2 to one million and the other of which has elements from 1 to one million.
  - `Rand` means that we do operations using two maps, one of which contains a pseudo-randomly shuffled list with 500,000 elements chosen from 1 to one million.
  - `Rand2` means that we do operations using two maps, one of which contains a pseudo-randomly shuffled list with 500,000 elements that are not used in `Rand` sample data.

### Benchmark results
  - `Original` refers to the latest commit [5efbdbd](https://github.com/monkey-mas/scalaz/commit/5efbdbd95968dd939df401f628360ec09c1b0015) of `series/7.3.x`.
  - `IMap_rm_eq` refers to the commit after we changed [delta to 3](https://github.com/monkey-mas/scalaz/commit/ecadcc866ea9944fcaf12941d19e2c4b589215f6) and [key-comparison](https://github.com/monkey-mas/scalaz/commit/da1912304799d44b3be12e7165bbfd7ec0387c52).
  - `IMap_balanceLR` refers to [the commit](https://github.com/monkey-mas/scalaz/commit/497a24490d6c764e7d5e2a9436e47160bbe0fbe1) after we improved our balance functions.
  - `IMap_latest` refers to [the latest commit](https://github.com/scalaz/scalaz/pull/1171/commits/73f6a086fcc415461e019005ffa37a0df9f05ac0) of this PR where all the changes are applied.

You can see the actual benchmark sample data and code from [here](https://github.com/monkey-mas/scalaz/blob/compare-map-benchmark/src/main/scala/jmh/MapBenchmark.scala) and results from [here](https://github.com/monkey-mas/scalaz/tree/compare-map-benchmark/result)
To execute the benchmark that I've done, please try the command below.
```
sbt -mem 8192 "jmh:run -i 50 -wi 30 -f 1 -t 1 -bm avgt"
```

Following are the benchmark result bar charts:

insertion
delete
difference
intersection
union

### DIFF
--- a/core/src/main/scala/scalaz/Map.scala
+++ b/core/src/main/scala/scalaz/Map.scala
@@ -76,7 +76,7 @@ sealed abstract class ==>>[A, B] {
     }
 
   /** alias for [[delete]] */
-  def -(k: A)(implicit o: Order[A]) =
+  def -(k: A)(implicit o: Order[A]): A ==>> B =
     delete(k)
 
   /** removes a key/value pair - O(log n) */
@@ -225,7 +225,7 @@ sealed abstract class ==>>[A, B] {
     case Bin(k,v,l,r) => ISet.Bin(k,l.keySet,r.keySet)
   }
 
-  def toList =
+  def toList: List[(A, B)] =
     toAscList
 
   def toAscList: List[(A, B)] =
@@ -234,10 +234,10 @@ sealed abstract class ==>>[A, B] {
   def toDescList: List[(A, B)] =
     foldlWithKey(List.empty[(A, B)])((xs, k, x) => (k, x) :: xs)
 
-  def member(k: A)(implicit n: Order[A]) =
+  def member(k: A)(implicit n: Order[A]): Boolean =
     lookup(k)(n).isDefined
 
-  def notMember(k: A)(implicit n: Order[A]) =
+  def notMember(k: A)(implicit n: Order[A]): Boolean =
     !member(k)
 
   def lookupIndex(k: A)(implicit o: Order[A]): Option[Int] = {
@@ -295,7 +295,7 @@ sealed abstract class ==>>[A, B] {
         }
     }
 
-  def deleteAt(i: Int) =
+  def deleteAt(i: Int): A ==>> B =
     updateAt(i, (A, B) => None)
 
   @tailrec
@@ -358,7 +358,7 @@ sealed abstract class ==>>[A, B] {
         empty
     }
 
-  def updateMax(f: B => Option[B]) =
+  def updateMax(f: B => Option[B]): A ==>> B =
     updateMaxWithKey((_: A, b) => f(b))
 
   def updateMaxWithKey(f: (A, B) => Option[B]): A ==>> B =
@@ -380,7 +380,7 @@ sealed abstract class ==>>[A, B] {
     * insert v into the map at k. If there is already a value for k,
     * append to the existing value using the Semigroup
     */
-  def updateAppend(k: A, v: B)(implicit o: Order[A], bsg: Semigroup[B]) =
+  def updateAppend(k: A, v: B)(implicit o: Order[A], bsg: Semigroup[B]): A ==>> B =
     alter(k, old ⇒ Some(old.map(bsg.append(_, v)).getOrElse(v)))
 
   def minViewWithKey: Option[((A, B), A ==>> B)] =
@@ -527,10 +527,10 @@ sealed abstract class ==>>[A, B] {
     }
   }
 
-  def unionWith(other: A ==>> B)(f: (B, B) => B)(implicit o: Order[A]) =
+  def unionWith(other: A ==>> B)(f: (B, B) => B)(implicit o: Order[A]): A ==>> B =
     unionWithKey(other)((_: A, b: B, c: B) => f(b, c))
 
-  def unionWithKey(other: A ==>> B)(f: (A, B, B) => B)(implicit o: Order[A]) = {
+  def unionWithKey(other: A ==>> B)(f: (A, B, B) => B)(implicit o: Order[A]): A ==>> B = {
     def hedgeUnionWithKey(cmplo: A => Ordering, cmphi: A => Ordering, a: A ==>> B, b: A ==>> B): A ==>> B =
       (a, b) match {
         case (t1, Tip()) =>
@@ -1009,7 +1009,7 @@ sealed abstract class MapInstances extends MapInstances0 {
 
   implicit def mapCovariant[S]: Traverse[S ==>> ?] =
     new Traverse[S ==>> ?] {
-      override def findLeft[A](fa: S ==>> A)(f: A => Boolean) =
+      override def findLeft[A](fa: S ==>> A)(f: A => Boolean): Option[A] =
         fa match {
           case Bin(_, x, l, r) =>
             findLeft(l)(f) match {
@@ -1025,7 +1025,7 @@ sealed abstract class MapInstances extends MapInstances0 {
             None
         }
 
-      override def findRight[A](fa: S ==>> A)(f: A => Boolean) =
+      override def findRight[A](fa: S ==>> A)(f: A => Boolean): Option[A] =
         fa match {
           case Bin(_, x, l, r) =>
             findRight(r)(f) match {
@@ -1041,7 +1041,7 @@ sealed abstract class MapInstances extends MapInstances0 {
             None
         }
 
-      override def map[A, B](fa: S ==>> A)(f: A => B) =
+      override def map[A, B](fa: S ==>> A)(f: A => B): S ==>> B =
         fa map f
 
       override def foldMap[A, B](fa: S ==>> A)(f: A => B)(implicit F: Monoid[B]): B =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/Map.scala
```diff
@@ -813,16 +864,53 @@ sealed abstract class ==>>[A, B] {
...
     }
 
+  def splitRoot: List[A ==>> B] =
+    this match {
+      case Tip()           => List.empty[A ==>> B]
+      case Bin(k, x, l, r) => List(l, singleton(k, x), r)
+    }
+
   // Utility functions
+  // Because the following fucntions depend on some public interface, such as deleteFindMax,
```
xuwei-k: s/fucntions/functions

aloiscochard:
@monkey-mas that is an impressive piece of work, awesome!

Special thanks for taking care of writing benchmarks :)

tests/src/test/scala/scalaz/MapTest.scala
```diff
@@ -640,6 +794,120 @@ object MapTest extends SpecLite {
       val f = (k: Int, a: String, result: String) => result + "(" + k.toString + ":" + a + ")"
       fromList(List(5 -> "a", 3 -> "b")).foldrWithKey("Map: ")(f) must_== "Map: (5:a)(3:b)"
     }
+
+    "foldMapWithKey" ! forAll { a: Byte ==>> Byte =>
+      val intAddition = new Monoid[Int] {
```
xuwei-k: unused?

core/src/main/scala/scalaz/Map.scala
monkey-mas: Oops, thanks for pointing it out. Will fix it.

monkey-mas:
@aloiscochard 
Thanks a lot for your comment! No problem, I was just doing my job!
Let's keep doing our best!! :)

tests/src/test/scala/scalaz/MapTest.scala
monkey-mas: Oh, it actually is. I didn't explicitly provide `intAddition` to `foldMapWithKey` as an argument as the signature is like below.
`def foldMapWithKey[C](f: (A, B) => C)(implicit F: Monoid[C]): C =`

The following might be better for this case? WDYT?
```
a.foldMapWithKey(f)(intAddition) must_=== res
```

xuwei-k: We can use default `intInstance: Monoid[Int]` in `AnyVal.scala`.
https://github.com/scalaz/scalaz/blob/v7.3.0-M2/core/src/main/scala/scalaz/std/AnyVal.scala#L234-L237

I think we should just remove `intAddition` in this file.

monkey-mas: Fixed!
https://github.com/scalaz/scalaz/pull/1171/commits/1a175018a79be26f5fb5ef2725e1b2d4d448bb27#diff-1907a0925b3477216b774a9c03108d76R785

aloiscochard:
Had time to lead the diff and it does looks great! :+1: 

I would love to forward port it to scalaz8 at some point, where there will be a dedicated jmh enabled `benchmarks` module.

Are the jmh bench available somewhere?

monkey-mas:
Hi @aloiscochard,

> Are the jmh bench available somewhere?

As I've mentioned, I used [sbt-jmh](https://github.com/ktoso/sbt-jmh) plugin. The benchmark code I've written is [this one](https://github.com/monkey-mas/scalaz/blob/compare-map-benchmark/src/main/scala/jmh/MapBenchmark.scala) and the corresponding results are stored [here](https://github.com/monkey-mas/scalaz/tree/6a1db2c7fd462a66461eeff44a438cc45bee490d/result).

AFAIU, to take benchmark with sbt-jmh, what we have to do are:
- [Enable the plugin](https://github.com/monkey-mas/scalaz/commit/cba38e7deb798a34fad18f025f787b52d5ff5f3f)
- Create benchmark code under `src` directory
  - Benchmark methods should be annotated with `@Benchmark`
  - If there're some data or variables you'd like to use for your benchmark, you should define a class annotated with `@State(Scope.Benchmark)` that I renamed to `@JmhState(Scope.Benchmark)` as there might be a name collision from scalaz module. The time elapsed within the `@State` classes is *NOT* considered for your benchmark execution time such as `avgt`. ref. [Benchmark State](http://tutorials.jenkov.com/java-performance/jmh.html#benchmark-state).
  -  You can inject your `@State` as an argument for example [like this](https://github.com/monkey-mas/scalaz/blob/6a1db2c7fd462a66461eeff44a438cc45bee490d/src/main/scala/jmh/MapBenchmark.scala#L61). 
- Execute your benchmark
  - Like you can find from [sbt-jmh README](https://github.com/ktoso/sbt-jmh#adding-to-your-project), a 10-time warm-up or even more is recommended because of JVM JIT compilation mechanism.
    You can set this up with `wi` option.
  - You can also set your benchmark mode such as `AverageTime/avgt` or `All/all` with `bm` option.
  - For example, I executed the following command for my benchmark:
   ```
   sbt -mem 8192 "jmh:run -i 50 -wi 30 -f 1 -t 1 -bm avgt"
   ```

Hope this helps! :)
If there's a question or something ambiguous, please let me know!

monkey-mas:
@xuwei-k 
To complete this PR, I've done three additional changes:

1. [Changed the interface of `mergeWithKey`](https://github.com/scalaz/scalaz/pull/1171/commits/448646eb4ec80ecdc14c8b859f18324eebbe67bf) 
  - With this change, the implementation of `{union,difference,intersection}WithKey` doesn't need explicit type annotations [like this](https://github.com/scalaz/scalaz/pull/1171/commits/8d0ad8bb40214d7d9b7d5efddfbee720f69be441#diff-a69589bbcd232b8a39f491a89e249f70R537).
2. [Modified the interface of `differenceWith`](https://github.com/scalaz/scalaz/pull/1171/commits/099194d542171ab31cc9a52df1547ad20e82f6e3).
  - We have `XxxWith(m: A ==>> B)(f: ...)` method format in common, such as `unionWith` or `intersectionWith`. This helps type inference so it's worth modifying the interface.
3. [Remove unnecessary type annotations](https://github.com/scalaz/scalaz/pull/1171/commits/f9c36d127e49b25fb83419103e1b3fba0c7920c4) due to modification of  `mergeWithKey`.

xuwei-k:
LGTM. Thanks awesome work! 👍 
I'll release 7.3.0-M3 soon.

monkey-mas:
Thanks a lot for your help and review!!! :D

PR #1175, opened 2016-05-31
More laziness.

aloiscochard:
As pointed out by @S11001001, `Apply` wasn't lazy enough.

What about `Bind`, it looks similar to want we have in 7 atm, should we make it more lazy as well?


base/src/main/scala/data/MaybeInstances.scala
```diff
@@ -5,10 +5,10 @@ import typeclass._
 import Maybe.{Just, Empty}
 
 trait MaybeInstances extends MonadClass[Maybe] {
-  override def ap[A, B](ma: Maybe[A])(mf: Maybe[A => B]): Maybe[B] =
+  override def ap[A, B](ma: => Maybe[A])(mf: Maybe[A => B]): Maybe[B] =
     ma.fold(a => map[A => B, B](mf)(f => f(a)), Empty())
```
S11001001: This ap is symmetric, but `mf` should be checked first.

base/src/main/scala/typeclass/MonadInstancesStdLib.scala
```diff
@@ -3,15 +3,15 @@ package typeclass
 
 trait MonadInstancesStdLib {
   implicit val option: Monad[Option] = new MonadClass[Option] {
-    override def ap[A, B](oa: Option[A])(f: Option[A => B]): Option[B] = oa.flatMap(a => f.map(_(a)))
-    override def flatMap[A, B](oa: Option[A])(f: A => Option[B]): Option[B] = oa.flatMap(f)
+    override def ap[A, B](oa: => Option[A])(f: Option[A => B]): Option[B] = oa.flatMap(a => f.map(_(a)))
```
S11001001: Likewise, `f` should be checked first.

base/src/main/scala/typeclass/MonadInstancesStdLib.scala
```diff
@@ -3,15 +3,15 @@ package typeclass
...
+    override def flatMap[A, B](oa: => Option[A])(f: A => Option[B]): Option[B] = oa.flatMap(f)
     override def map[A, B](oa: Option[A])(f: A => B): Option[B] = oa.map(f)
     override def pure[A](a: A): Option[A] = Option(a)
   }
 
   implicit val list: Monad[List] = new MonadClass[List] {
-    override def ap[A, B](xs: List[A])(f: List[A => B]): List[B] = xs.flatMap(a => f.map(_(a)))
-    override def flatMap[A, B](xs: List[A])(f: A => List[B]): List[B] = xs.flatMap(f)
+    override def ap[A, B](xs: => List[A])(f: List[A => B]): List[B] = xs.flatMap(a => f.map(_(a)))
```
S11001001: This is backwards wrt the default definition in `BindClass`; the `f` effect should be "first".

base/src/main/scala/typeclass/Bind.scala
```diff
@@ -3,7 +3,7 @@ package typeclass
 
 trait Bind[M[_]] {
   def apply: Apply[M]
-  def flatMap[A, B](ma: M[A])(f: A => M[B]): M[B]
+  def flatMap[A, B](ma: => M[A])(f: A => M[B]): M[B]
```
S11001001: This doesn't seem to be hugely beneficial; except for representable/distributive functors, we must always inspect `ma`.

It might be worth introducing something like ermine's [Suspendable](https://bitbucket.org/ermine-language/ermine-scala/src/40647cf0c50219e2a411a7e242fe16dc724d1e3c/core/src/main/scala/com/clarifi/reporting/Run.scala?at=default&fileviewer=file-view-default#Run.scala-8:11) that users could wrap around `ma` for non-strict `f`, e.g. [usage](https://bitbucket.org/ermine-language/ermine-writers/src/355667c10039839d71b076b420412106e0cd6e3d/writers/javafx/src/main/scala/com/clarifi/reporting/writers/JavaFXWriter.scala?at=default&fileviewer=file-view-default#JavaFXWriter.scala-143).

S11001001: ...this is because the usefully lazy signature seems to be

```scala
[A, B](ma: M[A])(f: (=> A) => M[B]): M[B]
```

but that seems annoying for other reasons.

PR #1179, opened 2016-06-02
Fix #1178: rename "clazz" to "typeclass"

### LINKED ISSUE #1178: rename "clazz" to "typeclass" for new users
This is for the Scalaz 8 branch.  I don't think I have the ability to add a label.

I hate to debate names to death, but in some cases, I think it's worth some rework.  Here's my thoughts:

- "Clazz" a name often used by Java programmers for all their reflection insanity. 
- Even "class" is an unfortunate name because of what it means to Java programmers.
- "Typeclass" is at least a Google-able term that gets people closer to the truth.
- "Typeclass" is only four characters more -- not impossibly long, right?

The PR is easy enough to make, which I'll do shortly.  I don't know if writing up this issue was a needless step process-wise.

shajra:
Here's all I did:

```shell
$ find . -name '*.scala' -exec perl -i -wlp -e 's/clazz/typeclass/g' {} +
$ mv ./base/src/main/scala/clazz ./base/src/main/scala/typeclass
```


puffnfresh:
:+1:

vmarquez:
👍

shajra-cogscale:
Cool, I think we have enough consensus to merge.  Maybe a quick chime from @aloiscochard might complete a good set?  Otherwise, is it primarily @xuwei-k who does the merging?

shajra:
Cool, I think we have enough consensus to merge the PR.  Maybe it might be nice to hear from @aloiscochard too?  I'm curious, who's taking responsibility for merging PRs for the Scalaz 8 branch?

vmarquez:
@shajra the 'rule of thumb' has been if there are 2 thumbs up, we can merge.  For this one, I'd perosnlly prefer to hear @aloiscochard's thoughts just as a courtesy, :-),  he might have an argument for keeping it the way it is.

aloiscochard:
:+1: 

Totally agree with process, and I like the rename, let's go for it :)

PR #1180, opened 2016-06-02
Alternative StateT representation that admits stack-safe operations for tail-recursive monads

TomasMikula:
I propose this alternative representation of `StateT`:

```scala
case class Wrap[F[_], S1, S2, A](run: S1 => F[(S2, A)]) extends IndexedStateT[F, S1, S2, A]
case class FlatMap[F[_], S1, S2, S3, A, B](a: IndexedStateT[F, S1, S2, A], f: (S2, A) => IndexedStateT[F, S2, S3, B]) extends IndexedStateT[F, S1, S3, B]
```

The benefit is that one does not need to trampoline `StateT` anymore to achieve stack-safety (the trampoline is built-in).

The trade-off is that iterated `mapK` will not be stack-safe even for a lazy inner monad, because the `mapK` operation cannot be represented lazily in this new representation, and thus iterated `mapK`s will build up a stack-unsafe function. This can be justified by the fact that `mapK` changes the inner monad, and there will only be a constant number of different monads in any program, therefore at most a constant number of `mapK`s should be necessary in any program. This is analogous to iterated `Free.mapSuspension` not being stack-safe.


aloiscochard:
@TomasMikula  The main problem with this, is that you loose some performance when using it in a strict way and a small number of elements.

I did present an approach at LambdaConf which at least in spirit seems accepted by contributors for scalaz8 which would make us have specific strict/non-strict implementation.

You can find an experimental implementation here:
https://github.com/aloiscochard/scato/blob/master/transformers/src/main/scala/StateT.scala

I feel like we should really focus on this for scalaz8, the representation used in scato is not the optimal one which is more like cats.Eval and seems similar to what you did here, isn't it?

TomasMikula:
@aloiscochard Having distinct strict and non-strict implementations probably makes the most sense. Not sure though where the scato implementation fits into this picture, because it is lazy, yet it is not stack-safe for strict monads. It actually looks a lot like my [other experimental representation](https://github.com/TomasMikula/cats/blob/5f125e69c5105be63d38e5301f89b8c00f65ed1e/core/src/main/scala/cats/data/StateT.scala) (also not stack-safe for strict monads), in which my goal was to make the representation look like

```scala
S => F[(S, A)]
```

I achieved

```scala
S -> F[(S, A)]
```

where `->` is a lazy representation of function composition.

The representation in this PR has the added benefit that it allows for stack-safe execution for `BindRec` monads.

I don't have experience in rigorous benchmarking, but I might put together some poor man's benchmark to see how the performance of the strict case compares to the current scalaz implementation.

aloiscochard:
@TomasMikula not sure what you mean by not stack-safe for strict monad?

For example if you run it with `Identity` it does not stack overflow, in which case would that happen?

TomasMikula:
@aloiscochard Create `StateT` via 10000x `flatMap` and then run, i.e. [this test case.](https://github.com/TomasMikula/scalaz/blob/edaab495ed6aea9bae24a9783a0db14da8f5da7d/tests/src/test/scala/scalaz/StateTTest.scala#L78-L81)

aloiscochard:
@TomasMikula okay, seems it's exactly the same as what I'm doing in the `scato` bencharks there: https://github.com/aloiscochard/scato/blob/master/benchmarks/src/main/scala/Interpreters.scala or am I missing something?

TomasMikula:
@aloiscochard Hmm, right... Let me think of another test case.

TomasMikula:
@aloiscochard Can you try this with `scato`?

```scala
def nasty: State[Int, Int] =
  State[Int, Int](n => (n-1, n-1)).flatMap(i =>
    if(i > 0) nasty
    else State[Int, Int](n => (n, n))
  )

nasty.eval(10000) // must_=== (0)
```

I added this as a test case to this PR.

TomasMikula:
So I did some benchmarking and the proposed new representation performs significantly better in all tested scenarios.

I tested combinations of the following use-cases:

1) The embedded monad is strict and stack-safety is not needed.
2) The embedded monad is strict and stack-safety is required.
3) The embedded monad is already trampolined anyway.

a) `StateT` instance is run only once.
b) `StateT` instance is run many times.

L) a chain of left-associated `flatMap`s.
R) a chain of right-associated `flatMap`s.

This gives a total of 12 use cases.

Intuition could tell us that the representation proposed in this PR would perform worse at
 - 1), because there is a (built-in) trampoline even when trampoline is not needed;
 - 3), because there is a double trampoline;
 - b), because the new representation is cheap to construct, but expensive to run, because it is "interpreted";
 - L), because on each run, `flatMap` need to be re-associated to the right;

but none of this proved to be true.

I hijacked `scato` for the purpose, just because it has setup for benchmarks. [Here](https://github.com/TomasMikula/scato/blob/31506ebfec8b15d7d176abe399e94569ac652abd/benchmarks/src/main/scala/Interpreters.scala) are the benchmarks run with `scalaz` 7.3.0-M3, and [here](https://github.com/TomasMikula/scato/commit/3873b8d1fc8b0d88ef8882b6a590f611c60c9aa2) are the changes done for benchmarking this PR.

These are the results:

|     | scalaz 7.3.0-M3 (ops/s) | this PR (ops/s) | change |
| --- | ----------------------- | --------------- | ------ |
| 1.a.L. `strict_unsafe_left`         | 119349.345 | 266637.472 | 2.23x |
| 1.a.R. `strict_unsafe_right`        | 121467.521 | 388086.403 | 3.19x |
| 1.b.L. `strict_unsafe_reused_left`  |   1461.120 |   3519.458 | 2.40x |
| 1.b.R. `strict_unsafe_reused_right` |   1309.615 |   3914.850 | 2.98x |
| 2.a.L. `strict_safe_left`           |     46.260 |    121.726 | 2.63x |
| 2.a.R. `strict_safe_right`          |     69.192 |    218.881 | 3.16x |
| 2.b.L. `strict_safe_reused_left`    |      0.505 |      1.568 | 3.10x |
| 2.b.R. `strict_safe_reused_right`   |      0.682 |      2.161 | 3.16x |
| 3.a.L. `trampolined_left`           |     44.399 |    165.146 | 3.71x |
| 3.a.R. `trampolined_right`          |     70.482 |    241.748 | 3.42x |
| 3.b.L. `trampolined_reused_left`    |      0.517 |      2.114 | 4.08x |
| 3.b.r. `trampolined_reused_right`   |      0.699 |      2.562 | 3.66x |

vmarquez:
Wow looks great @TomasMikula I will look at this a bit more tomorrow but I'm thinking we'll definitely go with this!

aloiscochard:
@TomasMikula awesome work! thanks so much for the effort, I'll definitely reuse your benchmark to test the scato encoding :-)

:+1:

edmundnoble:
@TomasMikula JSYK `mapK` calls are absolutely not limited by the number of distinct monadic types in a program, any more than `map` calls are limited by the number of distinct ordinary types in a program. Consider: `def repeatedly[F[_]: Functor, A](fa: F[A])(trans: F ~> F): F[A] = trans(repeatedly(fa)(trans))`

TomasMikula:
@edmundnoble I didn't mean that you can't write a program with arbitrary number of `mapK` calls, just that you shouldn't.

Do you have a useful example where the number of `mapK` calls is not bounded by a constant?

I would say that if you have an unbounded number of `mapK` calls, you have a performance leak, and keeping it stack-safe will only delay its detection.

edmundnoble:
@TomasMikula Your argument is exactly the same for the HK-type level as otherwise; people only have a constant number of types in their program, so your `map` calls don't need to be stack-safe! People don't have a "lot" of types in their program, until they start to loop on an `A => A Either B` or a `F ~> Coproduct[F, G, ?]` and suddenly stack-safety is a concern. I'd need a convincing reason that `mapK` differs from `map` in this regard.

TomasMikula:
@edmundnoble 

> Your argument is exactly the same for the HK-type level as otherwise;

Nothing to do with higher-kinded types; that's not what I base my argument on.

Rather, it is based on what role `F[_]` plays in `StateT[F[_], S, A]`&mdash;it represents an effect. Functions passed to `mapK` are effect homomorphisms (monad homomorphisms), for example an injection into a larger monad. My argument then is that these transitions between effects don't advance the computation. Therefore, if you have a loop in effect transitions, you loop without making progress.

edmundnoble:
Monadic endomorphisms are as useful to construct recursive fixpoints as endomorphic functions. Can you provide some evidence that this isn't the case? It seems fairly obvious to me... for example, tweaking my previous example only a little:
```scala
def repeatedlyK[F[_]: Functor, A](fa: F[A])(trans: F ~> F, stop: F ~> Const[Boolean, ?]): F[A] =
  if (stop(fa).getConst) fa
  else trans(repeatedly(fa)(trans, stop))
``` 

I also don't understand why you seem to think type changing operations are the only ones that can make computational progress.

TomasMikula:
> Monadic endomorphisms are as useful to construct recursive fixpoints as endomorphic functions. 

Absolutely. Let me try to clarify what I mean by not making progress. From the point of view of `StateT[F, S, A]`, `F[_]` is the type of effects. When I do `hoist` (I realized `mapK` has slightly different signature that also transforms the state and value, not just the effect type), I'm not changing anything about the core computation (which has the form of `S => (S, A)`). I'm just manipulating the effects, not advancing the core computation.

Sure, it's all relative and at some point effects are just values. And you might want to run some recursive computation on these values. But that computation will not involve values of `StateT[F, S, A]` (simply because the function passed to `hoist` (or `mapK`) doesn't have access to any value of `StateT[F, S, A]`). Here's how you would use your recursive definition of `repeatedlyK`:

```scala
def untilStop[F[_]: Functor](trans: F ~> F, stop: F ~> Const[Boolean, ?]): F ~> F =
  λ[F ~> F](fa => repeatedlyK(fa)(trans, stop))

val trans: F ~> F = ???
val stop: F ~> Const[Boolean, ?] = ???

val st: StateT[F, S, A] = ???

val st1 = st.hoist(untilStop(trans, stop))
```

From the point of view of `StateT`, it is a single transformation on effects.

Everything so far applies more generally to any monad transformer, not just `StateT`.

In addition, `StateT[F, S, A]` specifically is isomorphic to `S => F[(S, A)]`. Since you can't introspect a function, any (potentially recursive) computation taking a value of type `StateT[F, S, A]`, absent of a value of type `S`, will not depend on the particular value of `StateT[F, S, A]` passed in. Specifically, how many times `hoist` is invoked and with what arguments does not depend on the value of `StateT[F, S, A]` passed in. So all the transformations to be passed to `hoist` can be computed independently of any value of `StateT`. Since it doesn't matter in what order `hoist` is invoked relative to other methods (like `flatMap`), the invocations can be reordered so that all `hoist`s appear in sequence, but at that point they can be replaced with a single `hoist` taking a composition of all the transformations. As for `mapK`, I would remove it from the public interface as too powerful and potentially dangerous (result can depend on the value of effect).

edmundnoble:
@TomasMikula Sorry, I was under the impression that `mapK` was another name for `hoist` 😝 you are right that you *can* re-order the hoist calls, but that may require tighter coupling between modules; the same argument applies IMO to having/not having a stack-safe `map` call, just `hoist` is called less often. Thus AFAIK it's nice to be able to re-order them arbitrarily without breaking the stack, but TBH you are certainly not alone when it comes to transformers having a non-stack-safe `mapK`; not a huge sticking point for me, just making sure we know it's a potential place for stack-unsafety. Perhaps documenting it would be worth it?

TomasMikula:
Yeah, might be worth documenting. Will include a line when I get to rebasing this.

It should be possible to support stack-safe `hoist`/`mapK` with a third case

```scala
case class MapK[F[_], G[_], S1, S2, S3, A, B](a: IndexedStateT[F, S1, S2, A], f: F[(S2, A)] => G[(S3, B)], F: BindRec[F]) extends IndexedStateT[G, S1, S3, B]
```

but I don't think it's worth the complication.

edmundnoble:
Me neither, certainly not without benchmarks. Putting TC instances inside data may be a bit suspect IMO, specially until we know they're serializable. Good to know, though.

TomasMikula:
> Putting TC instances inside data may be a bit suspect

I know there is some reluctance towards that, but I don't see it being different from storing anonymous functions that are themselves closures over a typeclass instance. See the original `mapK` implementation in this diff: it creates an anonymous function that is a closure over `Monad[F]`.

edmundnoble:
@TomasMikula Agreed, IMO.

fosskers:
I just hit this issue, where `StateT` blows the stack after an embarrassingly low number of recursive calls. Is there something preventing this from being merged, or is there an equivalent fix in snapshot versions?

vmarquez:
@fosskers I'll have to pull in changes from another branch and fix some conflicts before I merge it in.  Also, with the current version of StateT, you can get stack safety by ensuring your `T` is stack safe.  If you don't have a `T` you can make it a `Trampoline`.  Does that help?  If you need more assistance on how to do this, let me know.

fosskers:
I'm just using the basic `State` (I'm writing some simple benchmarks), so I believe that's just `Identity` underneath?

fosskers:
...which doesn't seem to be enough. Here's the code in question:

```scala
     run("scalaz/countdown") { new Benchmark {
       def go: scalaz.State[Int, Int] = scalaz.State.get.flatMap({ n =>
         if (n <= 0) scalaz.State.state(n) else scalaz.State.put(n - 1) >> go
       })                                                                                                                                                                                                           

       def run() = go.run(1000)                                                                                                                                                                                     
     }}
```

vmarquez:
@fosskers if you check out this here:
https://github.com/scalaz/scalaz/blob/series/7.3.x/tests/src/test/scala/scalaz/StateTTest.scala#L80

You'll see that if you return a value inside  your state of `a => Trampoline.done(ret)` you'll get stack safety (since the type will be of `StateT[Trampoline, A, B]`).  Hopefully that is enough to get you started.

fosskers:
That did it, thanks @vmarquez !

```scala
       def go: scalaz.StateT[Trampoline, Int, Int] = scalaz.State.get.lift[Trampoline].flatMap({ n =>
         if (n <= 0)
           scalaz.StateT(_ => scalaz.Trampoline.done((n,n)))                                              
         else scalaz.State.put(n - 1).lift[Trampoline] >> go
       })
```

jdegoes:
@TomasMikula What's the status of this PR?

fosskers:
Merging this into `7.2` would also allow me to correct [this negative PR](https://github.com/fosskers/scalaz-and-cats#monads-and-monadic-recursion).

TomasMikula:
@jdegoes The current status is a merge conflict. 😆

If there's will to merge this, I will resolve the conflicts.

jdegoes:
@TomasMikula Well, @fosskers sure seems to want it! 

I did not follow all the conversation but the tradeoff described in the initial PR seems reasonable.

TomasMikula:
Rebased.

However, this will not be portable to 7.2.x, since it changes typeclass constraints on some methods.

Issue #1176, opened 2016-06-01
Reduce unneeded calculation of Traverse[List].traverse


gakuzzzz:
Hi all,

```scala
val f: Int => Option[String] = ??? // too heavy calculation 
(1 to 10).toList.traverse(f)
```

Currently, Traverse[List].traverse always apply `f` to all elements.
While on the other hand, cats.Traverse does not apply `f` to elements following an element what returns None.

It should reduce unneeded calculation.

Kenji(@xuwei-k) had created a [PoC](https://github.com/xuwei-k/scalaz/commit/ab7c4a34ff90c9).
However it does not stack safe.

Is it necessary to introduce something like the cats.Eval ?
Are there another good way?

thanks,

xuwei-k:
https://github.com/xuwei-k/scalaz/commit/103ee746786dbffa7baff9951a39aefab39239a2#commitcomment-17708174

But #1022 does not solve this :( 

https://github.com/scalaz/scalaz/compare/scalaz:288300a...xuwei-k:8b8be0d

```
[info] ! scalaz.TraverseTest.:issue 1176: Exception raised on property evaluation.
[info] > Exception: java.lang.AssertionError: [Some(4),Some(3),None,Some(2),Some(1)] !== [Some(1),Some(2)]
```

https://travis-ci.org/xuwei-k/scalaz/jobs/134638544#L5989

xuwei-k:
in cats 0.6.0

https://github.com/typelevel/cats/blob/v0.6.0/core/src/main/scala/cats/std/list.scala#L41-L53

```
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_77).
Type in expressions for evaluation. Or try :help.

scala> import cats._, syntax.all._, std.all._
import cats._
import syntax.all._
import std.all._

scala> val xs = List(Option(1), Option(2), None, Option(3), Option(4))
xs: List[Option[Int]] = List(Some(1), Some(2), None, Some(3), Some(4))

scala> var ys = List.empty[Option[Int]]
ys: List[Option[Int]] = List()

scala> val z = Traverse[List].traverse(xs){ x =>
     |   ys ::= x
     |   x
     | }
z: Option[List[Int]] = None

scala> ys
res0: List[Option[Int]] = List(None, Some(2), Some(1))
```

S11001001:
[...]

PR #1181, opened 2016-06-02
Add Apply.apM, which allows for short-circuiting implementations.

TomasMikula:
The new `Apply#apM` method is similar in spirit to `cats`'s [`Apply#map2Eval`](https://github.com/typelevel/cats/blob/1236cf8782757a154caf78ce80563146e5362adc/core/src/main/scala/cats/Apply.scala#L36-L59), but doesn't fix `M` to any particular monad (like `cats.Eval`).

There exists a default implementation that is not short-circuiting. Applicative effects that support short-circuiting should override this method. Implementation for `Option` and fix for #1176 are included in this PR.

### DIFF
--- a/core/src/main/scala/scalaz/Apply.scala
+++ b/core/src/main/scala/scalaz/Apply.scala
@@ -20,6 +20,18 @@ trait Apply[F[_]] extends Functor[F] { self =>
     */
   def ap[A,B](fa: => F[A])(f: => F[A => B]): F[B]
 
+  /** Version of `ap` that takes one parameter, the one whose `F`-action
+    * should be performed later, in an additional context `M`.
+    *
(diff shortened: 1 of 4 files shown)


TomasMikula:
So I realized that it would be unlawful to use `apM` with anything but `Id` or lazy value (`Trampoline`, `Need`). An alternative signature could be

```scala
def apM[M[_], A, B](mfa: => M[F[A]])(f: => F[A => B])(implicit M: M <~> Id): M[F[B]]
```

but would it be OK to provide implicit instances of `Trampoline <~> Id` and/or `Need <~> Id`?

gakuzzzz:
It's cool

TomasMikula:
Closing as this cannot be merged as is. #1189 can be used for further discussion.

PR #1182, opened 2016-06-03
Correct function names in Task's ScalaDoc

peterneyens:
Changed the function name references from `run` to their `unsafePerformSync` equivalent.


xuwei-k:
👍

Issue #1177, opened 2016-06-01
Where should laws live?


vmarquez:
Should we have another file per TC, or put them in the Instance file?

@S11001001 @aloiscochard @puffnfresh @tonymorris

shajra:
At work we made a small learning project, and we had some room to lay things out how we thought we'd expect things to be.  We ended up with everything related to a type class in the same file.  This included both laws and types classes.  I made [an example gist](https://gist.github.com/shajra/74f28c65a32c2a0b88d05dcf34a51ebb).

I'm not sure exactly what all the arguments are for different ways of organizing things, but agreeing to a pattern will help make things go faster.

shajra:
Actually, I'm just now sitting with the branch, and realizing it's already factored out into separate files.  That's fine by me.  For example, I see a file called "MonadLaws.scala" with a trait called "MonadLaws" that mixed into a "laws" object in the companion object of the Monad.  I'm not sure if it's too many files for one abstraction (Monad, MonadClass, MonadInstances, MonadInstancesStdLib, MonadLaws), but it makes sense.  And maybe it's nice to have not too much in a single file.

jbgi:
```MonadInstances``` is actually not needed: the instance definition(s) it contains should probably be defined in companion objects, ie. the identity monad would go into the companion object of ```Identity```. I think @aloiscochard already agreed on that in scato but we forgot to follow-up on his point.

vmarquez:
@jbgi that sounds fine to me, maybe do a PR and have @aloiscochard review?  I'm not too picky where things are, just want to make sure it's logical and things are well demarcated.

aloiscochard:
I don't feel like we should have laws available at runtime for users of the library, does anyone see a strong reason to that?

I think we should definitely provide `Aribtrary` (or actually what it's called in @xuwei-k scalapros lib), but I'm unsure about laws.

@jbgi I think we still need `MonadInstances` to provide instances for the standard library, isn't it?

shajra:
If I have a custom monad or whatever in my code, I might want to test the laws.  Maybe we could have a "scalaz-test" jar that included laws and useful Arbitrary (or whatever they are in scalaprops) instances?

aloiscochard:
seems sensible to have it as an other module (jar), let's see if we can do it without making the usage of them too painful in the base module.

aloiscochard:
I close this one, and I'll create relevant tasks.

PR #1183, opened 2016-06-03
Scalaz8: uniformize location of monad instances

jbgi:
as per https://github.com/scalaz/scalaz/issues/1177#issuecomment-223207891
cc @aloiscochard


jbgi:
> @jbgi I think we still need MonadInstances to provide instances for the standard library, isn't it?

@aloiscochard yes. I moved instances for the standard library into MonadInstances (2nd commit).

aloiscochard:
Superseeded.

PR #1184, opened 2016-06-03
Scalaz8/Base hierarchy: add missing implicits for Bind. Ensure no ambiguities.

jbgi:
For each trait of the hierarchy, a particular type class should appear only once as return type.
cc @aloiscochard


aloiscochard:
Nice catch, and actually the title of that PR will be the first line I'll add in `CONTRIBUTING.md`!

aloiscochard:
Merged as b43386baf5ad5c316aba4c688512565f29845766

PR #1185, opened 2016-06-03
Scalaz8: Remove redundant implicits to avoid ambiguities.

jbgi:
Companion objects should only declare implicits for the type classes
at the top of a hierarchy. The resolution of subclasses should be
left to BaseHierarchy, otherwise ambiguities arise.
cc @aloiscochard 

Without this patch the following fails to compile:
```scala
import scalaz._
import data.Disjunction._
import typeclass._

class Test {

  def test[A, M[_] : Applicative : Bind](ma : M[A]) {}

  def run(): Unit = {
    val  v = right(1) : String\/Int
    test(v)
  }

}
```


aloiscochard:
Thanks again @jbgi for removing the artifacts of our experiment!

aloiscochard:
Merged as 743b75af61bea556d0d398a0bded7a084a248196

PR #1190, opened 2016-06-07
[Scalaz 8] add Const, Semigroup and Monoid

julien-truffaut:
I wanted to see how the new typeclass encoding will deal with `Const` which previously required a trait hierarchy.

PR #1191, opened 2016-06-07
Add `Const` data type and the `Semigroup` and `Monoid` typeclasses.

aloiscochard:
This superseed https://github.com/scalaz/scalaz/pull/1190.

Just fixing a few stylish things but have to create this one to make jenkins and branch protection happy.

### DIFF
--- a/base/src/main/scala/data/Const.scala
+++ b/base/src/main/scala/data/Const.scala
@@ -1,0 +1,8 @@
+package scalaz
+package data
+
+final case class  Const[A, B](getConst: A){
+  def retag[C]: Const[A, C] = this.asInstanceOf[Const[A, C]]
+}
+
+object Const extends ConstInstances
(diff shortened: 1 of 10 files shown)


base/src/main/scala/data/Const.scala
```diff
@@ -0,0 +1,8 @@
+package scalaz
+package data
+
+final case class  Const[A, B](getConst: A){
+  def retag[C]: Const[A, C] = this.asInstanceOf[Const[A, C]]
```
puffnfresh: We should use `liftCvf` as described in "[When can Liskov be lifted?](http://typelevel.org/blog/2014/03/09/liskov_lifting.html)"

aloiscochard: Oh nice! Thanks for the reference, I didn't know about this one but I suppose that's the one @S11001001 mentioned to me during lambda conf.

PR #1192, opened 2016-06-08
Forget structure

vmarquez:
Just thought I'd get my feet wet with trying to understand where everything goes.  We also may not want/need this depending...

### DIFF
--- a/base/src/main/scala/data/Forget.scala
+++ b/base/src/main/scala/data/Forget.scala
@@ -1,0 +1,9 @@
+package scalaz
+package data
+import typeclass._
+
+final case class Forget[A, B, C](forget: B => A){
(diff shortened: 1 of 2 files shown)


aloiscochard:
@vmarquez interesting, where does that come from?

Was looking into https://hackage.haskell.org/package/lens but did not saw it, I'll migrate hopefully soon my optics from the experimental stuff but I don't think I was using a typeclass like this one.

kcsongor:
@aloiscochard `Forget r` is the kleisli category of the `Const r` functor, and you get a getter when used as the profunctor in optics. You could write it as `UpStar (Const r)`

aloiscochard:
@kcsongor that surely helps, thanks a lot :-)

:+1: we should probably squash before merging though.

PR #1193, opened 2016-06-09
Corecursive list

S11001001:
A different take on ephemeral streams.

/cc @aloiscochard @vmarquez


jedws:
👍

jbgi:
I found this blog post interesting on the subject: http://jyp.github.io/posts/controlled-fusion.html
In the blog post, CorecursiveList is named "Nu list". What is interesting is how to combine "Nu List" (unfold) with "Mu List" (fold) while preserving a uniform interface.

jdegoes:
See also [Matryoshka](https://github.com/slamdata/matryoshka/tree/master/core/shared/src/main/scala/matryoshka).

PR #1194, opened 2016-06-09
Change default Scala version to 2.12.1.

TomasMikula:
Is it perhaps time to change the default Scala version to 2.11.8? This will not have much impact, since `scalaz` is cross-built anyway, but `scalaVersion` happens to be the version that is used by `publishLocal`. So I suppose if most devs are on 2.11.8, this makes sense.


TomasMikula:
Yup, the time has moved on since then. Rebased and changed to 2.12.1.

PR #1195, opened 2016-06-09
Add method Foldable#fold1Opt.

TomasMikula:
`fold1Opt` is to `fold` what `foldMap1Opt` is to `foldMap`.


TomasMikula:
The failure is unrelated. Can someone please restart the build?

puffnfresh:
So the error doesn't get lost:

```
[info] ! scalaz.concurrent.FutureTest.Future:not deadlock when using Nondeterminism#chooseAny: Exception raised on property evaluation.
[info] > Exception: java.lang.RuntimeException: Timeout occured, possible deadlock.
```

TomasMikula:
It is green now.

vmarquez:
👍

PR #1196, opened 2016-06-10
Reduce running time of tests.

TomasMikula:
This reduces the running time of `testsJVM/test` on my laptop from 622s to 87s and avoids 7 `OutOfMemoryError`s during tests.


aloiscochard:
:+1: 

We preserve stack safety test and massive gain on testing time, nice!

PR #1200, opened 2016-06-13
use Liskov instead of Leibniz in AssociativeOps

xuwei-k:
before

```
scala> import scalaz._, std.tuple._, syntax.associative._
import scalaz._
import std.tuple._
import syntax.associative._
scala>  val a1 = (1, (2, 3))
a1: (Int, (Int, Int)) = (1,(2,3))

scala> val b1 = a1.reassociateLeft
<console>:26: error: could not find implicit value for parameter ev: scalaz.Leibniz.===[(Int, Int),(TT, C)]
       val b1 = a1.reassociateLeft
                   ^

scala> val b1 = a1.reassociateLeft[Int, Int]
b1: ((Int, Int), Int) = ((1,2),3)
```

after

````
scala> import scalaz._, std.tuple._, syntax.associative._
import scalaz._
import std.tuple._
import syntax.associative._
scala>  val a1 = (1, (2, 3))
a1: (Int, (Int, Int)) = (1,(2,3))

scala> val b1 = a1.reassociateLeft
b1: ((Int, Int), Int) = ((1,2),3)
```

### DIFF
--- a/core/src/main/scala/scalaz/syntax/AssociativeSyntax.scala
+++ b/core/src/main/scala/scalaz/syntax/AssociativeSyntax.scala
@@ -4,13 +4,13 @@ package syntax
 /** Wraps a value `self` and provides methods related to `Associative` */
 final class AssociativeOps[F[_, _],A, B] private[syntax](val self: F[A, B])(implicit val F: Associative[F]) extends Ops[F[A, B]] {
   ////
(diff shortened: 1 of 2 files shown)


puffnfresh:
:+1:

jedws:
👍

PR #1201, opened 2016-06-17
Traversable consistency

vmarquez:
I wasn't sure if we should do it this way, but I was trying to make Traversable instances consistent with how the monad instances are.


vmarquez:
I suppose this is really a question/proposal for @aloiscochard, there may be totally valid reasons why this is a bad idea.

aloiscochard:
@vmarquez this is perfect! :+1: 

Please squash the commits before merging, ty! :)

aloiscochard:
Superseeded by #1205

PR #1202, opened 2016-06-19
Disjunction/Maybe

aloiscochard:
A few things to polish, found while impl. some optics.

### DIFF
--- a/base/src/main/scala/data/DisjunctionFunctions.scala
+++ b/base/src/main/scala/data/DisjunctionFunctions.scala
@@ -1,10 +1,17 @@
 package scalaz
 package data
 
-trait DisjunctionFunctions {
-  @inline def -\/[L](value: L): Disjunction[L, Nothing] = Disjunction.-\/(value)
-  @inline def \/-[R](value: R): Disjunction[Nothing, R] = Disjunction.\/-(value)
+import Disjunction.{\/, \/-, -\/}
 
-  @inline def left[L](value: L): Disjunction[L, Nothing] = Disjunction.-\/(value)
-  @inline def right[R](value: R): Disjunction[Nothing, R] = Disjunction.\/-(value)
+trait DisjunctionFunctions {
+  @inline def -\/[L](value: L): Disjunction[L, Nothing] = -\/(value)
+  @inline def \/-[R](value: R): Disjunction[Nothing, R] = \/-(value)
+
+  @inline def left[L](value: L): Disjunction[L, Nothing] = -\/(value)
+  @inline def right[R](value: R): Disjunction[Nothing, R] = \/-(value)
+
+  def either[A, B, C](ac: A => C)(bc: B => C): A \/ B => C = _ match {
+    case -\/(l)  => ac(l)
+    case \/-(r) => bc(r)
+  }
 }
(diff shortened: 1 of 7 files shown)


base/src/main/scala/data/DisjunctionFunctions.scala
```diff
@@ -1,10 +1,17 @@
 package scalaz
 package data
 
+import Disjunction.{\/, \/-, -\/}
+
 trait DisjunctionFunctions {
-  @inline def -\/[L](value: L): Disjunction[L, Nothing] = Disjunction.-\/(value)
-  @inline def \/-[R](value: R): Disjunction[Nothing, R] = Disjunction.\/-(value)
+  @inline def -\/[L](value: L): Disjunction[L, Nothing] = -\/(value)
```
puffnfresh: Should `Disjunction` be covariant?

tonymorris: Variance annotations should never be used. Consequently, the use of `Nothing` is suspicious.

aloiscochard: I think it was initially invariant, but after discussing with @S11001001 and/or @vmarquez I got convinced it should be covariant.

I can't remember the details, but let's now settle on something and I'll add it in `CONTRIBUTING`.

Do we want to allow covariance on datatype? what was the strong reason in favor of it?

My initial thinking was that if the user always create proper constructor for it's ADT, there is no need for having variance on datatype. That is quite a cost for the user though.

Once we agree as a general rule, I'll update `CONTRIBUTING` and `Maybe` (which is currently invarriant) accordingly.

aloiscochard: I just added two commits removing variance on `Disjunction` and as it went fine, updating the `CONTRIBUTING` file accordingly.

S11001001: Meh, I've come around to finding it convenient for `\/`, because an expression typically fixes only one side, and the variance lets the unification be delayed. As long as you don't want to add any methods with higher-kinded args—i.e kill it as soon as it gets in the way of anything.

jbgi: @S11001001 could (a part of) that convenience be replaced by adding ```toRight```/```toLeft``` methods to ```IdOps```? ie.:
```scala
  final def toLeft[R]: A\/R =
    -\/(self)

  final def toRight[L]: L\/A =
    \/-(self)
```

puffnfresh: @jbgi I do think those help, just a little bit.

tonymorris: I am of the opinion that variance annotations should never be used. The penalty is never worth the apparent convenience.

aloiscochard: All right, so I'll merge this PR as is, and we'll go with a no-variance approach.

I feel like by having proper syntax we can minimize the cost of being explicit about both side so we can at least be explicit only once.

(I have in mind for example, introducing some `BoolOps` to make the case of `if (bool) left(a) else right(b)` digestible)

PR #1204, opened 2016-06-20
adding flatmap and map

vmarquez:
this is a bit more convenient than just folding. :)


aloiscochard:
You can't do it this way, because now we have a duplicate impl. with https://github.com/scalaz/scalaz/blob/series/8.0.x/base/src/main/scala/data/MaybeInstances.scala

Sounds like it's time to give https://github.com/typelevel/machinist a try :-)

vmarquez:
Closing because https://github.com/scalaz/scalaz/pull/1206 negates the need for this.

PR #1206, opened 2016-06-21
Feature/8 machinist

aloiscochard:
Here is a proposed `meta` module with `machinist` integration to provide zero-cost typeclass syntax.

It works in for-comprehension as well, sample usage:
```
scala> import scalaz.Prelude._
import scalaz.Prelude._

scala> just(42)
res0: scalaz.data.Maybe[Int] = Just(42)

scala> just(42).map(_ / 2)
res1: scalaz.data.Maybe[Int] = Just(21)

scala> for { x <- just(42) } yield (x / 2)
res2: scalaz.data.Maybe[Int] = Just(21)
```

This actually get resolved to explicit TC instance calls (more details here: http://typelevel.org/blog/2013/10/13/spires-ops-macros.html).

I'm unsure about the impact on compilation time but I think we should go for it, it's easy to change later and we won't pollute our data type with typeclass syntax methods.

### DIFF
--- a/base/src/main/scala/typeclass/Bind.scala
+++ b/base/src/main/scala/typeclass/Bind.scala
@@ -6,7 +6,7 @@ trait Bind[M[_]] {
   def flatMap[A, B](ma: M[A])(f: A => M[B]): M[B]
 }
 
-object Bind extends BindInstances {
+object Bind extends BindInstances with BindFunctions {
   def apply[F[_]](implicit F: Bind[F]): Bind[F] = F
 
   object syntax extends BindSyntax
--- a/base/src/main/scala/typeclass/BindFunctions.scala
+++ b/base/src/main/scala/typeclass/BindFunctions.scala
@@ -1,0 +1,6 @@
+package scalaz
+package typeclass
+
+trait BindFunctions {
+  def flatMap[M[_], A, B](ma: M[A])(f: A => M[B])(implicit M: Bind[M]): M[B] = M.flatMap(ma)(f)
+}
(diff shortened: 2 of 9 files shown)


vmarquez:
Good call.   it's also a really small dependency (LOC wise) so if we need or want to eventually roll our own it doesn't seem too difficult. 👍

tonymorris:
Others have expressed to me a concern with depending on typelevel code.

aloiscochard:
@tonymorris agree, but same apply to `kind-projector`.

The thing is, both seems pretty useful and we don't have to compel to any CoC to use them.
About `machinist` we could run our own later anyway if we end up doing lot of customization.

In this two specific cases, `kind-projector` and `machinist`, I feel like this is fine.

aloiscochard:
After looking in more details the only line currently used are the followings:
https://github.com/typelevel/machinist/blob/master/src/main/scala/spire/ops/Ops.scala#L276-L283

So we could arguably roll our own right now and adapt in future.

S11001001:
Let's just use machinist. We can fork later if necessary—free software licenses are irrevocable—but I don't think it will be.

puffnfresh:
I'm suspicious of macros at all, especially ones outside of the project. I'd vote to define it within the project and move on.

tonymorris:
FWIW, I have no opinion. I was just making it known that which has been privately expressed to me.

aloiscochard:
I agree in general with the approach to free software as described by @S11001001.

On the other hand, on a pure technical aspect I feel like we might actually have more code in our `Ops` than in the one already in `machinist`, and we'll have lot of custom shape.... and reuse very little (and trivial) logic.

So I agree with @puffnfresh, and I'll copy `unpack` and remove the dependencies.

In general, it feels like keeping control yourself of macro that will be used by our end user is a good practice, might even help to maintain binary compatibility as well.

PR #1217, opened 2016-07-03
Safe alternative type classe definitions.

jbgi:
also simplify type class templates usage by removing final modifier on derived methods traits.

### DIFF
--- a/base/src/main/scala/Identity.scala
+++ b/base/src/main/scala/Identity.scala
@@ -12,5 +12,6 @@ object Identity {
     override def ap[A, B](fa: Identity[A])(f: Identity[A => B]): Identity[B] = Identity(f.run.apply(fa.run))
     override def pure[A](a: A): Identity[A] = Identity(a)
     override def flatMap[A, B](oa: Identity[A])(f: A => Identity[B]): Identity[B] = f(oa.run)
+    override def flatten[A](ma: Identity[Identity[A]]): Identity[A] = ma.run
   }
 }
--- a/base/src/main/scala/data/ConstInstances.scala
+++ b/base/src/main/scala/data/ConstInstances.scala
@@ -4,20 +4,20 @@ package data
 import scalaz.typeclass._
 
 trait ConstInstances {
-  implicit def traverse[R]: Traversable[Const[R, ?]] = new TraversableClass[Const[R, ?]] {
+  implicit def traverse[R]: Traversable[Const[R, ?]] = new TraversableClass.Template[Const[R, ?]] {
     def map[A, B](ma: Const[R, A])(f: A => B): Const[R, B] = ma.retag
 
     def traverse[F[_], A, B](ta: Const[R, A])(f: A => F[B])(implicit F: Applicative[F]): F[Const[R, B]] =
       F.pure(ta.retag)
 
-    def sequence[F[_], A](ta: Const[R, F[A]])(implicit F: Applicative[F]): F[Const[R, A]] =
+    override def sequence[F[_], A](ta: Const[R, F[A]])(implicit F: Applicative[F]): F[Const[R, A]] =
       F.pure(ta.retag)
 
     def foldLeft[A, B](fa: Const[R, A], z: B)(f: (B, A) => B): B = z
 
     def foldRight[A, B](fa: Const[R, A], z: => B)(f: (A, => B) => B): B = z
 
-    def toList[A](fa: Const[R, A]): List[A] = Nil
+    override def toList[A](fa: Const[R, A]): List[A] = Nil
   }
 
   implicit def apply[R](implicit R: Semigroup[R]): Apply[Const[R, ?]] = new Apply[Const[R, ?]] {
--- a/base/src/main/scala/data/ConstInstances.scala
+++ b/base/src/main/scala/data/ConstInstances.scala
@@ -4,13 +4,10 @@ package data
 import scalaz.typeclass._
 
 trait ConstInstances {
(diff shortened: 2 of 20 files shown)


aloiscochard:
That looks great thank you JB! I'll review this carefully ASAP.

vmarquez:
Hi @jbgi I've been looking through this a bit.  Can you go into the motivation for changing?  It seems like this could potentially allow for easier code reuse, at the cost of a bit of inheritance and extra indirection.  Is that the main goal?

For instance, in Profunctor though, I can't think of a reason why something might be 'dimappable' but not lmapable or rmappable.  Perhaps in that case it makes less sense, but I think I can see why it is more useful for type classes such as Monad.  I'll keep reviewing, but it's possible this is solving other issues I am not aware of, so eager to hear more about this approach.

jbgi:
@vmarquez the `LeftRightXorDimap` trait is a bit misleading ; it is only a technical artefact: it is intended to be the scala equivalent of the haskell pragma
`{-# MINIMAL dimap | (lmap, rmap) #-}`

It does not mean to indicate that Profuctor cannot implement both. It is just there when you want a minimal definition: then you can mixin either `LeftRightMap` (and implement dimap yourself) or `Dimap`  (and implement left/righmap yourself) but you cannot mixin both (because it would not compile due to them implementing the same trait `LeftRightXorDimap` with a different type parameter)

Hope it makes the whole design clearer.

aloiscochard:
@jbgi I had finally a chance to look in detail, nice trick to use constrains to prevent invalid mixins!

The gain of safety comes with a complexity/maintenance cost but I think it's worth it if we can keep things nicely decoupled.

It might require a few iteration and we'll have to find conventions for organizing and naming the encoding, but I'm all for the added safety :+1: 

A few suggestions:
- Do we need both the defaults `typeclass` definition variant **and** all the defaults template variants? I think no, we should instead embrace the template system for what it is, as we did for the default method implementation: nothing in the typeclass, but instead in the template. In other words I think we should remove: `Choice.Left`, `Choice.Right`, `Profunctor.LeftRightMap`, ... and instead move the default method directly in their template. That will reduce the number of types and I think we keep same functionality as long as we promote and document template usage.
- We need a naming convention when there is template alternatives, we can just name them what they are `StrongClass.First`, `ChoiceClass.Left`, `TraversableClass.Sequence`, ... in addition with previous point I feel it should make the design more compact, separated and hopefully easier to understand.
- I would like to find a naming convention for those "constraint traits", just now, I feel like we won't have more than one per class... like, users we'll have to choose just one, anyway that will probably be most case, so for those we could just have a specific name we use in every situation like that, like `Alt[_]` (standing for Alternative).

How do you feel about those change @jbgi ?

Thanks for the great work, it's nice to see how safer we can get typeclass definition :-)

jbgi:
>Do we need both the defaults typeclass definition variant and all the defaults template variants?

Indeed, it was a bit too much, I removed alternative templates in my last commit. 

>nothing in the typeclass, but instead in the template

I think template should be limited to deriving inherited methods from super-class. when it is possible to derive methods from other methods of the same type class then we should do so at the typeclass level: templates can always mixin `Alt` traits in type class but sometimes we do not use templates, like in https://github.com/scalaz/scalaz/blob/series/8.0.x/base/src/main/scala/data/DownStarInstances.scala#L7
In that particular case both `{r,l}map` and `dimap` is defined but I think we should have been able to use [one of the `Alt` traits](https://github.com/scalaz/scalaz/pull/1217/files#diff-fc3b590d6c5d6e4fb918b3b27a4db4a1R12) instead if we wanted.

>we could just have a specific name we use in every situation like that, like Alt[_] (standing for Alternative). 

I renamed the traits to `Alt`. But the `Alt` traits cannot be one and only, because sometimes we can mix more than one, like in https://github.com/scalaz/scalaz/pull/1217/files#diff-7549a630a64b466c240406d15e5cdb71R7

aloiscochard:
@jbgi okay, I agree, that makes total sense to follow the path you suggest. So Alt on typeclasses and not on template.

I think it's getting there, thanks! I just have one small concern for which I'm not sure to have a good answer.

ATM, you name the alternative according to the default implementation you provide (ie. `Bind.Flatten`/`Bind.FlatMap`), I would have intuitively thought that if the name of the alternative is `FlatMap` it means I have to *implement* `flatMap` and not that I get it for free (so basically the opposite of what we have here)..

But maybe that's just me? and I can understand this can be seen as quite an arbitrary decision... so we should probably get vote from others as well.

Sure thing about `Alt`, I mean one trait per typeclass, not a global one indeed!

Great stuff!

jbgi:
>  I would have intuitively thought that if the name of the alternative is FlatMap it means I have to implement flatMap

I was undecided for the same reason. At one point I even had `FlatMap extends Alt[Flatten]` and `Flatten extends Alt[FlatMap]` :smile: . I finally decided to follow the "convention" used in templates, of naming the trait after the derived method name.

I think we should decide after what is most intuitive when implementing an instance:
do one decide upfront that he/she want to implement FlatMap and not flatten,
or do one decide after implementing one of flatmap/flatten than he/she want to derive the other method...
please vote ;)

jbgi:
@aloiscochard I apply your suggestion to the naming convention of Alt traits. It does make things clearer when mixing them in instance definitions.

aloiscochard:
I'll rebase and squash this one in a new PR, I think it's ready for final review. I'm very keen on merging this nice addition :-)

aloiscochard:
Merged as https://github.com/scalaz/scalaz/pull/1226

PR #1218, opened 2016-07-11
Add monad transformer laws.

TomasMikula:
From [Monad transformer (Wikipedia)](https://en.wikipedia.org/wiki/Monad_transformer):

```haskell
lift . return = return
lift (m `bind` k) = (lift m) `bind` (lift . k)
```


vmarquez:
Cool, I like the idea of testing liftM consistency.   Will review a bit more but so far 👍

puffnfresh:
Good stuff :+1:

PR #1219, opened 2016-07-13
Simplify syntax usage in `base`.

aloiscochard:
This simplify syntax usage by removing the `syntax` companion (which seems unnecessary, it was introduced by fear of polluting scope with std instances, but by encouraging Prelude usage it should not be an issue).

It does also move the `Prelude` in `base` which could introduce a cyclic dependency for which I'm unsure it will end well (feedback?).

The price to pay otherwise is explicit import of every module when working in `base`: 
https://github.com/scalaz/scalaz/commit/a1150001d02872fe9d481a355df1465d99e6a021

I kinda liked having prelude separated but I understand the concerns raised by @vmarquez .


aloiscochard:
ping @S11001001 @puffnfresh @runarorama

jbgi:
@aloiscochard maybe one way to avoid cyclic dependencies would be having `Prelude` extends `TypeClassPrelude` itself extending a `DataPrelude`.
Also to avoid cyclic dependencies between `typeclass` and `data`, typeclass instances and `BaseHierarchy`  could be moved into the `typeclass` package,  and `Leibniz` and `Identity` into `data`...

aloiscochard:
@jbgi I don't think that works in practice, typically one would like to use the `typeclass` Prelude when defining `data` instances.

jbgi:
@aloiscochard yeah. so maybe there should be a single `base` package, with the type classes and all (and only) the `data` that are used in type class definitions.

aloiscochard:
@jbgi yeah, I am generally enthusiast about such approach, I proposed `containers` as a module name for others general data structure outside of `base`.

But it seems like most people prefer a "one big module" approach (arguing the containers module would be rather small), we can anyway try to follow such practice by enforcing decoupling at package level vs module level (still we have to find names).

vmarquez:
I like this a lot.  It simplifies things a bit and seems to work well, and should allow us to get syntax in our implementations (if we can use a macro to not incur a performance penalty).   👍

tonymorris:
Modules should be sorted based on degree of compromise with scala.

aloiscochard:
@tonymorris I really this idea! seems a reasonable way to compromise on modularity.

I'll consider this as accepted without objection by the end of the week.

tonymorris:
I just don't think any other form of "break into modules" is viable.

aloiscochard:
I'll add a note in `CONTRIBUTING.md` about modularity as formalized here, and merge this one.

PR #1224, opened 2016-07-28
Example of using Codensity for resource management

wedens:
There is a problem in this example: you can yield closed resource. But solving this problem is not in scope of this example :)


aloiscochard:
:+1:

tpolecat:
Cool, I never had any idea what `Codensity` could be used for. 👍

PR #1226, opened 2016-07-31
Default implementation and alternatives for typeclasses.

aloiscochard:
Original work by @jbgi.

What I did since https://github.com/scalaz/scalaz/pull/1217

1.  Relax self constrain to be on the class trait instead of the template (as in https://github.com/scalaz/scalaz/commit/51f7fde3529b870221d0f6f8aa59bf107e14ce61)
2. Finally decide to move all default and alternative definition in the template companion (as in https://github.com/scalaz/scalaz/commit/3f22f078f237d68b25854469e9863fc00e4ab93b)

Reasoning for 1 is that, well, you never know... maybe one day someone will want to mix those default impl. without using the template itself (I can see that happening for diamond-like structure in the hierarchy), so let's be as relaxed as possible here.

Reasoning for 2, while I totally agree that those default and alternatives should be defined as relaxed as possible (as stated above), in term of usage and imports, I feel like it seems more appropriate to define them in the template companion.

It just feel more consistent (and nicely decoupled) to keep everything related to the definition of a class instance in the template (`FooClass`), while keeping the definition of the class clean (`Foo`).

The cost of that choice is that you have to do `import typeclass.FooClass._` to use the templates (default or alternatives) when defining an instance, otoh we won't pollute the namespace of code simply using the typeclass syntax and method using `import typeclass.Foo._`.

If we really feel like we should merge the default/alternatives in the companion of the class itself (`Foo`), then we can go by mixing a trait so we keep them in an other file... but then to be consistent it feel like we should provide all template feature there, and it seems quite a lot to put in the scope.

Which is why I propose the design you see here.

Hope that make sense, looking for feedback, particularly from Jean-Baptiste on the relaxed constraints. 

ping @S11001001 @tonymorris @vmarquez @puffnfresh


aloiscochard:
Here is a usage example, when defining `Maybe` instances: https://github.com/aloiscochard/scalaz/blob/3f22f078f237d68b25854469e9863fc00e4ab93b/base/src/main/scala/data/MaybeInstances.scala#L4-L9

base/src/main/scala/typeclass/MonadClass.scala
```diff
@@ -1,10 +1,16 @@
...
+
 trait MonadClass[M[_]] extends Monad[M] with BindClass[M] with ApplicativeClass[M] {
   final def monad: Monad[M] = this
 }
 
 object MonadClass {
-  trait Template[M[_]] extends MonadClass[M] with Monad.Map[M]
+  trait Template[M[_]] extends MonadClass[M] with Map[M]
+
+  trait Map[M[_]] extends FlatMap[M] { self: Monad[M] with Bind[M] with Applicative[M] with Functor[M] =>
```
jbgi: here we could remove the `Monad[M]` `self` constraint to be even more relaxed.

jbgi:
:+1: 
Relaxed self constraints looks good (I wanted to achieved that but hit a scalac bug when using extends directly, using self constraints apparently solve this),
as well as separating stuff needed for instance construction from the type class api.

vmarquez:
👍 

@aloiscochard if you want to fix this and merge go for it, otherwise if you don't have time I can re-pull/merge and close this one.   Let's get it in since i'm sitting on a couple PRs i'll want to open after this! :)

vmarquez:
If no one has any objections ( @aloiscochard ?) I'll open a new PR that has the conflicts resolved and merge it.

aloiscochard:
@vmarquez please go for it!

Pretty busy lately, so it's awesome if you can take care of this one :-) ty!

vmarquez:
closing because https://github.com/scalaz/scalaz/pull/1248 has these changes.

PR #1232, opened 2016-07-31
Liskov lift dat shit.

aloiscochard:
Following an original idea from @S11001001 (http://typelevel.org/blog/2014/03/09/liskov_lifting.html) and suggestion made by @puffnfresh on the mailing list (which I do discover I was not subscribed to since.... quite a while).

I'm convinced `Liskov.liftCvf` will be useful in other situation, Stephen gave me some ideas about it :)

### DIFF
--- a/base/src/main/scala/data/Maybe.scala
+++ b/base/src/main/scala/data/Maybe.scala
@@ -4,11 +4,11 @@ package data
 sealed abstract class Maybe[A] {
   final def fold[B](f: A => B, b: => B): B = this match {
     case Maybe.Just(a)  => f(a)
-    case Maybe.Empty()  => b
+    case _              => b
   }
 }
 
 object Maybe extends MaybeFunctions with MaybeInstances with MaybeSyntax {
-  final case class Empty[A]() extends Maybe[A]
+  final private[data] case object Empty extends Maybe[Nothing]
   final case class Just[A](a: A) extends Maybe[A]
 }
--- a/base/src/main/scala/data/MaybeFunctions.scala
+++ b/base/src/main/scala/data/MaybeFunctions.scala
@@ -1,16 +1,18 @@
 package scalaz
 package data
 
+import typeclass.Liskov._
+
 import Maybe.{Empty, Just}
 
 trait MaybeFunctions {
-  def empty[A]: Maybe[A] = Empty()
+  def empty[A]: Maybe[A] = liftCvf[Nothing, A, Maybe](isa[Nothing, A]).apply(Empty)
   def just[A](a: A): Maybe[A] = Just(a)
 
   def maybe[A, B](n: B)(f: A => B): Maybe[A] => B = _ match {
-    case Empty() => n
-    case Just(x) => f(x)
+    case Empty    => n
+    case Just(x)  => f(x)
   }
 
-  def fromOption[A](oa: Option[A]): Maybe[A] = oa.fold[Maybe[A]](Empty[A])(Just(_))
+  def fromOption[A](oa: Option[A]): Maybe[A] = oa.fold[Maybe[A]](empty[A])(Just(_))
 }
--- a/base/src/main/scala/typeclass/LiskovFunctions.scala
+++ b/base/src/main/scala/typeclass/LiskovFunctions.scala
@@ -9,6 +9,7 @@ trait LiskovFunctions {
   def isa[A, B >: A]: A <~< B = Liskov.refl
 
   /** http://typelevel.org/blog/2014/03/09/liskov_lifting.html **/
+  def liftCtf[F[_]: Contravariant, A, B](a: A <~< B): F[B] <~< F[A] = a.asInstanceOf[F[B] <~< F[A]]
   def liftCvf[A, B, F[_]: Functor](a: A <~< B): F[A] <~< F[B] = a.asInstanceOf[F[A] <~< F[B]]
 
   /**Subtyping is antisymmetric */
--- a/base/src/main/scala/typeclass/LiskovFunctions.scala
+++ b/base/src/main/scala/typeclass/LiskovFunctions.scala
@@ -1,0 +1,20 @@
+package scalaz
+package typeclass
+
+import data.===
+import Liskov.<~<
+
+trait LiskovFunctions {
+  /**Lift Scala's subtyping relationship */
+  def isa[A, B >: A]: A <~< B = Liskov.refl
+
+  /** http://typelevel.org/blog/2014/03/09/liskov_lifting.html **/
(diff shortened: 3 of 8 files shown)


base/src/main/scala/data/Maybe.scala
```diff
@@ -4,11 +4,11 @@ package data
...
   final def fold[B](f: A => B, b: => B): B = this match {
     case Maybe.Just(a)  => f(a)
-    case Maybe.Empty()  => b
+    case _              => b
   }
 }
 
 object Maybe extends MaybeFunctions with MaybeInstances with MaybeSyntax {
-  final case class Empty[A]() extends Maybe[A]
+  final case class Empty[A]() extends Maybe[Nothing]
```
puffnfresh: Can this now become an object?

aloiscochard: Yes! Much nicer.. I'll force push now.

base/src/main/scala/data/Maybe.scala
```diff
@@ -4,11 +4,11 @@ package data
...
   final def fold[B](f: A => B, b: => B): B = this match {
     case Maybe.Just(a)  => f(a)
-    case Maybe.Empty()  => b
+    case _              => b
   }
 }
 
 object Maybe extends MaybeFunctions with MaybeInstances with MaybeSyntax {
-  final case class Empty[A]() extends Maybe[A]
+  final case object Empty extends Maybe[Nothing]
```
puffnfresh: Awesome! :+1:

base/src/main/scala/data/MaybeFunctions.scala
```diff
@@ -1,16 +1,18 @@
...
 
 trait MaybeFunctions {
-  def empty[A]: Maybe[A] = Empty()
+  def empty[A]: Maybe[A] = liftCvf[Nothing, A, Maybe](isa[Nothing, A]).apply(Empty)
   def just[A](a: A): Maybe[A] = Just(a)
 
   def maybe[A, B](n: B)(f: A => B): Maybe[A] => B = _ match {
-    case Empty() => n
-    case Just(x) => f(x)
+    case Empty    => n
```
S11001001: NB: A ~ Nothing in this context. As long as you're comfortable with that, liftCvf is fine.

base/src/main/scala/typeclass/LiskovInstances.scala
```diff
@@ -0,0 +1,25 @@
+package scalaz
+package typeclass
+
+import scala.language.implicitConversions
+
+import Liskov.<~<
+
+trait LiskovInstances {
+  /**Lift Scala's subtyping relationship */
+  implicit def isa[A, B >: A]: A <~< B = new (A <~< B) {
```
S11001001: This must not be implicit, and can be = refl.

base/src/main/scala/typeclass/LiskovFunctions.scala
```diff
@@ -0,0 +1,13 @@
+package scalaz
+package typeclass
+
+import Liskov.<~<
+
+trait LiskovFunctions {
+  /** http://typelevel.org/blog/2014/03/09/liskov_lifting.html **/
+  def liftCvf[A, B, F[_]: Functor](a: A <~< B): F[A] <~< F[B] = a.asInstanceOf[F[A] <~< F[B]]
```
S11001001: Please provide liftCtf as well; it is more primitive.

base/src/main/scala/typeclass/LiskovFunctions.scala
```diff
@@ -0,0 +1,13 @@
...
+
+  /**Subtyping is transitive */
+  def trans[A, B, C](f: B <~< C, g: A <~< B): A <~< C =
+    g.subst[λ[`-α` => α <~< C]](f)
```
S11001001: You may wish to provide antisymmetry too, `(A <~< B, B <~< A) => (A === B)`. One point in opposition: I don't think it can be written with `subst`.

base/src/main/scala/typeclass/LiskovFunctions.scala
S11001001: Also, you can use this cast to get stuck in Scalazzi-safe code, unless you seal Liskov.

aloiscochard: Oh that's why it was sealed! I did not understood the reasoning while porting, I shouldn't have changed that...

Ty!

aloiscochard:
Ouch, okay, I did not understood the implication initially thanks @S11001001 for the NB.

I'm much less comfortable with such unsoundness:

```
  def unsound[A, B](ma: Maybe[A], a: A, f: A => B): B = ma match  {
    case Empty    => a
    case Just(a)  => f(a)
  }
```

In action:
```
scala> unsound[Int, String](empty, 42, _.toString)
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
```

Honestly that looks very dangerous to me, any thought ? @tonymorris @puffnfresh @vmarquez

puffnfresh:
@aloiscochard I'm not sure. I don't think we should expose unapply in any case. I definitely don't like having to be careful, though.

aloiscochard:
@puffnfresh actually, in the meantime I realize we should really not expose Unapply (in fact the type constructors should be `private[data]` or something) ... according to this, it make the unsound code impossible to write and I'm again totally fine with this change. :+1: 

We should of course carefully review any usage of it, I'll add a note in `CONTRIBUTING` about limitation of usage.

aloiscochard:
Okay, I've added a few changes thanks to your feedback.

@S11001001 mind taking an other look? ty!

base/src/main/scala/data/Maybe.scala
```diff
@@ -4,11 +4,11 @@ package data
...
   final def fold[B](f: A => B, b: => B): B = this match {
     case Maybe.Just(a)  => f(a)
-    case Maybe.Empty()  => b
+    case _              => b
   }
 }
 
 object Maybe extends MaybeFunctions with MaybeInstances with MaybeSyntax {
-  final case class Empty[A]() extends Maybe[A]
+  final private[data] case object Empty extends Maybe[Nothing]
```
aloiscochard: Lost in a force push, but @puffnfresh was asking about hiding `Just` as well here.

For which I say that I'm waiting on the result of investigation with optics, to then choose an appropriate strategy and review all datatype constructors.

jbgi:
A potential benefit of hiding sub-classes/unapply is that we could provide lazy constructors, like in https://github.com/aloiscochard/scalaz/pull/1

PR #1237, opened 2016-08-03
Make swap a member of Arrow, so that an instance of Arrow may override `swap`

Atry:


xuwei-k:
```
[info] scalaz-core: found 1 potential binary incompatibilities while checking against org.scalaz:scalaz-core_2.11:7.2.4 
[error]  * method swap()java.lang.Object in trait scalaz.Arrow is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scalaz.Arrow.swap")
```

https://travis-ci.org/scalaz/scalaz/jobs/149491928#L892

PR #1238, opened 2016-08-04
Make swap a method of Arrow, so that an instance of Arrow may override `swap`

Atry:
This is a continuation of #1237


core/src/main/scala/scalaz/Arrow.scala
```diff
@@ -29,12 +29,13 @@ trait Arrow[=>:[_, _]] extends Split[=>:] with Strong[=>:] with Category[=>:] {
   def >>>[A, B, C](fab: (A =>: B), fbc: (B =>: C)): (A =>: C) =
     compose(fbc, fab)
 
+  /** Swaps a pair. */
+  def swap[X, Y] = arr[(X, Y), (Y, X)] {
```
xuwei-k: Could you add the explicit return type?

PR #1239, opened 2016-08-12
Dequeue.toBackIList bugfix

swalker2m:
This PR proposes a small bugfix to `Dequeue.toBackIList`.  The issue is that when working with a `FullDequeue` with more than 2 elements, the order of the resulting `IList` is incorrect as shown in this REPL session:

````
scala> val empty = Dequeue.empty[Char]
empty: scalaz.Dequeue[Char] = EmptyDequeue

scala> ('a' +: 'b' +: 'c' +: empty).toIList
res0: scalaz.IList[Char] = [a,b,c]

scala> ('a' +: 'b' +: 'c' +: empty).toBackIList
res1: scalaz.IList[Char] = [c,b,a]

scala> ('a' +: 'b' +: 'c' +: 'd' +: empty).toIList
res2: scalaz.IList[Char] = [a,b,c,d]

scala> ('a' +: 'b' +: 'c' +: 'd' +: empty).toBackIList
res3: scalaz.IList[Char] = [d,b,c,a]
````

In `res3` the `front` list of the `FullDequeue` is `a :: b :: c :: Nil`.  The head correctly becomes the last element but `front.tail` must be reversed.

### DIFF
--- a/core/src/main/scala/scalaz/Dequeue.scala
+++ b/core/src/main/scala/scalaz/Dequeue.scala
@@ -108,7 +108,7 @@ sealed abstract class Dequeue[A] {
   def toBackIList: IList[A] = this match {
     case EmptyDequeue() => IList.empty
     case SingletonDequeue(a) => ICons(a, IList.empty)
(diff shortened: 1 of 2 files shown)


tpolecat:
Oops. 👍

Issue #1225, opened 2016-07-29
Missing type classes in Isomorphism


kcsongor:
I noticed that while the Isomorphism package provides a large set of derivable type classes through isomorphisms, some are missing.

Is there a particular reason why, for example, IsomorphismMonadReader isn't defined? I find that this can be fairly useful in practice.

I could make a PR to add these, and it would also be nice to have a list of the missing ones.

kcsongor:
It's more of a question than a request: would it be accepted if I made a PR adding isomorphisms for type classes like MonadReader?

jedws:
Csongor,

There isn't a rigorous approach to the inventory of instances. People
notice that useful ones are missing and then add them.

I'm pretty sure your PR would be welcome, as would any attempt at an
inventory of what is and isn't there.

cheers,
jed.

kcsongor:
Using the information in the build script to determine the traits that correspond to type-classes, and diffing that with the existing Isomorphism instances, I think the following type-classes have no Isomorphism instance defined for them:

Align
Arrow
Associative
Catchable
Category
Choice
ComonadStore
Compose
Cozip
Divide
Divisible
Enum
InvariantFunctor
IsEmpty
LiftControlIO
LiftIO
MonadControlIO
MonadError
MonadIO
MonadReader
MonadState
MonadTell
Nondeterminism
PlusEmpty
ProChoice
Profunctor
Resource
Split
Strong
Unzip
Zip

I'm going to make a PR in the next couple of days, when I get around to adding these. Some of them might be tricky, but I will see.

kcsongor:
Note: `PlusEmpty` does have an instance defined for it, but instead of using the `IsomorphismPlusEmpty` naming convention, it is (somewhat confusingly) called `IsomorphismEmpty`

PR #1243, opened 2016-08-18
Add missing Isomorphism instances (#1225)

kcsongor:
This adds Isomorphism*s for the following type classes:
- [x] Align
- [x] Arrow
- [x] Associative
- [x] Catchable
- [x] Category
- [x] Choice
- [x] ComonadStore
- [x] Compose
- [x] Cozip
- [x] Divide
- [x] Divisible
- [x] Enum
- [x] InvariantFunctor
- [x] IsEmpty
- [x] LiftControlIO
- [x] LiftIO
- [x] MonadControlIO
- [x] MonadError
- [x] MonadIO
- [x] MonadReader
- [x] MonadState
- [x] MonadTell
- [x] Nondeterminism
- [x] PlusEmpty (see note below)
- [x] ProChoice
- [x] Profunctor
- [x] Resource
- [x] Split
- [x] Strong
- [x] Unzip
- [x] Zip

A few things to note:
- `PlusEmpty` did have an iso already, but instead of `IsomorphismPlusEmpty`, this is called `IsomorphismEmpty`. I did not change it, but I think it should be changed.
- There is a separate `Isomorphism.scala` in the `effect` package for the `effect` (MonadIO and friends) type classes.
- I added a `hoistRunInBase` to IO.scala, which turns a `RunInBase[F, X]` to `RunInBase[G, X]`, given that there is a natural isomorphism between F and G - this is useful for writing `IsomorhpismLiftControlIO`. `hoist` might not be the best name for this, since it is symmetric.
- `IsomorphismAssociative` - I'm almost certain that this one requires a `Bifunctor[G]` instance (along with `Associative[G]`) so that we can map the natural transformations over both holes, which is needed for converting between `F` and `G`, as `Associative` assumes a nested type structure. Note that `Profunctor[G]` would also work here, since we we have a natural isomorphism, and that just means that the other direction would need to be mapped over the first hole, if that hole was contravariant. However, having looked at the existing `Associative` instances, they seem to be traditionally Bifunctors (Either and Tuple2).
- Added some `override` modifiers to the existing functions to keep things unified.

### DIFF
--- a/core/src/main/scala/scalaz/Isomorphism.scala
+++ b/core/src/main/scala/scalaz/Isomorphism.scala
@@ -26,7 +26,7 @@ sealed abstract class IsomorphismsLow0 extends IsomorphismsLow1 {
   }
 }
 
-sealed abstract class Isomorphisms extends IsomorphismsLow0{
+sealed abstract class Isomorphisms extends IsomorphismsLow0 {
 
   /**Isomorphism for arrows of kind * -> * -> * */
   trait Iso[Arr[_, _], A, B] {
@@ -184,8 +184,6 @@ trait IsomorphismSemigroup[F, G] extends Semigroup[F] {
 trait IsomorphismMonoid[F, G] extends Monoid[F] with IsomorphismSemigroup[F, G] {
   implicit def G: Monoid[G]
 
-  def iso: F <=> G
-
   def zero: F = iso.from(G.zero)
 }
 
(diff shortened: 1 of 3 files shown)


vmarquez:
Nice.  Looking good so far, going to review more in depth before an official +1 but thanks a lot for this.

kcsongor:
A minor thing that I will address tonight: I said that an `IsomorphismAssociative` instance needs a Bifunctor on G. I take this back, composing Liskov properly should do the job, without requiring the container to be a `Bifunctor`.
Starting out from the reflexive law, then constructing Liskov arrows from the natural transformations `F <~>`, then lifting them into the reflexive arrow, I'm certain this is possible.

kcsongor:
My earlier idea of using Liskov arrows wouldn't work, because those are strictly concerned with subtype relations, and don't compose with natural transformations, as I had originally thought :(

So as of yet, I don't have a better idea than the Bifunctor[G] instance

xuwei-k:
merged 3b6926e0d1480fc86821424094d39ab646a776eb

PR #1245, opened 2016-08-19
Add transformer instance for RegionT (via Hoist)

kcsongor:
RegionT is a legal monad transformer, but it didn't have a MonadTrans instance defined, so I added it.


kcsongor:
Somewhat unrelated to this PR, but there's an isomorphism between `RegionT[S, P[_], A]` and `Kleisli[P, IORef[List[RefCountedFinalizer]], A]` (by definition), and so this class could benefit from
https://github.com/scalaz/scalaz/pull/1243.

kcsongor:
The build fail doesn't seem relevant?

vmarquez:
@xuwei-k can you restart travis?  The error seems seems unrelated to this PR. (one is a nondeterminism deadlock error and the other is a node.js pipe issue)

Issue #1244, opened 2016-08-19
Monad transformer map*T functions


kcsongor:
The major Haskell monad transformer libraries (mtl, transformers) define functions like `mapExceptT` and `mapStateT`, that map functions working on the underlying representations to functions acting on the transformers themselves.

For example: `mapExceptT :: (m (Either e a) -> n (Either e' b)) -> ExceptT e m a -> ExceptT e' n b` (from mtl).

The implementations of these functions are of course trivial, which makes me wonder - why are they not present in scalaz?
From a practical perspective, they can make writing certain type class instances much easier that are of the form `instance MonadTypeClass m => MonadTypeClass (ExceptT e m)`, where MonadTypeClass has some *tricky* functions, i.e. ones that aren't of the 'simply liftable' form.
(MonadCatchIO comes to mind)

(Excuse my use of Haskell syntax, it's too late to get the Scala one right).

puffnfresh:
Yes, these are often lacking and it's very annoying :+1:

kcsongor:
I'll make a PR then in the next few days

kcsongor:
And while I'm at it, I'll try to find places where it could make existing code more concise (I already know of one example, which is an open PR)

PR #1246, opened 2016-08-19
Add mapT to monad transformers (#1224)

kcsongor:
Added the `mapT` functions to the following transformers, as explained in https://github.com/scalaz/scalaz/issues/1244

- [x] ExceptT
- [x] KleisliT
- [x] ListT
- [x] MaybeT
- [x] OptionT
- [x] RWST
- [x] RegionT
- [x] StateT
- [x] TheseT
- [x] UnwriterT
- [x] WriterT

- The `hoist` function is a special case of `mapT` (`hoist` is `mapT` with natural transformation), so I rewrote those.
- Also, updated TheseT's MonadCatchIO instance to be more concise, using mapT.
- `Kleisli` and `StateT` already had the functionality, called `mapK`, where the `K` I think comes from `Kleisli` (confusingly). For these, I just provided a `mapT` alias while keeping the old ones for compatibility.

### DIFF
--- a/core/src/main/scala/scalaz/EitherT.scala
+++ b/core/src/main/scala/scalaz/EitherT.scala
@@ -71,6 +71,9 @@ final case class EitherT[F[_], A, B](run: F[A \/ B]) {
   def map[C](f: B => C)(implicit F: Functor[F]): EitherT[F, A, C] =
     EitherT(F.map(run)(_.map(f)))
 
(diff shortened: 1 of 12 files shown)


kcsongor:
https://github.com/scalaz/scalaz/pull/1161 could immediately benefit from this

vmarquez:
makes sense, easy 👍 for me.

PR #1250, opened 2016-09-01
Flip arguments of tailrecM to improve type inference.

TomasMikula:
This is a proposal to flip the order of `tailrecM` parameters to improve type inference.

The original signature is

```scala
def tailrecM[A, B](f: A => F[A \/ B])(a: A): F[B]
```

The signature proposed in this PR is

```scala
def tailrecM[A, B](a: A)(f: A => F[A \/ B]): F[B]
```

The new signature allows to pass the function argument `f` as a lambda, without having to explicitly specify type arguments of `tailrecM` or `f`'s parameter type.

This PR is mostly a mechanical flipping of arguments, though I did inline some local `def`s as lambdas to demonstrate the gained type inference (see e.g. changes to `Free`/`FreeT`).

I think this flipped argument order is more consistent with the argument order of `bind`, which in `scalaz` is also flipped compared to Haskell/PureScript.

This is, of course, a breaking change.

FWIW, I [suggested](https://github.com/typelevel/cats/issues/616#issuecomment-219555071) the same in `cats` and they went with it.

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Task.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Task.scala
@@ -304,7 +304,7 @@ object Task {
       }
       def fail[A](e: Throwable): Task[A] = new Task(Future.now(-\/(e)))
       def attempt[A](a: Task[A]): Task[Throwable \/ A] = a.attempt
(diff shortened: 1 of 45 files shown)


xuwei-k:
:+1:

PR #1253, opened 2016-09-06
Update comment at ISet#map about Functor requirements.

TomasMikula:
The main point I'm trying to make is that it is a requirement on the _functions_ passed to `map`, rather than a requirement on the `Order` instance of the target type (as the previous comment was suggesting).

### DIFF
--- a/core/src/main/scala/scalaz/ISet.scala
+++ b/core/src/main/scala/scalaz/ISet.scala
@@ -379,9 +379,25 @@ sealed abstract class ISet[A] {
     }
 
   /**
-    * For the `Functor` composition law to hold it is important that the `Order[B]` is substitutive for the `Order[A]` –
-    * that is, that the `Order[B]` should be __no stronger__, it should not distinguish two `B` instances that would
-    * be considered as equal `A` instances.
+    * The `Functor` composition law only holds for functions that preserve
+    * equivalence, i.e. for functions `f` such that
+    *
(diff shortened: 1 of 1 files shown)


core/src/main/scala/scalaz/ISet.scala
```diff
@@ -379,9 +379,23 @@ sealed abstract class ISet[A] {
...
+    * _equality_, i.e. the finest equivalence, i.e. satisfying the
+    * _substitution property_ (which is the above property quantified over
+    * all `f`, see [[https://en.wikipedia.org/wiki/Equality_(mathematics)#Some_basic_logical_properties_of_equality Wikipedia page on Equality]]),
+    * the requirement holds for all `f` by definition.
+    *
+    * When `Order` instances are viewed as "mere" equivalences (as opposed
+    * to equalities), we can loosely say that `ISet` is an (endo-)functor
+    * in the category _Equiv_ of sets with an equivalence relation (where
+    * the morphishms are equivalence-preserving functions, i.e. exactly the
+    * functions satisfying the above requirement).
```
S11001001: Could you say something like "By contrast, [[Functor]] instances relate to the \_Scal\_ category...", so it is clear that the former functor doesn't meet the latter's definition?

TomasMikula: Yeah, that will clarify it a bit more. Done.

PR #1254, opened 2016-09-06
[WIP] Workaround for ambiguous implicits issue in MTL-style, fixes #1110

### LINKED ISSUE #1110: MTL-style doesn't seem to work in Scala
Let's say you're trying to write methods which uses MTL-style classes such as `MonadError` and `MonadReader`. Good job!

But if you go to use more than one constraint on a method, you get ambiguous implicits:

```
scala> import scalaz._
import scalaz._

scala> import scalaz.syntax.monad._
import scalaz.syntax.monad._

scala> def x[F[_], A](implicit E: MonadError[F, Int], R: MonadReader[F, String]) = R.ask.flatMap(E.raiseError(_.length))
<console>:13: error: ambiguous implicit values:
 both value R of type scalaz.MonadReader[F,String]
 and value E of type scalaz.MonadError[F,Int]
 match expected type scalaz.Bind[F]
       def x[F[_], A](implicit E: MonadError[F, Int], R: MonadReader[F, String]) = R.ask.flatMap(E.raiseError(_.length))
                                                                                     ^
```

The problem is that both `MonadError` and `MonadReader` extend `Monad`, so there's 2 possible instances to choose from. They should be consistent with each-other, but scalac doesn't know that.

What can we do about this problem?

adelbertc:
Work in progress, but putting it up to get some feedback.

This is a proposed workaround to make MTL-style a bit nicer in the current subtype encoding. Assuming nobody has objections I will do similar for the other MTL type classes:

- [x] MonadError
- [x] MonadListen
- [ ] MonadReader
- [ ] MonadState
- [x] MonadTell

Unsure if we want it for these:

- [ ] ApplicativePlus/MonadPlus
- [ ] BindRec

Pinging @puffnfresh @aloiscochard

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Task.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Task.scala
@@ -287,8 +287,8 @@ class Task[+A](val get: Future[Throwable \/ A]) {
 
 object Task {
 
-  implicit val taskInstance: Nondeterminism[Task] with BindRec[Task] with Catchable[Task] with MonadError[Task,Throwable] =
-    new Nondeterminism[Task] with BindRec[Task] with Catchable[Task] with MonadError[Task, Throwable] {
+  implicit val taskInstance: Nondeterminism[Task] with BindRec[Task] with Catchable[Task] with MonadError[Task,Throwable] with Monad[Task] =
+    new Nondeterminism[Task] with BindRec[Task] with Catchable[Task] with MonadError[Task, Throwable] with Monad[Task] {
       val F = Nondeterminism[Future]
       def point[A](a: => A) = Task.point(a)
       def bind[A,B](a: Task[A])(f: A => Task[B]): Task[B] =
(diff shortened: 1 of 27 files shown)


tonymorris:
:+1:

tonymorris:
Also, :+1: all of the above.

puffnfresh:
Not a great encoding, but I think it's all we can do to fix #1110 right now. :+1:

TomasMikula:
Perhaps a compilation test should be included (e.g. the example from #1110).

adelbertc:
@TomasMikula Yep I'll add that once I'm done.

aloiscochard:
Agree with @puffnfresh, :+1:

adelbertc:
Still WIP folks - will comment again once it's ready for review.

While doing this change I noticed that I need to be careful. As an example `Kleisli`'s `Monad` instance is defined through it's `MonadReader` instance, like:

```scala
val kleisliInstance[F[_]: Monad, E]: MonadReader[Kleisli[F, E, ?], E] = ...
```

with the change `MonadReader[F]` no longer automagically implies `Monad[F]` so `Monad[Kleisli[Task, Int, ?]]` would not resolve. This is fixed by changing the signature above to:

```scala
val kleisliInstance[F[_]: Monad, E]: MonadReader[Kleisli[F, E, ?], E] with Monad[Kleisli[F, E, ?]] = ...
```

I've tried to cover everything by grep-ing with each MTL type class. Hopefully I don't miss anything :|

adelbertc:
Closing in favor of new PR: https://github.com/scalaz/scalaz/pull/1262

PR #1256, opened 2016-09-11
A generic syntax and meta-programming infrastructure for optimization (of functions and later optics).

aloiscochard:
Some of you might be aware of my initial investigation regarding static fusion of optics (ref: https://github.com/scalaz/scalaz/commit/af8c172e58c1d7a4e44be7bb95456d60b0d8b880), which was encoded using meta/type-level/generic-programming... and as @puffnfresh pointed out, that was a lot of *programming...

At the end of my exploration I realized the exact same approach could be used to give function composition I nicer syntax (what that I did always dream of, not the reversed one from the std-lib), while retaining maximum performance.

I then decided to redesign everything from scratch and here is the results in the form of an optimized function composition syntax.

```
[info] Benchmark                   Mode  Cnt       Score       Error  Units
[info] FunctionsCompose.baseline  thrpt   20  815039.970 ± 74897.254  ops/s
[info] FunctionsCompose.scala     thrpt   20  690074.313 ± 75335.130  ops/s
[info] FunctionsCompose.scalaz    thrpt   20  817839.772 ± 72035.421  ops/s
```

source: https://github.com/scalaz/scalaz/commit/aa2e5470498c768ce9b7bb35c1508fc16286fdd4

Of course, it might looks like a lot of meta-programming for a small benefit (~8-10%), but please keep mind that everything besides `ComposerFunctions` (ref: https://github.com/scalaz/scalaz/commit/df45d2ba295e7d098b25dbc39c963ff37c0f663a) will be used as is to implement static fusion of optics (which show a much higher performance benefit according to my prototype).

Finally, I would like to emphasis on few key points of the current encoding vs an approach that would imply type-level programming:

1. Better compilation time, I don't have metrics to prove it but as we have full control over the type checking we do it better.

2. Friendlier error messages, that is a **big** selling point for this approach (and will really shine when using it for optics), as we have complete control over the typing/code-gen phases we can retrieve nice message with context.

3. Very few data-structure leak in the run-time space, as all intermediate data-structure are removed during macro expansion there is no `HList` or other newtype returned to the user during the process.

4. Instead of encoding twice the typechecking logic (once as TLP and once in the macro), we just do it once and can reuse it in both context (which result in **2** - better error messages)

Actually when I say "better error messages", we should keep in mind that traditional fun-dep approach does not give you any messages at all, but simply a "method does not exist" in case of a pimp, or "implicit not found" in other context.

```scala
  val f: Int => Double = _ + 0.5
  val g: Double => String = _.toString
  val h: String => Int = _.toDouble.toInt + 10

  (f, h, g).compose
```

```
[error] /home/alois/oss/scalaz8/benchmarks/src/main/scala/FunctionsCompose.scala:36: can't compose;
[error]     Int => Double
[error]   with
[error]     String => Int
[error]   as
[error]     Double is not a String
[error]   (f, h, g).compose
[error]       ^
[error] one error found
```

Note: I took the extra of using colors in compiler error in order to highlight semantic information (ref: https://pbs.twimg.com/media/CsFcy9XXgAEqqgD.jpg:large)

ping @vmarquez @tonymorris @S11001001 @puffnfresh @jbgi @julien-truffaut


base/src/main/scala/system/ComposerMacros.scala
```diff
@@ -0,0 +1,107 @@
...
+  def materializeOps[XS: c.WeakTypeTag, O: c.WeakTypeTag](c: Context): c.Expr[Ops[XS, O]] = {
+    import c.universe._
+
+    val tpe = c.weakTypeOf[XS]
+    unpack(c)(tpe) match {
+      case Left(err) =>
+        // We abort the implicit resolution, giving a chance for others to resolve.
+        c.abort(c.enclosingPosition, err)
+      case Right(ts) =>
+        val x = ts(0)
```
jbgi: looks like x, y are not used (leftover?)

jbgi:
nice! I also think it is the right approach to squeeze max perf. But then handling all corner cases is trickier than with type-level generic programming. Eg.
```scala
  def main(args: Array[String]): Unit = {
     val c = (f, g, h, f, ((x: Double) => x.toString), h, f, g, h, f, g, h).compose : Int => Int
     c(42)
  }
```
gives
```
Exception in thread "main" java.lang.BootstrapMethodError: java.lang.NoSuchMethodError: scalaz.benchmarks.FunctionsCompose$.scalaz$benchmarks$FunctionsCompose$$$anonfun$6$adapted(Ljava/lang/Object;)Ljava/lang/String;
	at scalaz.benchmarks.FunctionsCompose$$anon$2.apply$mcII$sp(FunctionsCompose.scala:29)
	at scalaz.benchmarks.FunctionsCompose$.main(FunctionsCompose.scala:30)
	at scalaz.benchmarks.FunctionsCompose.main(FunctionsCompose.scala)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
Caused by: java.lang.NoSuchMethodError: scalaz.benchmarks.FunctionsCompose$.scalaz$benchmarks$FunctionsCompose$$$anonfun$6$adapted(Ljava/lang/Object;)Ljava/lang/String;
	at java.lang.invoke.MethodHandleNatives.resolve(Native Method)
	at java.lang.invoke.MemberName$Factory.resolve(MemberName.java:975)
	at java.lang.invoke.MemberName$Factory.resolveOrFail(MemberName.java:1000)
	at java.lang.invoke.MethodHandles$Lookup.resolveOrFail(MethodHandles.java:1389)
	at java.lang.invoke.MethodHandles$Lookup.linkMethodHandleConstant(MethodHandles.java:1745)
	at java.lang.invoke.MethodHandleNatives.linkMethodHandleConstant(MethodHandleNatives.java:477)
```

PR #1259, opened 2016-09-13
Only memoize EphemeralStream on request (#memoized).

S11001001:
I agree with @dmitryleskov's [conclusions about the standard use of weak memoization for `EphemeralStream`](http://blog.dmitryleskov.com/programming/scala/stream-hygiene-iii-scalaz-ephemeralstream-fills-quite-a-canyon/); it's unnecessary churn for most applications, especially when building "wholemeal"—memoizing the final stream may be desirable, but almost certainly not the intermediates.  And it's impossible to "opt out".

This flips the memoization to "opt in"; you can invoke `.memoized` and the whole stream will be memoized.  Otherwise, it will be treated just like an ordinary by-name cons list.

There are no binary compatibility breaks, but the behavior change may discourage backporting.

Incidentally, `EphemeralStream` is worth comparing to #1193.

### DIFF
--- a/core/src/main/scala/scalaz/EphemeralStream.scala
+++ b/core/src/main/scala/scalaz/EphemeralStream.scala
@@ -142,6 +142,10 @@ sealed abstract class EphemeralStream[A] {
 
   def zipWithIndex: EphemeralStream[(A, Int)] =
     zip(iterate(0)(_ + 1))
+
+  def memoized: EphemeralStream[A] =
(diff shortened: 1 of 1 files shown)


tonymorris:
many :+1:

Also `CorecursiveList` is a useful addition.

aloiscochard:
:+1:

vmarquez:
oh. yes please.

PR #1260, opened 2016-09-13
Use transform instead of map where possible on stdlib Maps

S11001001:
`Map#transform` only allows value updating, so potentially can be implemented without reindexing keys.  As far as I know, the standard library doesn't do this, but [it is possible](https://bitbucket.org/S11001001/record-map/src/4e9a60fb15f5b4567a379e14837826d1d465fa39/src/main/scala/Record.scala?at=default&fileviewer=file-view-default#Record.scala-180:188).  It's also a little nicer to not bother to return the key as well.

This makes the `Map` functions use `transform` where they were calling `map` but always returning the same key.


tonymorris:
:+1:

Also, one day exposing the `Map` structure might exist, so these optimisations could be written externally. @gpampara wrote a `Map` to assist this type of problem.

aloiscochard:
:+1:

PR #1263, opened 2016-09-25
Add lefts and rights functions to MonadPlus

jdnavarro:
I was reimplementing something similar to Haskell's `lefts` and `rights` but @tpolecat pointed out that there was already `separate` in MonadPlus which already does the same and it's more general.

`lefts` and `rights` were already present in `MonadPlus` but were private, so this commit exposes them publicly.


tpolecat:
👍

PR #1270, opened 2016-10-14
Add getOrElseF function to Option and Maybe.

gwils:
This function is useful when you have a Maybe and you say "If I already have a result, that's great, otherwise let's do some applicative work to try to get to a result"

I am open to suggestions for a better name for this thing, but a variant of it existed in MaybeT under this name, so I have opted for consistency for now.

### DIFF
--- a/core/src/main/scala/scalaz/Maybe.scala
+++ b/core/src/main/scala/scalaz/Maybe.scala
@@ -86,6 +86,9 @@ sealed abstract class Maybe[A] {
   final def orElse(oa: => Maybe[A]): Maybe[A] =
     cata(_ => this, oa)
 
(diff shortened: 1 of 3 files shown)


xuwei-k:
This is not binary compatible. Please re-submit to `series/7.3.x` branch.

```
[error]  * method getOrElseF(scala.Option,scala.Function0,scalaz.Applicative)java.lang.Object in trait scalaz.std.OptionFunctions is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scalaz.std.OptionFunctions.getOrElseF")
	at scala.sys.package$.error(package.scala:27)
```

https://travis-ci.org/scalaz/scalaz/jobs/167560515#L583

PR #1272, opened 2016-10-20
Reimplement Free#foldRun without the Functor contraint.

TomasMikula:
This is in line with previous changes to drop the `Functor` constraint from `Free`.

The signature has changed from

```scala
def foldRun[B](b: B)(f: (B, S[Free[S, A]]) => (B, Free[S, A]))(implicit S: Functor[S]): (B, A)
```

to

```scala
def foldRun[B](b: B)(f: λ[α => (B, S[α])] ~> (B, ?)): (B, A)
```

I also added test that `foldRun` can be used to implement a stack-safe state monad. (Previously, `foldRun` was not tested at all.)

**EDIT:** Also added `foldRunM`:

```scala
def foldRunM[M[_], B](b: B)(f: λ[α => (B, S[α])] ~> λ[α => M[(B, α)]])(implicit M0: Applicative[M], M1: BindRec[M]): M[(B, A)]
```

and tests that show that it can be used to implement stack-safe `StateT`.

### DIFF
--- a/core/src/main/scala/scalaz/Free.scala
+++ b/core/src/main/scala/scalaz/Free.scala
@@ -231,15 +231,17 @@ sealed abstract class Free[S[_], A] {
     }
 
   /** Runs to completion, allowing the resumption function to thread an arbitrary state of type `B`. */
-  final def foldRun[B](b: B)(f: (B, S[Free[S, A]]) => (B, Free[S, A]))(implicit S: Functor[S]): (B, A) = {
-    @tailrec def foldRun2(t: Free[S, A], z: B): (B, A) = t.resume match {
-      case -\/(s) =>
-        val (b1, s1) = f(z, s)
-        foldRun2(s1, b1)
-      case \/-(r) => (z, r)
+  @tailrec final def foldRun[B](b: B)(f: λ[α => (B, S[α])] ~> (B, ?)): (B, A) =
+    step match {
+      case Return(a) => (b, a)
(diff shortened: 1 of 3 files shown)


xuwei-k:
```
[error] /home/travis/build/scalaz/scalaz/example/src/main/scala/scalaz/example/TrampolineUsage.scala:36: missing parameter type
[error]     val (steps, sorted1) = quickSort[Function0, Int](xs).foldRun(0)((i, f) => (i + 1, f()))
[error]                                                                      ^
[error] /home/travis/build/scalaz/scalaz/example/src/main/scala/scalaz/example/TrampolineUsage.scala:36: missing parameter type
[error]     val (steps, sorted1) = quickSort[Function0, Int](xs).foldRun(0)((i, f) => (i + 1, f()))
[error]                                                                         ^
[error] two errors found
```

https://travis-ci.org/scalaz/scalaz/jobs/169132253#L541

TomasMikula:
Oops. I only ran `testsJVM/test`. Will fix the examples. Thanks for pointing that out.

TomasMikula:
Fixed.

TomasMikula:
Now there's a spurious failure.

xuwei-k:
> Now there's a spurious failure.

I restarted travis job

TomasMikula:
Thanks!

I just added another commit that adds `foldRunM` and tests that use it to implement stack-safe `StateT`.

PR #1273, opened 2016-10-20
extend stateTMonadPlus type with MonadState

danslapman:
This PR annotates stateTMonadPlus type as `MonadPlus[StateT[F, S, ?]] with MonadState[StateT[F, S, ?], S]`

Motivation:
It is impossible to use both stateTMonadPlus and stateTMonadState due to ambiguity, but it is necessary in some cases.

### DIFF
--- a/core/src/main/scala/scalaz/StateT.scala
+++ b/core/src/main/scala/scalaz/StateT.scala
@@ -162,7 +162,7 @@ sealed abstract class StateTInstances2 extends StateTInstances3 {
 }
 
 sealed abstract class StateTInstances1 extends StateTInstances2 {
-  implicit def stateTMonadPlus[S, F[_]](implicit F0: MonadPlus[F]): MonadPlus[StateT[F, S, ?]] =
+  implicit def stateTMonadPlus[S, F[_]](implicit F0: MonadPlus[F]): MonadPlus[StateT[F, S, ?]] with MonadState[StateT[F, S, ?], S] =
     new StateTMonadStateMonadPlus[S, F] {
       implicit def F: MonadPlus[F] = F0
     }


TomasMikula:
Hi, [here](https://github.com/scalaz/scalaz/blob/d7b370f039e1e47484f023af8e3e03390b567d53/tests/src/test/scala/scalaz/StateTTest.scala#L25-L26) are tests for the absence of ambiguity. What is the case that is not covered?

danslapman:
@TomasMikula consider the following code:
```
implicit val ifsmi = StateT.stateTMonadState[Int, Future]
  import ifsmi.{get, put}

  val algo = for {
    _ <- ifsmi.whileM_(get map (_ < 10),
      for {
        i <- get
        _ <- put(i + 1)
      } yield ()
    )
    v1 <- get
  } yield v1

println(Await.result(algo.run(0), 10.seconds))
```
If i want to use .point here, i need to have an implicit MonadPlus instance in scope both with stateTMonadState, and here comes ambiguity.

TomasMikula:
What about `ifsmi.point`?

danslapman:
@TomasMikula sorry, I misposted, I faced the problem while implemented my `asyncsreams` library, the proper example is [here](https://github.com/danslapman/asyncstreams/blob/master/src/test/scala/asyncstreams/AsyncStreamMonadicOperationsTests.scala#L60-L7) (I had to implement it by-hand without scalaz). I need there StateTMonadState to have get and put and MonadPlus to filter. (My bad talking about `point`)

TomasMikula:
So how would you rewrite that code with this fix. Would you ask for implicit `MonadPlus with MonadState`?

danslapman:
Precisely. With type given in my PR al will work fine.

TomasMikula:
Yeah, it's weird to assume that whenever you can get both `MonadPlus` and `MonadState`, you can get them as `MonadPlus with MonadState`. Plus scalaz-8.0 will have subtyping free encoding of typeclasses, so you will actually never have `MonadPlus with MonadState`.

danslapman:
It's a pity, but ok, got it :(

PR #1274, opened 2016-10-21
MonadState cleanup

TomasMikula:
* Change `state` signature to match the one in mtl, i.e.

   ```scala
   def state[A](f: S => (S, A)): F[A]
   ```

   (The current signature is equivalent to `point`.)
 * Make `init` an alias for `get` (which was the case for all implementations, anyway).
 * Move abstract methods to the top.


vmarquez:
this seems like a no brainer. 👍

TomasMikula:
There's a spurious failure.

TomasMikula:
Thanks to whomever restarted the build. All clean now.

PR #1276, opened 2016-10-22
Fix the `plus` operation for `FreeT`

TomasMikula:
See #1275.


TomasMikula:
Perhaps it was not clear, but I consider this ready to be merged (unlike #1275, which I created solely for the purpose of demonstrating the issue).

The key of the fix is to not stop at the first occurrence of `M` and perform `plus` on it, but keep evaluating (`join`-ing `M`s) until the first `Suspend` is reached.

Let me know if you have questions/concerns.

vmarquez:
👍

PR #1280, opened 2016-10-24
Override reduceUnordered in Task's Nondeterminism

puffnfresh:
Previously we were overriding gatherUnordered which was not actually
called when Task.gatherUnordered was called with its default argument.
Instead chooseAny was being called, which gave very poor performance:

```scala
val start = System.currentTimeMillis
Task.gatherUnordered((1 to 1000).map(Task.now)).unsafePerformSync
println(System.currentTimeMillis - start)
```

For me, the above code before this change gives 563ms.

After this change gives 2ms.

This also makes the underlying Task code consistent with Future. When we
call Task.gatherUnordered, we're now pretty much calling the equivalent
of Future.gatherUnordered.

We should also figure out why Future's chooseAny is very slow, or if
chooseAny is even useful.

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Task.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Task.scala
@@ -297,10 +297,16 @@ object Task {
         new Task ( F.map(F.chooseAny(h.get, t map (_ get))) { case (a, residuals) =>
           a.map((_, residuals.map(new Task(_))))
         })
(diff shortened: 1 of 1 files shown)


jedws:
👍

puffnfresh:
@xuwei-k if this gets merged, it'd be great to get a backport!

tonymorris:
:+1: merge and backport

PR #1302, opened 2016-12-09
Add derivation for MonadState using MonadTrans

dstechenko:
I was using a stack of monads with a `StateT` and found that when I need a `MonadState` I can't get one derived on a `MonadTrans`.
Not sure if this is useful, but thought it might help someone.

### DIFF
--- a/core/src/main/scala/scalaz/MonadState.scala
+++ b/core/src/main/scala/scalaz/MonadState.scala
@@ -27,5 +27,19 @@ object MonadState {
 
   ////
 
+  implicit def promotedMonadState[G[_], F[_], S](
+    implicit
+    mpo: MonadPartialOrder[G, F],
+    ms: MonadState[F, S]
+  ): MonadState[G, S] = new MonadState[G, S] {
+
+    override def get: G[S] = mpo.promote(ms.get)
+    override def put(s: S): G[Unit] = mpo.promote(ms put s)
+
+    override def point[A](a: => A): G[A] = mpo.MG point a
+    override def bind[A, B](ga: G[A])(f: A => G[B]): G[B] = mpo.MG.bind(ga)(f)
+
+  }
+
   ////
 }
--- a/tests/src/test/scala/scalaz/StateTTest.scala
+++ b/tests/src/test/scala/scalaz/StateTTest.scala
@@ -80,4 +80,11 @@ object StateTTest extends SpecLite {
       .foldLeft(StateT((s:Int) => Trampoline.done((s,s))))( (a,b) => a.flatMap(_ => b))
     4000 must_=== result(0).run._1
   }
+
+  "MonadState must be derived for any stack of Monads" in {
+    type StateStack[A] = State[Int, A]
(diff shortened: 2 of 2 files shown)


TomasMikula:
I wonder if this can be made slightly more general with [`MonadPartialOrder`](https://github.com/scalaz/scalaz/blob/d7b370f039e1e47484f023af8e3e03390b567d53/core/src/main/scala/scalaz/MonadTrans.scala#L44), i.e. do this not just for a transformer of `F`, but for any monad "bigger" than `F`:

```scala
implicit def promotedMonadState[G[_], F[_], S](implicit
  gf: MonadPartialOrder[G, F],
  ms: MonadState[F, S]
): MonadState[G, S] = new MonadState[G, S] {
  override def get: G[S] = gf.promote(ms.get)
  override def put(s: S): G[Unit] = gf.promote(ms put s)
}
```

Since for any transformer `MT`, `MT[F, ?]` is "bigger" than `F` via [this implicit instance of `MonadPartialOrder[MT[F, ?], F]`](https://github.com/scalaz/scalaz/blob/d7b370f039e1e47484f023af8e3e03390b567d53/core/src/main/scala/scalaz/MonadTrans.scala#L80), this will give you your `MonadState[MT[F, ?], S]`.

dstechenko:
Interesting point.
I tried your suggestion like this:

```
implicit def promotedMonadState[G[_], F[_], S](
    implicit
    ms: MonadState[F, S],
    mpo: MonadPartialOrder[G, F]
  ): MonadState[G, S] = new MonadState[G, S] {

    override def get: G[S] = mpo.promote(ms.get)
    override def put(s: S): G[Unit] = mpo.promote(ms put s)

    override def point[A](a: => A): G[A] = mpo.MG point a
    override def bind[A, B](ga: G[A])(f: A => G[B]): G[B] = mpo.MG.bind(ga)(f)

  }
```

But I get `error: could not find implicit value for parameter F: scalaz.MonadState[ListS,Int]` for my Stack type.

Seems like resolution diverges on `promotedMonadState`.

TomasMikula:
Do you mind giving the definition of `ListS`?

TomasMikula:
This works for me:

```scala
  type ListS[A] = ListT[State[Int, ?], A]

  implicitly[MonadState[ListS, Int]]
```

but apparently is dependent on the "-Ypartial-unification" compiler flag or the [SI-2712-fix plugin](https://github.com/milessabin/si2712fix-plugin).

dstechenko:
@TomasMikula well, as it turned out I messed up the order of implicits.
In my code snippet `ms: MonadState[F, S]` comes first, so it was diverging for a reason.
Fixed it, also works w/o partial unification for me as well.

TomasMikula:
Maybe add one line to tests that proves that the implicit resolution works?

Otherwise 👍  from me.

dstechenko:
Done, thanks.

TomasMikula:
@dstechenko Thanks, can you squash the commits?

Unless others have comments, I am ready to merge this.

dstechenko:
@TomasMikula done.

PR #1306, opened 2016-12-15
Cast-free type tags

S11001001:
By replacing the structural type with an abstract [existential] type, the `subst`/`unsubst` code sees that a tagged type is equal to the untagged type, so it doesn't need to cast.

* Now, `subst` cannot be used to derive `unsafeCoerce` in concert with singleton type patterns.
* Tags/untags no longer form an optimization barrier for inlining optimizers, because there are neither internal nor external inconsistencies when tagging.

### DIFF
--- a/tests/src/test/scala/scalaz/TagTest.scala
+++ b/tests/src/test/scala/scalaz/TagTest.scala
@@ -7,6 +7,14 @@ import Tags.{Multiplication => Mult}
 import org.scalacheck.Prop.forAll
 
 object TagTest extends SpecLite {
+  "k.@@" should {
+    "be abstract" in {
+      val r = (List[Int](42) ++ List[Int @@ Mult]()).toSet
+      r: Set[Any]
+      true
+    }
+  }
+
   "of.subst" should {
     "substitute" ! forAll {xs: List[Int] =>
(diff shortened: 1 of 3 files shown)


edmundnoble:
@S11001001 You mention that tags with casts form an inlining barrier. Did you come across a concrete case where this version performs better?

tests/src/test/scala/scalaz/TagTest.scala
```diff
@@ -7,6 +7,14 @@ import Tags.{Multiplication => Mult}
 import org.scalacheck.Prop.forAll
 
 object TagTest extends SpecLite {
+  "k.@@" should {
+    "be abstract" in {
+      val r = (List[Int](42) ++ List[Int @@ Mult]()).toSet
+      r: Set[Any]
```
puffnfresh: Nice!

S11001001:
@edmundnoble No, it's just been bugging me since either @sjrd or @DarkDimius (or maybe both? 😕) mentioned casting as a problem at Scala Days NYC.

sjrd:
Not me. So it must have been @DarkDimius.

edmundnoble:
I remember @DarkDimius mentioning this during a dotty linker talk, saying casts can mess up the call graph analysis and end up disabling a lot of the specialization optimizations in linker. Perhaps that's it.

PR #1307, opened 2016-12-16
adds a Semigroup and Monoid for IdT

jedws:
adds a Semigroup for IdT if there is a Plus instance for the underlying F type.

From https://hackage.haskell.org/package/reducers-3.12.1/docs/Data-Semigroup-Alt.html

### DIFF
--- a/core/src/main/scala/scalaz/IdT.scala
+++ b/core/src/main/scala/scalaz/IdT.scala
@@ -60,6 +60,9 @@ sealed abstract class IdTInstances0 extends IdTInstances1 {
 
   implicit def idTOrder[F[_], A](implicit F: Order[F[A]]): Order[IdT[F, A]] =
     F.contramap(_.run)
+
+  implicit def idTMonoid[F[_]: PlusEmpty, A]: Semigroup[IdT[F, A]] =
+    Monoid.monoidInvariantFunctor.xmap(PlusEmpty[F].monoid[A], IdT.apply, _.run)
 }
 
 sealed abstract class IdTInstances extends IdTInstances0 {
--- a/core/src/main/scala/scalaz/IdT.scala
+++ b/core/src/main/scala/scalaz/IdT.scala
@@ -72,6 +72,9 @@ sealed abstract class IdTInstances extends IdTInstances0 {
 
   implicit def idTEqual[F[_], A](implicit F: Equal[F[A]]): Equal[IdT[F, A]] =
     F.contramap(_.run)
+
+  implicit def idTSemigroup[F[_]: Plus, A]: Semigroup[IdT[F, A]] =
+    Semigroup.semigroupInvariantFunctor.xmap(Plus[F].semigroup[A], IdT.apply, _.run)
 }
 
 object IdT extends IdTInstances
--- a/core/src/main/scala/scalaz/IdT.scala
+++ b/core/src/main/scala/scalaz/IdT.scala
@@ -61,8 +61,8 @@ sealed abstract class IdTInstances0 extends IdTInstances1 {
   implicit def idTOrder[F[_], A](implicit F: Order[F[A]]): Order[IdT[F, A]] =
     F.contramap(_.run)
 
-  implicit def idTMonoid[F[_]: PlusEmpty, A]: Semigroup[IdT[F, A]] =
-    Monoid.monoidInvariantFunctor.xmap(PlusEmpty[F].monoid[A], IdT.apply, _.run)
+  implicit def idTSemigroup[F[_]: Plus, A]: Semigroup[IdT[F, A]] =
+    Semigroup.semigroupInvariantFunctor.xmap(Plus[F].semigroup[A], IdT.apply, _.run)
 }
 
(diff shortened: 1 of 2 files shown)


TomasMikula:
If there's a reason why we don't provide _implicit_ semigroup for every `F[_]: Plus` (e.g. because it might collide with other semigroup on `F`), what makes `IdT` different? The linked `newtype Alter` seems to be dedicated to that purpose (turning `Plus` of the underlying `F` into a semigroup), but that's not the case for `IdT`.

puffnfresh:
@TomasMikula this is different because there is just one Semogroup for IdT; no ambiguous or orphan implicits.

I was going to submit a PR for Alter but @tonymorris pointed out that it should just be IdT. I agree. IdT seems like a great place for this.

TomasMikula:
OK, I see how it doesn't cause ambiguous implicits, but

```scala
Semigroup[Foo[Int]]
```

and 

```scala
Semigroup[IdT[Foo, Int]]
```

being (potentially) different semigroups might come out as a surprise. Seems like `IdT` is then serving two unrelated purposes.

jbgi:
@TomasMikula if understand correctly, `Semigroup[Foo[Int]]` being different than `Semigroup[IdT[Foo, Int]]`) would mean `Plus[F].semigroup[Int]` being different than`Semigroup[F[Int]]`, which would be a problem in the first place, I think. (violate coherence?)

TomasMikula:
@jbgi

> would mean `Plus[F].semigroup[Int]` being different than `Semigroup[F[Int]]`

Yes, but `Plus[F].semigroup[Int]` is not implicit.

jbgi:
true. but really, `Plus[Foo]` is just a first class version of, and should behave the same as:
`implicit def fooSemigroup[A]: Semigroup[Foo[A]]`
and it is not allowed to have both the above and
`implicit def intFooSemigroup: Semigroup[Foo[Int]]`

not directly related: maybe if we had proper support for universal quantification (hint, hint), `Plus[F]` may not even exist and we could use `∀[A => Semigroup[Foo[A]]]` instead.

TomasMikula:
@jbgi Makes sense, but then by the same reasoning `Plus[Foo].semigroup[A]` could/should be implicit for all `Foo`s, not just the ones that look like `Foo[A] = IdT[Bar, A]`.

puffnfresh:
@jbgi I'm not sure that `Plus` and `Semigroup` should have consistent behaviour.

@tonymorris @aloiscochard @S11001001 any opinions on using `IdT[F[_], A]` to derive a `Semigroup` from a `Plus`?

S11001001:
:-1:

This should use the underlying `Semigroup` or `Monoid` instead.  `Alt` doesn't really match `IdT`; it's used to explicitly request the alternative (sorry) instance, as @TomasMikula says.

@jbgi Consistency between `Plus` and `Semigroup` is not required for coherence, as @puffnfresh says. In fact, *inconsistency* is desirable in many cases. See also ["the uninteresting monoids of certain monads"](https://failex.blogspot.com/2015/10/the-uninteresting-monoids-of-certain.html).

jbgi:
@S11001001 thanks for the enlightening read. I'd still be curious of a case where the "right" `Plus` would not be consistent with the "right" `Monoid`...

S11001001:
@jbgi `Maybe`, whose `Monoid` lifts the underlying `Semigroup`.

jbgi:
_mais c'est bien sûr !_ thanks @S11001001.

tonymorris:
:+1:

tonymorris:
I just realised by discussion with @S11001001 that there is not `Semigroup[A]` constraint on the `Semigroup` instance thereby making the following property:

    IdT(x) |+| IdT(y) != IdT(x |+| y)

I think it is better to constrain as:

    (Apply[F], Semigroup[A]) => Semigroup[IdT[F, A]]

    (Applicative[F], Semigroup[A]) => Monoid[IdT[F, A]]

or, a suggestion from @S11001001 

    Semigroup[F[A]] => Semigroup[IdT[F, A]]

TomasMikula:
> I think it is better to constrain as:
>
>     (Apply[F], Semigroup[A]) => Semigroup[IdT[F, A]]
>
>     (Applicative[F], Semigroup[A]) => Monoid[IdT[F, A]]

That is a different semigroup/monoid than the one in this PR. This one feels like "multiplicative", whereas the `Plus`/`PlusEmpty`-based feels like "additive" (multiplicative/additive analogy made based on the number of elements for `F = List`). Should `IdT` be a marker for getting the multiplicative semigroup/monoid?

> or, a suggestion from @S11001001 
>
>     Semigroup[F[A]] => Semigroup[IdT[F, A]]

This probably won't hurt, but still doesn't derive a semigroup/monoid as a specialization of `Plus`/`PlusEmpty`, which seems to be the original goal of this PR.

S11001001:
@tonymorris The constraint I suggested is the only one that will preserve that distributive property.

@TomasMikula Right, when I said "`Alt` doesn't really match `IdT`", I was saying that the premise of the PR "derive our implementation from Data.Semigroup.Alt" is faulty.

PR #1311, opened 2016-12-22
Add `Inject.unapply`.

sellout:
This makes it possible to pattern-match on unknown Coproducts, given
appropriate `Inject` constraints.

E.g.
```scala
def foo[F[_], A](fa: F[A])(implicit In: SomeComponent :<: F) = fa match {
  case In(SomeComponent(_)) => true
  case _ => false
}
```


TomasMikula:
👍

Would be nice to add an example to tests.

sellout:
Cool. I’ll do that ASAP.

TomasMikula:
👍

PR #1313, opened 2017-01-15
Make Monoid[M] => Applicative[λ[α => M]] no longer implicit

S11001001:
The instance is too "default instance"-y, and incoherent to boot; a compiler error is better.

This effectively reverts 186693a3869b9c887452911b6bc51d47038a3ee0.

### DIFF
--- a/core/src/main/scala/scalaz/Applicative.scala
+++ b/core/src/main/scala/scalaz/Applicative.scala
@@ -114,7 +114,7 @@ object Applicative {
 
   ////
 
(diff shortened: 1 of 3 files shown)


S11001001:
Discussion from https://twitter.com/S11001001/status/820682842317488129

/cc @purefn

tonymorris:
I am of the opinion that these need to be deleted altogether, but this is at least an improvement.

tonymorris:
Better is, `case class WrappedApplicative[F[_], A]`, then `implicit def MonoidApplicative: Applicative[λ[α => A](implicit M: Monoid[F[A]])`

S11001001:
I'll just delete it.

Issue #1234, opened 2016-08-01
Implement/Port Leibniz equality.


aloiscochard:
@S11001001 how do you feel about scalaz7 impl.? could be ported as is?

aloiscochard:
Actually it's already in... but it's very primitive and need some love when compared with 7 impl.

PR #1314, opened 2017-01-16
[WIP] #1234 - Leibniz & Liskov

alexknvl:
I have ported some of the original Leibniz / Liskov functionality over to scalaz8 branch. 

 * I am still not clear on whether the upper and lower bounds on `Leibniz` are *needed*. I understand that without them `subst[F]` on `F[_ <: X]` won't work, but is it *really* ever used in such a context?
 * I changed some of the names, but I am happy to stick with scalaz7 naming if that's an issue.
 * It is possible to generalize `Leibniz` to higher-kinded types (e.g. equality between type constructors).
 * There are no `lower` functions yet since there is no `Injective` typeclass.
 * A lot of `Liskov` companion object functions are still missing.
 * Should I keep all companion object functions in a `...Functions` trait?
 * `=:=` and `<:<` might have most of the required functionality (most importantly `substitute`) [in 2.13](https://github.com/scala/scala/pull/5623).
 * I have adapted some ideas from a similar PR to cats by @tel. The top comment in `Leibniz` is from the [article](http://portal.acm.org/citation.cfm?id=583852.581494).


TomasMikula:
> I am still not clear on whether the upper and lower bounds on `Leibniz` are *needed*. I understand that without them `subst[F]` on `F[_ <: X]` won't work, but is it *really* ever used in such a context?

I have used it in such context before.

alexknvl:
@TomasMikula The problem is that it adds a lot of noise to an otherwise simple idea, makes type-inference harder, and makes some of the user-facing methods harder to use because you have to supply them with additional type parameters (e.g. `lift` functions) :( Furthermore, you can make a `Groupoid`/`Category` instance only for some specific choice of `L` and `H` (since there is no `GeneralizedGroupoid` right now). And AFAIK scalaz is moving away from subtyping / variance.

Overall it feels like it's a marginal improvement with a huge cost. The same constraints `>: L <: H` can always be supplied using a typeclass during construction:
```scala
class X[A <: AnyVal](a: A)
// can be replaced by:
class X[A](a: A, ev: A <~< AnyVal)
// or if the evidence is not required after construction:
class X[A](a: A)
object X {
  def apply[A](a: A)(implicit ev: A <~< AnyVal): X[A] = ???
}
```

TomasMikula:
> The same constraints `>: L <: H` can always be supplied using a typeclass

The problem is when the type constructor with type bounds is not under your control, so you can't change it. Do we say "f*** you" to all such use cases?

**Aside:** I think Scala's bounds checking is unnecessarily strict. If you have

```scala
class Foo[A <: Bar]
```

I believe you should be allowed to talk about `Foo[Int]` (it just might be uninhabited). This would make things simpler.

alexknvl:
What do you think of the solutions I outlined at the end of my previous post? I am mostly concerned with type-inference on methods and combinators and I'd like to optimize for the most common case of `Leibniz[Nothing, Any, A, B]`.

**P.S.**

> Do we say "f*** you" to all such use cases?

Well, it is not like every single data type taking kind `* -> *` type parameters in Scalaz provides you a way to bound them with `>: L <: H`. Such constraints are unlikely to come up when you live in the Scalazzi world. Scalaz says "f*** you" to many other things like `null`, non-parametric `equals`, `hashCode`, exceptions, unconstrained side-effects, some of scala collections, (ought to say that to `with` intersection types) etc.

> I believe you should be allowed to talk about Foo[Int] (it just might be uninhabited).

That would raise some interesting questions like:
```scala
trait Bar { type Out = String }
trait Baz { type Out = Boolean }
class Foo [A <: Bar] { type Out = A#Out }
type X = Foo[Baz]#Out // what is X?
```

TomasMikula:
> What do you think of the solutions I outlined at the end of my previous post?

I like the one with wrapping.

Not sure how "all methods on syntax traits" helps if `===` is just a type alias (unless the methods are named differently anyway).


> That would raise some interesting questions like:
> ```scala
> trait Bar { type Out = String }
> trait Baz { type Out = Boolean }
> class Foo [A <: Bar] { type Out = A#Out }
> type X = Foo[Baz]#Out // what is X?
> ```

Good point. For that the constraint would have to be checked.

alexknvl:
[This](https://github.com/alexknvl/leibniz) is roughly what I had in mind. Turns out there is quite a bit of duplication needed to completely avoid allocations (calling methods on `BoundedLeibniz` would require unwrapping and then wrapping `Leibniz` again). I also implemented `LeibnizK` there (useful for type-aligned sequences of natural transformations).

TomasMikula:
If `Leibniz` extended `AnyVal`, wrapping would be without allocation, right?

S11001001:
@TomasMikula [Not quite.](http://scastie.org/10882)

alexknvl:
Sorry for neglecting this PR, I've been experimenting with `Leibniz`, `Liskov`, and a number of other related ideas in [a separate project](https://github.com/alexknvl/leibniz/tree/master/src/main/scala/leibniz). I understand why you want bounded version of `Leibniz` now.

alexknvl:
Some of the new "inventions" and old goodies:
 * `Is` - unbounded `Leibniz`.
 * `As` - unbounded `Liskov`.
 * `Leibniz` - bounded `Leibniz`.
 * `Liskov` - bounded `Liskov`.
 * `As1` - unbounded `Liskov` that uses existential types for its encoding, which allows you to call methods like `def f[X <: B](x: X)` given `a: A` and `As1[A, B]`.
 * `Liskov1` - ditto, but bounded.
 * `LeibnizK` - higher-order `Leibniz`.
 * `Eq` - experiments showed that you can make `===` even more powerful if you have an ability to lift term equality `a === b` to singleton type equality `a.type === b.type`. But you need an equality that satisfies `forall (a: A) (b: B) (f: A -> B) . f(a) === f(b) <=> a === b` assuming `f` follows Scalazzi rules.

TomasMikula:
If you want to replace this PR with your new `Is`, `As`, `Leibniz`, `Liskov`, `IsK`, I would give it a :+1:.

> experiments showed that you can make `===` even more powerful if you have an ability to lift term equality `a === b` to singleton type equality `a.type === b.type` (e.g. [this gist](https://gist.github.com/alexknvl/78b91119bb83ee8a2af75789e3bb240a), [this one](https://gist.github.com/alexknvl/5db6ef91bfd2753575275a4a37646dc3), and [this one](https://gist.github.com/alexknvl/860795487b1b60309fba608503c94c62)

In the last two gists there doesn't seem to be any guarantee that the types you are asserting to be equal are singleton types.

> the last gist also shows how to "prove" stuff in Scala, Idris-style

This is nice. Could it be used to prove something about types that are not fully known at compile time? E.g. how would you prove that `M + N = N + M` when all you know at compile time is `M <: Nat`, `N <: Nat`?

> But you need an equality that satisfies `forall (a: A) (b: A) (f: A -> Bool) . f(a) == f(b) <===> a === b`

I suppose you meant `forall (a: A) (b: A) . (forall (f: A -> Bool) . f(a) == f(b)) <===> a === b`.

alexknvl:
> If you want to replace this PR with your new `Is`, `As`, `Leibniz`, `Liskov`, `IsK`, I would give it a :+1:.

Okay, awesome. I'll get it done this week.

> In the last two gists there doesn't seem to be any guarantee that the types you are asserting to be equal are singleton types.

Right, those are early prototypes before I got into singleton types :wink:

> This is nice. Could it be used to prove something about types that are not fully known at compile time? E.g. how would you prove that `M + N = N + M` when all you know at compile time is `M <: Nat, N <: Nat`?

Well, ideally you would have the compiler (or a compiler plugin) realize that since the proof is total and equality is a singleton, you can just "force" it for arbitrary `M <: Nat` and `N <: Nat`. I guess you could have some kind of 
```scala
def assertTotal[P[_ <: Nat, _<: Nat]: Proof]
(proof: [M <: Nat, N <: Nat] (M, N) => P[N, M]): P[N, M]
```

> I suppose you meant `forall (a: A) (b: A) . (forall (f: A -> Bool) . f(a) == f(b)) <===> a === b)`.

Yep.

base/src/main/scala/scalaz/data/Leibniz.scala
```diff
@@ -1,16 +1,216 @@
...
+    type f3[α >: L3 <: H3] = Leibniz[LF, HF, F[A1, A2, A3], F[B1, B2, α]]
+    eq3.subst[f3](eq2.subst[f2](eq1.subst[f1](refl_[LF, HF, F[A1, A2, A3]])))
+  }
+
+  /**
+    * Unsafe coercion between types. `unsafeForce` abuses `asInstanceOf` to
+    * explicitly coerce types. It is unsafe, but needed where Leibnizian
+    * equality isn't sufficient.
+    */
+  def unsafeForce[A, B]: A === B =
```
TomasMikula: Why does this `unsafeForce` not need `@SuppressWarnings(Array("org.wartremover.warts.AsInstanceOf"))`, while all of the others do?

base/src/main/scala/scalaz/data/Liskov.scala
```diff
@@ -0,0 +1,161 @@
...
+  //  /**
+  //    * Given `A <~< B` we can prove that `F[B] <~< F[B]` for any
+  //    * contravariant `F[-_]`.
+  //    *
+  //    * @see [[liftCo]]
+  //    */
+  //  final def liftCt[F[-_]]: F[B] <~< F[A] = {
+  //    type f[+α] = F[α] <~< F[A]
+  //    substCo[f](refl)
+  //  }
```
TomasMikula: 👎 for commented-out code.

base/src/main/scala/scalaz/data/Leibniz.scala
alexknvl: Scalaz doesn't use WartRemover as far as I can tell, I will simply remove this annotation. All `asInstanceOf` require it, I was just being lazy with warnings in `leibniz` :wink:

alexknvl:
It looks like CI checks failed because it couldn't download SBT.

TomasMikula:
Sorry for waiting so long, I just wanted to give others the chance to review. If you find time to resolve the conflicts, I will merge this. If not, I will try to resolve them myself. Thanks!

alexknvl:
https://github.com/scalaz/scalaz/pull/1524

PR #1316, opened 2017-01-28
Don't pass around fictitious state (IvoryTower) in ST.

TomasMikula:
The [linked paper](http://research.microsoft.com/en-us/um/people/simonpj/papers/lazy-functional-state-threads.ps.Z) states (section 6.2.2) two reasons for passing the state around, none of which seems to apply in Scala:

> Firstly, the compiler “shakes the code around” quite considerably: is it possible that it might somehow end up changing the order in which the primitive operations are performed? No, it is not. The input state of each primitive operation is produced by the preceding operation, so the ordering between them is maintained by simple data dependencies of the explicit state, which are certainly preserved by every correct program transformation.

Since Scala is not pure, the compiler is already restricted in reordering the code, particularly wrt. reads and writes to mutable variables.

> Secondly, the explicit state allows us to express to the compiler the strictness of the primitive operations in the state.

Scala is already strict.

Similarly, state-passing could also be eliminated from `IO`.

### DIFF
--- a/effect/src/main/scala/scalaz/effect/IO.scala
+++ b/effect/src/main/scala/scalaz/effect/IO.scala
@@ -297,7 +297,7 @@ object IO extends IOInstances {
 
   /** An IO action is an ST action. */
   implicit def IOToST[A](io: IO[A]): ST[IvoryTower, A] =
-    st(io(_).run)
+    st(() => io(ivoryTower).run._2)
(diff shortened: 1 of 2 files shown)


edmundnoble:
Agreed, please remove this. No reason to remove the phantom type param.

PR #1317, opened 2017-01-30
[WIP] Fix for Task.onFinish's broken interaction with timed/runFor (DO NOT MERGE YET)

vmarquez:
Check the current version of scalaz:

```scala
Task { Thread.sleep(1000) }.onFinish(t => Task.delay { println("HELLO WORLD") } ).timed(1).unsafePerformSyncAttempt
```

Seems like it would be useful for this to print but it doesn't because the algebra isn't comprehensive enough to account for the interrupt cases in listenInterruptibly. 

For the same reason, though by a different mechanism, 

```scala
Task { Thread.sleep(1000) }.onFinish(t => Task.delay { println("HELLO WORLD") } ).unsafePerformSyncAttemptFor(1)
```

Also does not print in the current version.  Both of these issues are fixed in the PR.  Because this is a nontrivial change to the internals of Task, I will be running more tests from other projects that make heavy use of Task so for the time being I am just looking for reviews, not for a merge.  

Ultimately, I believe we should move to a new IO that @puffnfresh  has been working on and just typedef Task to it and/or provide some help for as much backwards compatibility as we can provide.  The main goal for this was to get the onFinish semantics fixed in the 7.3 release and then backport to 7.2 and maybe even 7.1 while we continue to work on a replacement for 7.4+


concurrent/src/main/scala/scalaz/concurrent/Task.scala
```diff
@@ -52,13 +53,23 @@ class Task[+A](val get: Future[Throwable \/ A]) {
...
-    new Task(get flatMap {
-      case -\/(e) =>
-        Task.Try(f(Some(e))).fold(e2 => Future.now(-\/(e2)), _.get *> Future.now(-\/(e)))
-      case r =>
-        Task.Try(f(None)).fold(e => Future.now(-\/(e)), _.get *> Future.now(r))
-    })
+   def onFinish(f: Option[Throwable] => Task[Unit]): Task[A] = {
+    import scalaz.std.function._
+    import scalaz.concurrent.Future._
+    val ttf = f.map(_.get.map(_ => ()))
```
djspiewak: Swallows exceptions thrown in finalizers.

concurrent/src/main/scala/scalaz/concurrent/Task.scala
```diff
@@ -168,11 +188,80 @@ class Task[+A](val get: Future[Throwable \/ A]) {
     unsafePerformSyncFor(timeout.toMillis)
 
   /**
+   * Run this computation to obtain an `A`, so long as `cancel` remains false.
+   * Because of trampolining, we get frequent opportunities to cancel
+   * while stepping through the trampoline, so this should provide a fairly
+   * robust means of cancellation.
+   */
+  def unsafePerformListenInterruptibly(cb: Throwable \/ A => Trampoline[Unit], cancel: AtomicBoolean): Unit = {
+    (this.get.stepInterruptiblyOnFinish(cancel, None): @unchecked) match {
```
djspiewak: I don't think you actually need `@unchecked` here.

concurrent/src/main/scala/scalaz/concurrent/Future.scala
```diff
@@ -235,7 +264,7 @@ sealed abstract class Future[+A] {
 object Future {
   case class Now[+A](a: A) extends Future[A]
   case class Async[+A](onFinish: (A => Trampoline[Unit]) => Unit) extends Future[A]
-  case class Suspend[+A](thunk: () => Future[A]) extends Future[A]
+  case class Suspend[+A](thunk: () => Future[A], andFinally: Option[Option[Throwable] => Future[Unit]]) extends Future[A]
```
djspiewak: This algebra makes it impossible to entirely prevent resource leaks.  A more appropriate signature would be `Suspend[+A](thunk: () => Future[A], andFinally: Option[Throwable \/ A => Future[Unit]])`.  As a use case, imagine a resource is acquired asynchronously after preemption; that resource needs to be freed.

concurrent/src/main/scala/scalaz/concurrent/Task.scala
```diff
@@ -181,10 +270,28 @@ class Task[+A](val get: Future[Throwable \/ A]) {
...
+          }
+        }
+      }
+      , timeoutInMillis, TimeUnit.MILLISECONDS)
+      unsafePerformListenInterruptibly(a => Trampoline.done({
+        if(done.compareAndSet(false, true)) 
+          cb(a)
+        else
+          cb(-\/(new TimeoutException(s"Timed out after $timeoutInMillis milliseconds")))
+        }), cancel)
```
djspiewak: This fixes the interactions between consecutive `timed` calls, though it still breaks monadic associativity.

vmarquez:
@xuwei-k spurious error in the JS test?  Runs OK locally.

concurrent/src/main/scala/scalaz/concurrent/Future.scala
vmarquez: @djspiewak can you give a code example?  You can always close over the A, no?

concurrent/src/main/scala/scalaz/concurrent/Task.scala
vmarquez: oh good catch.

concurrent/src/main/scala/scalaz/concurrent/Task.scala
vmarquez: yeah, It's already broken though, so this makes it less broken.

concurrent/src/main/scala/scalaz/concurrent/Future.scala
djspiewak: I'm pretty sure the following will not free the resource:

```scala
val rsrc: Task[R] = ???
def free(r: R): Task[Unit] = ???

rsrc flatMap { r => now(r) onFinish { _ => free(r) } }
```

If `rsrc` is interrupted, I'm pretty sure the `onFinish` won't run at all.

vmarquez:
@xuwei-k this works on my local for the JS test, seems a spurious error.  Can you restart the build (or is there a way for me to?)

concurrent/src/main/scala/scalaz/concurrent/Task.scala
vmarquez: ah, yeah, thanks.

concurrent/src/main/scala/scalaz/concurrent/Task.scala
vmarquez: ```Task { 4 }.onFinish(i => Task.delay { sys.error("blah") } )```

In the current version it has the same behavior, so while we can change it going forward, I'm not sure we should...

emilypi:
Outdated as of https://github.com/scalaz/scalaz/pull/2037

Issue #1020, opened 2015-10-14
*> is strict in second argument


Profpatsch:
```
none[Int] *> (sys.error("bang")) 
> java.lang.RuntimeException: bang
```

I’d argue it should be non-strict in the second argument, in order to make short-circuiting possible (if that is possible?)

Not sure about `<*`, whether it should be non-strict in second argument or first. Probably second.

tonymorris:
This came up on IRC. i think it goes back to whether `Apply` should be non-strict, and then the implications for any change on that, etc.

kevinmeredith:
Per http://stackoverflow.com/questions/36335417/make-argument-lazy-without-changing-function-signature#comment60293037_36335417, is there a sensible way to address this item?

Note - I'm **assuming** that using a `by-name` parameter is not acceptable here.

jbgi:
By using a church encoding (so not a case class) it is possible to create both strict and lazy constructors and thus not rely on by-name parameters. I took this approach to implement algebraic data types and laziness for java: https://github.com/derive4j/derive4j#first-class-laziness (see eg. https://gist.github.com/jbgi/43c1bd0ab67e3f4b9634#file-lists-java-L99)

tonymorris:
I have never found a good solution to this problem [in Scala], except to have every variant of strict or non-strict for every possible function argument. Of course, this is not practical, so just implement "whatever is likely to be useful for now" and if another variant comes up as required, then implement that too (etc etc).

kevinmeredith:
For my own understanding, why can't *> simply have a by name second parameter?

Further, what harm, if any, is done by replacing strictly evaluated arguments with lazy ones?

tonymorris:
It's a trade-off. I agree that there are benefits to have by-name
argument, but that incurs other penalties, such as not working well with
scala e.g. the first argument in object.method(args) is always strict,
and so why do we use those? Because type-inference works better, etc etc.

I have found no better solution to this problem (in scala) than to
simply have every permutation of strict/non-strict function arguments,
while omitting those that don't come up very often just due to the
workload. So a *> function that is non-strict in second argument is
definitely worthwhile, just it would have to be a different function,
and nobody has yet written it.

S11001001:
@Profpatsch By #1318, it's likely that `*>ByName` will be introduced with the nonstrict-right behavior for 7.3.0.

Profpatsch:
Nice. I’m just glad I don’t have to write Scala anymore.

[...]

PR #1318, opened 2017-02-11
Make *> non-strict in its second argument.

kevinmeredith:
Addresses #1020.

## Testing

I also added a few unit tests within this PR.

```
Welcome to Scala version 2.10.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).

import scalaz.syntax._
import scalaz.Scalaz._

scala> none[Int] *> ???
res1: Option[Nothing] = None
```

### DIFF
--- a/core/src/main/scala/scalaz/syntax/ApplySyntax.scala
+++ b/core/src/main/scala/scalaz/syntax/ApplySyntax.scala
@@ -11,8 +11,14 @@ final class ApplyOps[F[_],A] private[syntax](val self: F[A])(implicit val F: App
   /** Combine `self` and `fb` according to `Apply[F]` and discard the `A`(s) */
   final def *>[B](fb: F[B]): F[B] = F.discardLeft(self,fb)
 
+  /** Combine `self` and `fb` according to `Apply[F]` and discard the `A`(s), except 'fb' is non-strictly evaluated. */
+  final def `*>ByName`[B](fb: => F[B]): F[B] = F.discardLeft(self,fb)
+
   /** Combine `self` and `fb` according to `Apply[F]` and discard the `B`(s) */
   final def <*[B](fb: F[B]): F[A] = F.discardRight(self,fb)
+
+  /** Combine `self` and `fb` according to `Apply[F]` and discard the `B`(s), except 'fb' is non-strictly evaluated. */
+  final def `<*ByName`[B](fb: => F[B]): F[A] = F.discardRight(self,fb)
 
   /**
    * DSL for constructing Applicative expressions.
--- a/tests/src/test/scala/scalaz/ApplyTest.scala
+++ b/tests/src/test/scala/scalaz/ApplyTest.scala
@@ -43,4 +43,41 @@ object ApplyTest extends SpecLite {
   "<*>" in {
     some(9) <*> some({(_: Int) + 3}) must_===(some(12))
   }
+
+  import Scalaz.none
+
+  def err: Option[String] = sys.error("should be non-strict!")
+
+  "*>" in {
+    some(1)       *> some(2)   must_=== (some(2))
+    some(1)       *> none[Int] must_=== (none[Int])
+    none[Int]     *> none[Int] must_=== (none[Int])
+    none[Int]     *> some(2)   must_=== (none[Int])
+    (none[String] *> err).mustThrowA[RuntimeException]
+  }
+
+  "`*>ByName`" in {
+    some(1)       `*>ByName` some(2)   must_=== (some(2))
+    some(1)       `*>ByName` none[Int] must_=== (none[Int])
+    none[Int]     `*>ByName` none[Int] must_=== (none[Int])
+    none[Int]     `*>ByName` some(2)   must_=== (none[Int])
+    none[String]  `*>ByName` err       must_=== (none[String])
+  }
+
+  "<*" in {
+    some(1)       <* some(2)   must_=== (some(1))
+    some(1)       <* none[Int] must_=== (none[Int])
+    none[Int]     <* none[Int] must_=== (none[Int])
+    none[Int]     <* some(2)   must_=== (none[Int])
+    (none[String] <* err).mustThrowA[RuntimeException]
+  }
+
+  "<*ByName" in {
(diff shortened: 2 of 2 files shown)


kevinmeredith:
I don't know against which branch this PR should apply.

So please correct me if wrong.

vmarquez:
Thanks for the PR.  Couple things.  Is this always desirable?  Any previous discussion on why it was done this way?  And finally, this may have some performance impacts... 

I suspect some others may have strong feelings on this @puffnfresh  etc.

tonymorris:
It's not always desirable. You need (2^n) versions of every function for which there are n arguments for every combination of strict and non-strict argument. This is of course, silly. So just write the subset of those functions which are useful.

It is not a good idea in Scala to argue over whether a function argument should, or should not, be strict. Make both versions. The function being proposed here is useful, so it should be written in addition to the existing useful function that already exists.

Scala does not do non-strictness well. It is best to concede to that fact, and accept this outcome imo.

kevinmeredith:
Thanks for the review, Vincent and Tony.

I'm not sure how to implement two versions of `*>`, i.e. where the second argument is call by name and by value. I attempted to do so via this example:

```
scala> :paste
// Entering paste mode (ctrl-D to finish)

def f(x: => Int) = 42
def f(x: Int) = 66

// Exiting paste mode, now interpreting.

f: (x: => Int)Int <and> (x: Int)Int
f: (x: => Int)Int <and> (x: Int)Int

scala> f( {println("hi!"); 1 } )
hi!
res3: Int = 66
```

I'm guessing that overloading is not possible in this case?

Thanks

tonymorris:
Just name it something else.

kevinmeredith:
OK - I added `*>'`, reverting any changes to `*>`.

I'm not familiar with this library's naming conventions, so please let me know if it's OK.

puffnfresh:
I don't think it's practical for the identifier to have a `'` in it.

tonymorris:
I honestly don't know a solution to the explosion of a naming problem. All I know is that debating the strictness of a function argument, for scala, is not a good idea. Both sides of the debate win; produce two functions [then repeat].

tpolecat:
N.B. I recently learned that `fooʹ` is and always has been a valid identifier, you just have to know how to type the correct `ʹ` (U+02B9). However it's considered alpha so `*>ʹ` isn't valid.

kevinmeredith:
Please offer an alternative, @puffnfresh, and I will change it.

tonymorris:
You could come up with a naming scheme where `ʹ` denotes the strictness of arguments.

Clutching at straws I know.

kevinmeredith:
> You could come up with a naming scheme where `ʹ` denotes the strictness of arguments.

Is there any existing naming schema that differentiates strict from non-strict functions?

tonymorris:
Nothing I know of. Be the first!

kevinmeredith:
I'm going to check this [thread](https://groups.google.com/forum/#!topic/scala-user/SkjN0l9mXCk) over the next day or less. 

Then, I'll update this PR with my proposed re-name!

TomasMikula:
When naming higher-kinded versions of a thing, names like `Functor`, `FunctorK`, `FunctorKA` `HFunctorK`, etc. got confusing very quickly. I'm now using names like

```scala
`Functor{* -> *}`
`Functor{(* -> *) -> *}`
`Functor{(* -> *) -> * -> *}`
```
Slightly longer, but this scheme scales to any kinds.

Similarly, you might want to have a scheme that scales to any number of arguments.

tonymorris:
@TomasMikula somewhere in the scalaz history you'll find a similar naming scheme but for a different purpose: applying type constructors of a different kind. It also went through a stage of naming like `Apply1Of2`, but that doesn't work when the type constructor accepts type constructor arguments.

@kevinmeredith You are very brave asking [scala-user] for advice on naming schemes. Good luck!

kevinmeredith:
Rex Kerr shared his [experiences](https://groups.google.com/d/msg/scala-user/SkjN0l9mXCk/tOqk-M8XCwAJ) with such a naming convention.

But I'm not sure if you care for the `*>ByName` (surrounded by ticks), per Rex's comments.

tonymorris:
Told ya asking [scala-user] is a brave move :)

You still have the problem of multiple arguments and each combination of strictness.

kevinmeredith:
> You still have the problem of multiple arguments and each combination of strictness.

I see. However, would you ever want to call `*>`'s first argument in a non-strict fashion?

tonymorris:
Yep. You asking me to come up with an example?

kevinmeredith:
Please! I am asking for my own understanding.

kevinmeredith:
@S11001001 for the `<*` variant, you're expecting that the left argument is evaluated non-strictly?

Example:

```scala
??? `<*ByName` none[Int] 
```

would result in `None` rather than throw since the right argument short-circuited?

If so, would that require changing [self](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/syntax/ApplySyntax.scala#L5) to be evaluated in a call by-name fashion?

```
final class ApplyOps[F[_],A] private[syntax](val self: => F[A]) ...
```

S11001001:
@kevinmeredith No, left-strict, right-non for both `*>` and `<*`.

kevinmeredith:
Understood, @S11001001. I believe that my last commit addresses your requested changes.

TomasMikula:
A bit of bikeshedding regarding `ʹ` (U+02B9): There is a similar-looking letter in the same unicode block, `ˊ` (U+02CA):

| Code | Glyph | `monospace` glyph | Description |
| --- | --- | --- | --- |
| U+02B9 | ʹ | `ʹ` | Modifier Letter Prime |
| U+02CA | ˊ | `ˊ` | Modifier Letter Acute Accent |

The word "acute" might better convey the meaning (i.e. strictness) than the word "prime".

The monospaced glyph for Acute Accent doesn't look very good here on GitHub, but in both my IntelliJ and vim (with default fonts) it is better discernible from `'` than the Prime symbol.

PR #1322, opened 2017-02-26
support scala-native

xuwei-k:
/cc @densh @MasseGuillaume


project/plugins.sbt
```diff
@@ -1,5 +1,11 @@
 scalacOptions += "-deprecation"
 
+addSbtPlugin("org.scala-native" % "sbt-crossproject" % "0.1.0")
```
MasseGuillaume: sbt-scala-native depends on sbt-crossproject you can remove this line if you want to.

MasseGuillaume:
@xuwei-k really good. You can run the sbt task [nativeMissingDependencies](https://github.com/scala-native/scala-native/blob/master/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L40) to see if anything is missing at link time.

project/plugins.sbt
xuwei-k: Thanks. removed 😃

xuwei-k:
I will remove `java.sql` instances https://github.com/scalaz/scalaz/pull/1324

PR #1323, opened 2017-03-02
Adding Nondeterminism typeclass for Kleisli

tbrown1979:
Today I found that this type class would have been pretty helpful. This gist represents my issue a bit: https://gist.github.com/tbrown1979/e9f45f3eadff2e2ff05aa5d8e3b6418d

I had some `Task`s that are inside of `Kleisli`s and I wanted to have two run in parallel - but they aren't in a list. Normally I could do `Nondeterminism[Task].mapBoth(t1, t2) { (a, b) => ... }` but since we were working with `Kleisli[Task]` it wasn't that easy. I looked around at `ParallelTask` but found that `Tag`ging my `Task`s was a little clunky. With this type class I can now use the `nmap`s from this type class to have things run in parallel, and I don't have to tag my Tasks.

Credit to Rob Norris for suggesting this solution.

### DIFF
--- a/core/src/main/scala/scalaz/Kleisli.scala
+++ b/core/src/main/scala/scalaz/Kleisli.scala
@@ -244,6 +244,22 @@ abstract class KleisliInstances extends KleisliInstances0 {
     new KleisliCatchable[F, A] {
       implicit def F = F0
     }
+
+  implicit def kleisliNondeterminism[F[_]: Nondeterminism, E] = new Nondeterminism[Kleisli[F, E, ?]] {
+    val F = Nondeterminism[F]
+
+    def point[A](a: => A): Kleisli[F,E,A] = Kleisli( (x: E) => F.point(a) )
+    def bind[A, B](fa: Kleisli[F,E,A])(f: A => Kleisli[F,E,B]): Kleisli[F,E,B] = fa.flatMap(f)
+    def chooseAny[A](head: Kleisli[F,E,A],tail: Seq[Kleisli[F,E,A]]): Kleisli[F,E,(A, Seq[Kleisli[F,E,A]])] = {
+      Kleisli { (x: E) =>
+        val fa = head.run(x)
+        val sfa = tail.map(_.run(x))
+        F.map(F.chooseAny(fa, sfa)) { case (a, residuals) =>
+          (a, residuals.map(r => Kleisli { (x: E) => r } ) )
+        }
+      }
+    }
+  }
 }
 
 object Kleisli extends KleisliInstances {


core/src/main/scala/scalaz/Kleisli.scala
```diff
@@ -244,6 +244,22 @@ abstract class KleisliInstances extends KleisliInstances0 {
...
+    val F = Nondeterminism[F]
+
+    def point[A](a: => A): Kleisli[F,E,A] = Kleisli( (x: E) => F.point(a) )
+    def bind[A, B](fa: Kleisli[F,E,A])(f: A => Kleisli[F,E,B]): Kleisli[F,E,B] = fa.flatMap(f)
+    def chooseAny[A](head: Kleisli[F,E,A],tail: Seq[Kleisli[F,E,A]]): Kleisli[F,E,(A, Seq[Kleisli[F,E,A]])] = {
+      Kleisli { (x: E) =>
+        val fa = head.run(x)
+        val sfa = tail.map(_.run(x))
+        F.map(F.chooseAny(fa, sfa)) { case (a, residuals) =>
+          (a, residuals.map(r => Kleisli { (x: E) => r } ) )
```
tpolecat: The residuals throw away their environment which doesn't look right. Maybe this is why the instance didn't exist before. It may not make sense.

core/src/main/scala/scalaz/Kleisli.scala
```diff
@@ -244,6 +244,22 @@ abstract class KleisliInstances extends KleisliInstances0 {
     new KleisliCatchable[F, A] {
       implicit def F = F0
     }
+
+  implicit def kleisliNondeterminism[F[_]: Nondeterminism, E] = new Nondeterminism[Kleisli[F, E, ?]] {
```
xuwei-k: - please specify explicit return type
- use [trait KleisliMonad](https://github.com/scalaz/scalaz/blob/c05cf1c4c5cc17463702b74a2de837d01d98dcc0/core/src/main/scala/scalaz/Kleisli.scala#L323) for share implementations of `bind` and `point`

S11001001: Also, tests for ambiguity in the presence of `F[_]: Nondeterminism` will be needed; see many neighbors to these tests in `KleisliTest.scala`.

core/src/main/scala/scalaz/Kleisli.scala
S11001001: I agree, this `chooseAny` does not seem to have the behavior of a nondeterministic `uncons`.  `flatMap` on the result would yield surprising results, as the kleislis don't behave like kleislis anymore.

tpolecat:
Yeah @tbrown1979 sorry for leading you astray. You can implement "parallel sequence" (below) which is the main point of `Nondeterminism` but it looks like you can't implement the underlying primitive.

```scala
scala> val ks = List[Kleisli[Task,Int,String]]()
ks: List[scalaz.Kleisli[scalaz.concurrent.Task,Int,String]] = List()

scala> ks.sequenceU
res4: scalaz.Kleisli[scalaz.concurrent.Task,Int,List[String]] = Kleisli(<function1>)

scala> Kleisli(ks.traverseU(_.run).map(Nondeterminism[Task].gather))
res5: scalaz.Kleisli[scalaz.concurrent.Task,Int,List[String]] = Kleisli(<function1>)
```

tbrown1979:
Thanks for taking the time to review my pr, everyone. This suggestion stemmed from a post I made on gitter, which pertains to this gist: https://gist.github.com/tbrown1979/e9f45f3eadff2e2ff05aa5d8e3b6418d

I can close this pr as it seems this won't work, but I'm wondering if anyone could help me with the question posed in my gist?

PR #1324, opened 2017-03-03
use AutoCloseable. remove scalaz.std.effect.sql

xuwei-k:


xuwei-k:
> Note that Closeable doesn't extend AutoCloseable for Java6 compat

😢  

https://github.com/scala-js/scala-js/blob/v0.6.14/javalib/src/main/scala/java/io/Closeable.scala#L3

PR #1327, opened 2017-03-14
Enable test for Bifoldable laws

justjoheinz:
Though the laws were defined for Bifoldable, I could not find a test file to use them.

### DIFF
--- a/tests/src/test/scala/scalaz/BifoldableTest.scala
+++ b/tests/src/test/scala/scalaz/BifoldableTest.scala
@@ -1,0 +1,9 @@
+package scalaz
+
+import scalaz.scalacheck.ScalazProperties._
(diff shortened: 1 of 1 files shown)


xuwei-k:
`bitraverse.laws[Either]` include `bifoldable.laws[Either]`

- https://github.com/scalaz/scalaz/blob/f430a1ced1844623906b7f2b99d59a1050698e9a/scalacheck-binding/src/main/scala/scalaz/scalacheck/ScalazProperties.scala#L342
- https://github.com/scalaz/scalaz/blob/f430a1ced1844623906b7f2b99d59a1050698e9a/tests/src/test/scala/scalaz/std/EitherTest.scala#L13

PR #1340, opened 2017-03-31
Generate ScalaDoc for common type aliases

Atry:


Atry:
@xuwei-k , could you rebuild it on CI?

core/src/main/scala/scalaz/package.scala
```diff
@@ -140,9 +145,17 @@ package object scalaz {
...
 
+  /** @template */
   type StateT[F[_], S, A] = IndexedStateT[F, S, S, A]
+
+  /** @template */
   type IndexedState[-S1, S2, A] = IndexedStateT[Id, S1, S2, A]
-  /** A state transition, representing a function `S => (S, A)`. */
+  
+  /**
+    * A state transition, representing a function `S => (S, A)`.
```
S11001001: "A state..." should be on the same line as `/**`.

Atry:
@S11001001 @xuwei-k @TomasMikula  Any process?

PR #1342, opened 2017-03-31
Make Task an AnyVal

Atry:


concurrent/src/main/scala/scalaz/concurrent/Timer.scala
```diff
@@ -103,7 +103,7 @@ case class Timer(timeoutTickMs: Int = 100, workerName: String = "TimeoutContextW
     futureNondeterminism.choose(timeoutFuture, future).map(_.fold(_._1.left, _._2.right))
   }
 
-  def withTimeout[T](task: Task[T], timeout: Long): Task[Timeout \/ T] = {
+  def withTimeout[T](task: Task[T], timeout: Long, workaroundForDoubleDefinition: Unit = ()): Task[Timeout \/ T] = {
```
xuwei-k: use DummyImplicit ?

- https://github.com/scala/scala/blob/v2.12.1/src/library/scala/Predef.scala#L523-L534
- http://stackoverflow.com/a/34746255

Atry: What's the pros and cons of `DummyImplicit` in comparison with default parameter?

S11001001:
Given that these `AnyVal` things wantonly box/unbox, but I don't use `Task`, was wondering: do people, like, put these `Task` things in lists, or tuples, or, like, pass them to/return them from `Function1`s?

edmundnoble:
@S11001001 People do put them in Lists and Tuples to defer choosing to execute them concurrently or serially.

S11001001:
@edmundnoble In that case, perhaps this should not be made an `AnyVal`.

Atry:
I thought a monad transformer should be an `AnyVal` by default. Am I wrong?

edmundnoble:
For the record though putting them in Lists or Tuples is not always done between `flatMap` boundaries, which I think means this is still an opportunity to reduce allocations.

S11001001:
@Atry We use them for syntax extension, but not monad transformers, as [they box so easily](http://scastie.org/10882).

Atry:
I understand AnyVal improve performance for some cases, but not for other cases.

So what's the problem?

edmundnoble:
@S11001001 I'm confused as well here; it seems like AnyVal can only offer an advantage, despite not practically removing *all* allocations or even *most*.

S11001001:
This boxes a `Task`.

```scala
val t = List(Task(...))
```

This unboxes the task and boxes it again.

```scala
t.map(identity)
```

The trouble is not so much the allocations it doesn't remove, as the ones it adds.

Atry:
So, in order to avoid boxing, is `scalaz.Tag` a better choice for monad transformers?

Atry:
Then we have to remove many higher-kinded functions `flatMap` on instances and use `Ops` methods instead.

Atry:
Another choice is use some thing like (shapeless.newtype](https://static.javadoc.io/com.chuusai/shapeless_2.12/2.3.2/shapeless/newtype$.html) for monad transformers.

S11001001:
Something like `Tag` would be a better choice for monad transformers. I talk about it a little in ["Speaking of higher-kinded types…"](https://failex.blogspot.com/2017/04/and-glorious-subst-to-come.html). Expanding on this requires a major, major redesign, though, outside the scope of this PR.

I wouldn't mind a tagged type with `Ops` methods for `Task`, though.

Atry:
`Tag` is not suitable here because it does not erase the underlying type. As a result, the methods on `Ops` will be shadowed by member methods on the underlying type.

S11001001:
It does erase it.

Atry:
So `scalaz.@@` is like shapeless's `newtype`

S11001001:
Yes, though in 7.3.x it will not be implemented with casts anymore. #1306

kevinmeredith:
Per https://github.com/scalaz/scalaz/pull/1342#issuecomment-292692097, @S11001001, how costly is the `box -> un-box -> box` versus simply not using an `AnyVal`, i.e. allocating from the start? Thanks

Atry:
@S11001001 `@@` is not good for monad transformer due to https://github.com/scala/bug/issues/10283

concurrent/src/main/scala/scalaz/concurrent/Timer.scala
xuwei-k: - https://github.com/scala/scala/pull/5402/files#diff-6b2403127534a61cd03d206241b53bf4R667
- https://github.com/scala/scala/pull/5402/files#diff-96e5344d2869e1cabd560d60d9019b63R7

special treat by scalac

S11001001:
@kevinmeredith It depends on how often you cross the barrier.  For example, where `xs: List[Task[...]]`, `xs map identity` performs `2*xs.length` boxes.  Invoking `^(someTask, someOtherTask)(...)` will box the two inputs, then box the result, then unbox it.

@Atry I think we ought to just do nothing for now, then.

Unless @julianmichael, who was [playing with these a while ago](https://groups.google.com/d/msg/scala-user/GvzJAUDq1e0/mdk8MdfTAwAJ), has some ideas wrt working around scala/bug#10283?

julianmichael:
@S11001001 I played around a bit with the example in [scala/bug#10283](https://github.com/scala/bug/issues/10283) to no avail... everything I've tried requires an import that's specific to the value containing the opaquely sealed type. Since you can just have one value define the type family for `Task` (while sealing it), I think it could kind of work, but it'd probably mess up the design in scalaz. The user would have to explicitly import the conversions or instances from `object Task` or some other object where they are defined—undoubtably with a wildcard import—instead of being able to just import the `Task` type and take advantage of implicit resolution to values defined in the companion object.

There may be some fancier workaround, but I suspect it'd be complex and not worth the effort.

Atry:
@julianmichael Do you have any idea to manage priorities of implicit values in `import xxx._`?

julianmichael:
@Atry Sorry, I don't understand the question. You can use the standard method putting lower priority implicits in superclasses of the `xxx` object you import from. That's what I did with, for example, the `LowerCaseStrings` value that contained my opaque sealed type in the email that @S11001001 linked—with the effect that it would convert a `LowerCaseString` to its operations that preserve its type before falling back to the underlying `String`.

Do you have something more specific in mind?

Atry:
@julianmichael I used to think that the `extends LowPriorityImplicits0` trick only applies to implicit values on companion objects. I don't know if it works for `import`ed implicits.

julianmichael:
@Atry Ah—yep, it works:
```
scala> import scala.language.implicitConversions
import scala.language.implicitConversions

scala> trait A { implicit def numFallback[A](x: A): Int = 1 }
defined trait A

scala> object B extends A { implicit def num(x: Unit): Int = 2 }
defined object B

scala> import B._
import B._

scala> () + 0
res0: Int = 2

scala> ((), ()) + 0
res1: Int = 1
```

S11001001:
We won't do this, as it will cause more boxing in practice. Maybe someone will try again with a new SIP designed for type-safe programming, and we can try with that one.

PR #1348, opened 2017-04-06
Improve EphemeralStream's weakMemo to be optimistic through CAS (Rebased)

jbgi:
Rebased/squashed version of #1104 with the change mentioned in https://github.com/scalaz/scalaz/pull/1104#issuecomment-292121444

### DIFF
--- a/core/src/main/scala/scalaz/EphemeralStream.scala
+++ b/core/src/main/scala/scalaz/EphemeralStream.scala
@@ -1,6 +1,7 @@
 package scalaz
 
 import java.lang.ref.WeakReference
+import java.util.concurrent.atomic.AtomicReference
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/EphemeralStream.scala
```diff
@@ -297,17 +298,25 @@ object EphemeralStream extends EphemeralStreamInstances {
...
-        val x = f
-        v = Some(new WeakReference(x))
-        x
+      @inline
+      def genNew(x: V, old: WeakReference[V]): V = {
+        if (ref.compareAndSet(old, new WeakReference(x)))
+          x
+        else {
+          val crt = ref.get.get
+          if (crt != null) crt
```
S11001001: ...oh, I get it. 😄

PR #1349, opened 2017-04-06
Uncurry ephemeralstream fold and Reducer constructors

jbgi:
This make EphemeralStream foldLeft/foldRight accept uncurried function.

Both args are always applied at the same times anyway.
This make it closer to Foldable api and should improve perf a little.

Also changed EphemeralStream reverse implementation to skip List creation.
This also allows to preserve head cell laziness.


jbgi:
going to rebase this before merge.

fommil:
Apologies, I should have checked the open PRs before sending mine. LGTM.

PR #1351, opened 2017-04-07
adding monadError instance for Future on 7.1.x

vmarquez:


core/src/main/scala/scalaz/std/Future.scala
```diff
@@ -5,7 +5,8 @@ import scala.concurrent.{Await, CanAwait, ExecutionContext, Future}
 import scala.concurrent.duration.Duration
 
 trait FutureInstances1 {
-  implicit def futureInstance(implicit ec: ExecutionContext): Monad[Future] with Cobind[Future] =
+
+  implicit def futureInstance(implicit ec: ExecutionContext): Cobind[Future] with MonadError[({type λ[α,β] = Future[β]})#λ, Throwable] with Catchable[Future] =
```
S11001001: I *think* you can make this bin-compatible by putting `Monad[Future]` first again. Yeah it's weird.

vmarquez: @S11001001 I tried this and no luck.  I can keep playing around, but any other suggestions are appreciated. :)

S11001001: @vmarquez Eh. Well, how about, give your signature a new name, and take off the `implicit` keyword from `futureInstance` and deprecate it.  It's a simpler case than #582; no new priority layers are needed, your instance should go where the current one does.  That will still yield a mima warning, but might be except-able.

vmarquez: I'm still a bit stuck, perhaps @xuwei-k or you can help me understand further. When I tried to *add* a method to FutureInstances I got this error

```
[error]  * method newFutureInstance(scala.concurrent.ExecutionContext)scalaz.Monad in trait scalaz.std.FutureInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("scalaz.std.FutureInstances.newFutureInstance")
```

xuwei-k: There are two kinds; **scalaz datatype** and **std datatypes**.

**scalaz datatypes**: We can define type class instances in companion object or its super **classes** (eg. `sealed abstract class KleisliInstances9`). We can add new methods in companion object or super classes with keep binary compatibility.

**std datatypes**: We can't define type class instances in companion object. We **should** define in **trait** (not class) due to scalaz has some import style
- _a la carte_ import `import scalaz.std.scalaFuture._`
- `import scalaz.std.AllInstances._`
- `import scalaz.Scalaz._`

We **can't** add new methods in trait with keep binary compatibility (scala 2.11 or earlier)


In other words, adding new **std datatype** instances is harder than **Scalaz datatype** in scalaz7 😔 


Here is a unfortunate way add new typeclass instance.
https://github.com/xuwei-k/scalaz/commit/9c246bd5cfeed81bb9bc6d0f1e586f321a946388

S11001001: @vmarquez This error should be except-able, because we only support the three `import` lines @xuwei-k mentions, and all these objects will be recompiled to include the method you've added.

vmarquez:
Thank you @xuwei-k and @S11001001 for helping me.  I will hopefully be able to help back port more stuff now...

PR #1352, opened 2017-04-08
Override Functor#widen for covariant std collections

TomasMikula:
... to avoid mapping over the collection.


TomasMikula:
@S11001001 Done.

PR #1354, opened 2017-04-11
add Foldable#distinctBy

xuwei-k:


core/src/main/scala/scalaz/Foldable.scala
```diff
@@ -325,6 +325,9 @@ trait Foldable[F[_]]  { self =>
         else seen
     }.reverse
 
+  def distinctBy[A, B: Equal](fa: F[A])(f: A => B): IList[A] =
+    distinctE(fa)(Equal.equalBy(f))
```
jbgi: alternatively `distinctE` could be implemented in term of `distinctBy` so that the explicit `Equal` would not be required.

tonymorris:
:+1:

core/src/main/scala/scalaz/Foldable.scala
xuwei-k: `distinctE` use `Foldable[IList].element` and it require `Equal`

jbgi: `distinctBy` could be implemented via `any` :
```scala
  def distinctBy[A, B](fa: F[A])(f: A => B)(implicit B: Equal[B]): IList[A] =
    foldLeft(fa, IList.empty[A]) {
      case (seen, a) =>
        if (!IList.instances.any(seen)(f.andThen(B.equal(f(a), _))))
          a :: seen
        else seen
    }.reverse
```
and then
```scala
def distinctE[A](fa: F[A])(implicit A: Equal[A]): IList[A] =
    distinctBy(fa)(identity)
```
But yeah, it is a very minor variation.

TomasMikula: @jbgi Perhaps you want the type of `seen` to be `IList[B]` to avoid allocation of two `Function1`s per each element of `fa`.

jbgi: @TomasMikula the thing is that `seen` is also returned by the method (reversed) so that would require the fold to usea tuple `(List[A], List[B])`, but the gain of not reapplying f again and again could worth it. It probably depends on use-cases. 
On the other hand those methods are O(n*n) anyway so not sure it is worth micro-optimizing the constant factor.

TomasMikula: @jbgi Ah, missed that. I don't know, this seems to double the number of allocations compared to @xuwei-k's version.

TomasMikula: There's still this allocation of an anonymous function for each element: `a2 => B.equal(b1, f(a2))`, but now it's only 3/2 of allocations of the original.

jbgi: Well no, it is the same, because the implementation of `Foldable.element` is 
```scala
def element[A: Equal](fa: F[A], a: A): Boolean = any(fa)(Equal[A].equal(a, _))
```
(also create a anonymous function)

TomasMikula: 👍

PR #1356, opened 2017-04-13
Add missing explicit return type on public api methods.

jbgi:
- Serve as documentation
- should improve compile time and 
- avoid accidentally leaking implementations details / break binary compatibility.

also some other minor cleanups I found along the way.


core/src/main/scala/scalaz/FingerTree.scala
```diff
@@ -968,28 +968,28 @@ object FingerTree extends FingerTreeInstances {
...
-  def two[V, A](a1: A, a2: A)(implicit measure: Reducer[A, V]) =
+  def two[V, A](a1: A, a2: A)(implicit measure: Reducer[A, V]): Two[V, A] =
     Two(measure.snoc(measure.unit(a1), a2), a1, a2)
 
-  def three[V, A](a1: A, a2: A, a3: A)(implicit measure: Reducer[A, V]) =
+  def three[V, A](a1: A, a2: A, a3: A)(implicit measure: Reducer[A, V]): Three[V, A] =
     Three(measure.snoc(measure.snoc(measure.unit(a1), a2), a3), a1, a2, a3)
 
-  def four[V, A](a1: A, a2: A, a3: A, a4: A)(implicit measure: Reducer[A, V]) =
+  def four[V, A](a1: A, a2: A, a3: A, a4: A)(implicit measure: Reducer[A, V]): Four[V, A] =
```
S11001001: These four should all be ascribed `Finger[V, A]`.

core/src/main/scala/scalaz/Injective.scala
```diff
@@ -14,32 +14,32 @@ case class Injective4[T[_,_,_,_]]()
...
+  implicit def Product3Injective: Injective3[Product3] = Injective3[Product3]()
+  implicit def Product4Injective: Injective4[Product4] = Injective4[Product4]()
+  implicit def Product5Injective: Injective5[Product5] = Injective5[Product5]()
+  implicit def SetInjective: Injective[Set] = Injective[Set]()
+  implicit def Tuple2Injective: Injective2[Tuple2] = Injective2[Tuple2]()
+  implicit def Tuple3Injective: Injective3[Tuple3] = Injective3[Tuple3]()
+  implicit def Tuple4Injective: Injective4[Tuple4] = Injective4[Tuple4]()
+  implicit def Tuple5Injective: Injective5[Tuple5] = Injective5[Tuple5]()
+  implicit def ValidationInjective: Injective2[Validation] = Injective2[Validation]()
+  implicit def StreamInjective: Injective[Stream] = Injective[Stream]()
```
S11001001: We can remove the explicit tparams on the RHS; they should be inferred from the ascription.

jbgi:
Btw, I left out explicit return type on [some implicit defs](https://github.com/scalaz/scalaz/blob/series/7.3.x/effect/src/main/scala/scalaz/effect/MonadIO.scala#L43), because defining return type impacted implicit resolution and resulted in a infinite loop implementation. (all because of subtyping of course)

PR #1357, opened 2017-04-14
Parallel `Applicative` transformers for `FreeT`.

Atry:
Given an `Applicative` for `F[A] @@ Parallel`, there should be an `Applicative` for `FreeT[S, F, A] @@ Parallel` available as well.

For example, this patch allows running a `FreeT[S, Task, ?]` in parallel, enabling Dependency Injection feature in a `Task`.

### DIFF
--- a/core/src/main/scala/scalaz/FreeT.scala
+++ b/core/src/main/scala/scalaz/FreeT.scala
@@ -158,7 +158,59 @@ sealed abstract class FreeT[S[_], M[_], A] {
     }
 }
 
-sealed abstract class FreeTInstances6 {
+sealed abstract class FreeTInstances7 {
+  import scalaz.Tags.Parallel
+  implicit def freeTParallelApplicative[S[_], F[_]](
+      implicit FS: Functor[S],
+      BF: BindRec[F],
+      AF: Applicative[F],
+      parallelApplicativeF: Applicative[Lambda[X => F[X] @@ Parallel]]
+  ): Applicative[Lambda[X => FreeT[S, F, X] @@ Parallel]] =
+    new Applicative[Lambda[X => FreeT[S, F, X] @@ Parallel]] {
+      override def point[A](a: => A): FreeT[S, F, A] @@ Parallel = {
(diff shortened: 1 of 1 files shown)


wedens:
Isn't it basically `FreeApplicative`?

Atry:
I don't think `scalaz.FreeAp` can provide similar functions:

``` scala
implicit def freeTParallelApplicative[S[_], F[_]](
  implicit FS: Functor[S],
  BF: BindRec[F],
  AF: Applicative[F],
  parallelApplicativeF: Applicative[Lambda[X => F[X] @@ Parallel]]
): Applicative[Lambda[X => FreeT[S, F, X] @@ Parallel]]
```

Atry:
Any idea

TomasMikula:
@Atry It seems to me that this parallel `ap(fa, ff)` will synchronize on corresponding occurrences of `F` in `fa` and `ff`, as if zipping them. This is in contrast to running both to completion and only synchronize at the end.

Also this `ap` introduces a bunch of pure `F`s in the result. As a consequence, if you then `ap` the result with something else, this second `ap` will synchronize on a lot of pure `F`s, thus many times one side of the parallel computation will do basically nothing.

Maybe you want to consider using`FreeAp[Free[S, F, ?], A]` instead?

Atry:
@TomasMikula Do you mean `FreeAp[FreeT[S, F, ?], A]` ?

TomasMikula:
Yeah, fixed the typo. Wouldn't that work better for you?

Atry:
Yes. `FreeAp[FreeT[S, F, ?], A]` looks good?

Do you think we should add Monad transformers for `FreeAp`?

TomasMikula:
I'm not exactly sure what you mean by that.

But `FreeApT[S[_], F[_], A]`, a transformer version `FreeAp[S[_], A]`, could be useful.

Atry:
I wonder if `FreeAp[F[_], A]` itself could be a transformer, forwarding type class calls to `F`.

Atry:
The behavior is:
  * Creating a lazy AST for `ap` and all other `Applicative` methods
  * Forwarding `flatMap` and other `Monad` calls that beyond `Applicative` to `F` immediately

TomasMikula:
Yeah, OK, I think that makes sense 👍

PR #1359, opened 2017-04-17
profunctor composition and fixing some issues with profunctor stuff

vmarquez:
Should this be called Free Category?  @puffnfresh might know.

### DIFF
--- a/base/src/main/scala/data/ProCompose.scala
+++ b/base/src/main/scala/data/ProCompose.scala
@@ -1,0 +1,54 @@
+package scalaz
+package data 
+
+object ProCompose {
+  import typeclass.{Profunctor, Strong, Category, StrongClass, CategoryClass}
+  import typeclass.Strong._
+  import typeclass.Profunctor._
+
+  trait ProCompose[P[_, _], A, B] { } 
+
+  case class Pure[P[_, _], A, B](pab: A => B) extends ProCompose[P, A, B]
+
+  case class Composed[P[_, _], A, X, B](pab: P[A, X], more: ProCompose[P, X, B]) extends ProCompose[P, A, B]
+
+  implicit def ProComposeStrong[P[_, _]: Strong]: StrongClass.First[ProCompose[P, ?, ?]] = new StrongClass.First[ProCompose[P, ?, ?]] with CategoryClass[ProCompose[P, ?, ?]] { 
+    override def dimap[A, B, C, D](fab: ProCompose[P, A, B])(fca: C => A)(fbd: B => D): ProCompose[P, C, D] = 
+      fab match {
+        case Composed(pab, more) => Composed(pab.lmap(fca), rmap(more)(fbd))
+        case Pure(f) => Pure(f.dimap(fca)(fbd))
+      }   
+
+    override def rmap[A, B, D](fab: ProCompose[P, A, B])(fbd: B => D): ProCompose[P, A, D] =
+      fab match {
+        case Composed(pab, more) => Composed(pab, rmap(more)(fbd))
+        case Pure(f) => Pure(f.rmap(fbd))
+      }  
+
+    override def lmap[A, B, C](fab: ProCompose[P, A, B])(fca: C => A): ProCompose[P, C, B] =
+      fab match {
+        case Composed(pab, more) => Composed(pab.lmap(fca), more)
+        case Pure(f) => Pure(f.lmap(fca))
+      }   
+  
+    override def first[A, B, C](fab: ProCompose[P,A,B]): ProCompose[P,(A, C),(B, C)] = 
+      fab match {
+        case Composed(pab, more) => Composed(Strong[P].first(pab), first(more))
+        case Pure(f) => Pure(Strong[Function].first(f))
+      }   
+
+    override def second[A, B, C](fab: ProCompose[P,A,B]): ProCompose[P,(C, A),(C, B)] = 
+      fab match {
+        case Composed(pab, more) => Composed(Strong[P].second(pab), second(more))
+        case Pure(f) => Pure(Strong[Function].second(f))
+      }
+
+    override def id[A]: ProCompose[P,A,A] = Pure((a: A) => a)
+    
+    override def compose[A, B, C](fbc: ProCompose[P, B, C], fab: ProCompose[P, A, B]): ProCompose[P, A, C] = fab match {
+        case Composed(pab, more) =>  
+          Composed(pab, compose(fbc, more)) 
+        case Pure(f) => lmap(fbc)(f) 
+      }
+  }
+}
--- a/base/src/main/scala/typeclass/ChoiceClass.scala
+++ b/base/src/main/scala/typeclass/ChoiceClass.scala
@@ -8,13 +8,13 @@ trait ChoiceClass[P[_, _]] extends Choice[P] with ProfunctorClass[P] {
 }
 
 object ChoiceClass {
-  trait Left[P[_, _]] extends Alt[Left[P]] { self : Choice[P] =>
+  trait Left[P[_, _]] extends Alt[Left[P]] with Choice[P] { self : Choice[P] =>
     override def leftchoice[A, B, C](pab: P[A, B]): P[A \/ C, B \/ C]
     override def rightchoice[A, B, C](pab: P[A, B]): P[C \/ A, C \/ B] =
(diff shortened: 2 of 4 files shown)


TomasMikula:
Category would need identity morphisms:

```scala
case class Identity[P[_, _], A]() extends ProCompose[P, A, A]
```

I called this structure without identities `FreeCompose` [elsewhere](https://github.com/TomasMikula/type-aligned/blob/master/src/main/scala/typealigned/FreeCompose.scala).

base/src/main/scala/data/ProCompose.scala
```diff
@@ -0,0 +1,54 @@
...
+object ProCompose {
+  import typeclass.{Profunctor, Strong, Category, StrongClass, CategoryClass}
+  import typeclass.Strong._
+  import typeclass.Profunctor._
+
+  trait ProCompose[P[_, _], A, B] { } 
+
+  case class Pure[P[_, _], A, B](pab: A => B) extends ProCompose[P, A, B]
+
+  case class Composed[P[_, _], A, X, B](pab: P[A, X], more: ProCompose[P, X, B]) extends ProCompose[P, A, B]
```
edmundnoble: This is `Free2` over `Prearrow` (`Semigroupoid` (compose) + `arr` (injection from function), see Tony's proposed type class hierarchy). As well, the lifting injection `P ~~> ProCompose[P, ?, ?]` is unnecessarily complicated and inefficient: `(pab: P[A, B]) => Composed(pab, Pure(identity[B]))`. I recommend creating an `Ops` trait with cases for `P[A, B]` and `A => B`, and making `Composed` reference it instead of `P[_, _]`; if variants of this structure will be common later, you may get more mileage out of implementing this as `TypeAlignedList[Ops]`.

More importantly, there is no stack-safe usage of this structure; unless you introduce an `Evaluable`-like type class:
```scala
  trait Evaluable[F[_, _], G[_]] {
    def eval[A, B](a: G[A], arr: F[A, B]): G[B]
  }
```

edmundnoble:
@TomasMikula `Pure` provides an identity via `Pure(identity)`.

TomasMikula:
@edmundnoble Only when there's already identity on `P`, i.e. `def identity[A]: P[A, A]`. It doesn't provide a free identity.

TomasMikula:
@edmundnoble Ah, I overlooked that `Pure` wraps a function `A => B`. I was assuming it wraps `P[A, B]`.

base/src/main/scala/data/ProCompose.scala
TomasMikula: > there is no stack-safe usage of this structure

Why?

edmundnoble: Try to implement it yourself; for example, use `Kleisli` over `F`. You don't know enough from the definition of `ProCompose` to actually apply the function to the value; so you get stuck implementing it in a tail-recursive way. If you do it the obvious way using `Kleisli.compose` it's not stack-safe.

TomasMikula: > Try to implement it yourself

What is "it"? What is given to me and what should I try to implement?

Seems like you have one specific usage in mind that is not stack-safe, and even that only when "you do it the obvious way". That is different than saying "there is no stack-safe usage of this structure".

For example, this structure supports balanced composition (e.g. of `Kleisli`s), which will be stack safe for all practical instances of `ProCompose[Kleisli[F, ?, ?], A, B]`.

Also, using this with a something like `Evaluable` seems like a natural use of `ProCompose`. Do you have a usage in mind that doesn't necessarily involve `Evaluable`, but cannot be made stack-safe without it?

edmundnoble: Sorry for being unspecific; when I say "it", I mean an implementation of this `ProCompose` which has a stack-safe `foldMap` function, for *all* `F[_, _]` with a `Semiarrow` instance. It *is* possible to write a stack-safe `foldMap` for a given  `ProCompose`, *but* it requires writing specific code for every `F[_, _]` and is thus impossible to abstract over. That's the problem `Evaluable` solves.

TomasMikula: Not sure what exactly `Semiarrow` is (google doesn't help either), but if it gives you at least `arr` and `compose`, then you can.

> I'm curious why you think this is either balanced composition

I didn't say this _is_ balanced composition, but that this _supports_ balanced composition. A better word would have been _allows_ or _admits_.

edmundnoble: `Semiarrow` is just what one person called this construct on the Haskell mailing list; I don't know of a name that's been stable for a while for this, nor a few others in the `Semigroupoid->Strong/Choice` hierarchy.

My mistake, I misread; balanced composition sounds possible to me. My last concern is this: you need to pay a `O(n)` cost every time you compose a `ProCompose[F, A, B]` with a `ProCompose[F, B, C]` to perform the associated list concatenation, because of the cons-list structure. That could be fixed with a nice catenable `Queue` but we've no such thing yet; so `Vector` may be worth considering.

TomasMikula: 👎  for `Vector`, because:
 - not type aligned, thus would require `asInstanceOf`s;
 - a naive catenable will have cheaper `cons`, `snoc`, `++`, random access is not required, and IMO _O(1)_ `uncons` is not really necessary either, only that _n_ successive `uncons` will be _O(n)_.

It's easy to change the constructors so that a naive catenable is basically baked in.

edmundnoble: I agree, Vector is crap, and so are casts. It's totally possible to make this catenable and type-aligned; but it's not as generic as making the type-aligned sequence (free category over F) and then making this `ProCompose` into a free category over `F :+: Function1`. I feel like free categories are a useful addition to Scalaz 8, but if you disagree I can't see any other issue with making this catenable.

TomasMikula: > I feel like free categories are a useful addition to Scalaz 8, but if you disagree

No, I agree. I don't see it as a prerequisite for this PR, though.

In fact, `FreeCategory[F :+: Function1]` is slightly redundant for this&mdash;there are two ways to express identity morphisms:
 - `FreeCategory.lift(\/-(identity[A]))`
 - `FreeCategory.identity[A]`

`FreeCompose` would be sufficient.

And `FreeCategory[F]` could be expressed as `FreeCompose[F :+: ===]`. Though using these more generic building blocks costs in extra boxing, so I wouldn't mind occasionally just bake in a specialized version.

edmundnoble: It's not for the identity morphisms, it's for the `arr` function.

TomasMikula: What is? The redundancy is for identity morphisms.

edmundnoble: Yes, I noticed, that's very clever. I'm saying this `ProCompose` is a free `SemiArrow` cause it has the injection `(A => B) => ProCompose[F, A, B]`. Free categories don't have that.

TomasMikula: And once you add `arr` to free category, you have two representations of identity morphisms. (Not necessarily a big deal, though, just pointing that out.) To avoid that, you could start with free `Compose`.

edmundnoble:
Yeah; not sure myself practically how much of an issue that is, though I bet you do, given you wrote FreeBind ;)

TomasMikula:
Oh, I also wrote [`FreeCompose`](https://github.com/TomasMikula/type-aligned/blob/master/src/main/scala/typealigned/FreeCompose.scala), including the balanced composition (`reduce`) and your `Evaluable` is called `FunctorLike` there.

jdegoes:
@vmarquez This needs some TLC when you get a chance.

TomasMikula:
So what would be a good name for this? It provides `Profunctor` and `Category` for free, so `FreeProCategory`?

As @edmundnoble pointed out, `Profunctor with Category` is also called a `Prearrow` (an `Arrow` without `first`), so maybe `FreePrearrow`?

jdegoes:
> As @edmundnoble pointed out, Profunctor with Category is also called a Prearrow (an Arrow without first), so maybe FreePrearrow?

👍 

We should also consider taking the time to rename `Free` -> `FreeMonad` to have a globally consistent naming convention.

PR #1361, opened 2017-04-21
Minor improvements to Atomic and BooleanLatch

jbgi:


concurrent/src/main/scala/scalaz/concurrent/Atomic.scala
```diff
@@ -4,6 +4,7 @@ package concurrent
 import effect._
 
 import java.util.concurrent.atomic.AtomicReference
+import java.util.function.UnaryOperator
```
S11001001: @xuwei-k Do we still support Java 7 bd41ba63a3f00c00617b4e1a776ff45842fa8dc0 ?

xuwei-k: No, I think we should drop Java 7 in Scalaz 7.3. The reason for [`-target:jvm-1.7`](https://github.com/scalaz/scalaz/commit/bd41ba63a3f00c00617b4e1a776ff45842fa8dc0), Scala 2.10.6 does not accept `-target:jvm-1.8` and default is `-target:jvm-1.6`

jbgi:
should be merge-able.

jdegoes:
@jbgi Looks great, should we change jvm:target to 1.8?

jbgi:
@jdegoes should already be the case: https://github.com/scalaz/scalaz/blob/series/7.3.x/project/build.scala#L110 (and it's by default for scala 2.12).

jdegoes:
🙏

PR #1362, opened 2017-04-22
Adds `or` to NaturalTransformation

### LINKED ISSUE #1222: Lack of `or` method on NaturalTransformation
Given some algebra `X` and some other algebra `Y`, we might have a program

```
type DSL[A] = Coproduct[X, Y, A]
val prog: Free[Dsl, Unit] = ???
```

Given we already have some interpreters for both X and Y

```
val interX: X ~> Id = ???
val interY: Y ~> Id = ???
```

We want to interpret program `prog` using a new interpreter that combines `interX` and `interY`.

ScalaZ misses a convenient method `or` that would do just that so we could

```
progr.foldMap(interX or interY)
```

lloydmeta:
Closes #1222 

~~I couldn't add this directly on `NaturalTransformation` because of the variance markers on that trait for `F[_]` and `G[_]`~~

### DIFF
--- a/core/src/main/scala/scalaz/NaturalTransformation.scala
+++ b/core/src/main/scala/scalaz/NaturalTransformation.scala
@@ -21,6 +21,16 @@ trait NaturalTransformation[-F[_], +G[_]] {
 
   def andThen[H[_]](f: G ~> H): F ~> H =
     f compose self
+
+  /**
+    * Combines this [[scalaz.NaturalTransformation]] with another one to create one
+    * that can transform [[scalaz.Coproduct]].
+    *
+    * The current NaturalTransformation will be used to transform the Left (`F`) value of
+    * the [[scalaz.Coproduct]] while the other one will be used to transform the Right (`H`) value.
+    */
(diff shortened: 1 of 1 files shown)


lloydmeta:
I've tried multiple ways of writing this, but can't get over the following failure under 2.10

```
object creation impossible, since method apply in trait NaturalTransformation of type [A](fa: scalaz.Coproduct[F,H,A])G[A] is not defined
```

Any ideas?

TomasMikula:
Nice. You could have this on `NaturalTransformation` directly, though:

```scala
def or[H[_], F0[A] <: F[A], G0[A] >: G[A]](hg: H ~> G0): Coproduct[F0, H, ?] ~> G0 =
  λ[Coproduct[F0, H, ?] ~> G0](_.fold(self, hg))
```

lloydmeta:
@TomasMikula Thanks for the suggestion. Implemented in e18bf2e

TomasMikula:
If you squash the commits, this looks good to me 👍

lloydmeta:
@TomasMikula done :)

PR #1366, opened 2017-04-25
Add lazy constructor/map/flatMap on IList (experimental)

jbgi:
This an experiment to add a lazy constructor (`byNeed`) to `IList`.

The main downside is that you cannot pattern match directly anymore. Eg.
```scala
  def uncons[B](n: => B, c: (A, IList[A]) => B): B =
    this match {
      case INil() => n
      case ICons(h, t) => c(h, t)
    }
```
must be changed to either:
```scala
  def uncons[B](n: => B, c: (A, IList[A]) => B): B =
    this.value match {
      case INil() => n
      case ICons(h, t) => c(h, t)
    }
```
or
```scala
  def uncons[B](n: => B, c: (A, IList[A]) => B): B =
    this match {
      case IList(INil()) => n
      case IList(ICons(h, t)) => c(h, t)
    }
```

One usage of this `byNeed` constructor is [`mapLazy`](https://github.com/jbgi/scalaz/blob/6b416fcbd5cab89073c606bc2ebdbfec5167510f/core/src/main/scala/scalaz/IList.scala#L254), which may be used to implement efficiently things like: 
```scala
veryLargeList.mapLazy(f).take(10)
```

This approach could be generalized to other data types, and remove the need for `LazyOption` / `LazyEither`.

### DIFF
--- a/core/src/main/scala/scalaz/IList.scala
+++ b/core/src/main/scala/scalaz/IList.scala
@@ -2,18 +2,19 @@ package scalaz
 
 import scala.annotation.tailrec
 import scala.annotation.unchecked.uncheckedVariance
-import std.option.{ cata, none, some }
-import std.stream.{ toZipper => sToZipper }
-import std.tuple.{ tuple2Bitraverse => BFT }
-import Liskov.{ <~<, refl }
-import IList.{empty, single}
+import std.option.{cata, none, some}
+import std.stream.{toZipper => sToZipper}
+import std.tuple.{tuple2Bitraverse => BFT}
+import Liskov.{<~<, refl}
+import IList.{empty, foldRightLazy, single, byNeed}
+import scalaz.std.option
 
 /**
  * Safe, invariant alternative to stdlib `List`. Most methods on `List` have a sensible equivalent
  * here, either on the `IList` interface itself or via typeclass instances (which are the same as
  * those defined for stdlib `List`). All methods are total and stack-safe.
  */
-sealed abstract class IList[A] extends Product with Serializable {
+trait IList[A] extends Product with Serializable {
 
   // Operations, in alphabetic order
 
@@ -50,7 +51,7 @@ sealed abstract class IList[A] extends Product with Serializable {
 
   def collect[B](pf: PartialFunction[A,B]): IList[B] = {
     @tailrec def go(as: IList[A], acc: IList[B]): IList[B] =
-      as match {
+      as.value match {
         case ICons(h, t) =>
           if(pf isDefinedAt h) go(t, ICons(pf(h), acc))
           else go(t, acc)
@@ -65,6 +66,9 @@ sealed abstract class IList[A] extends Product with Serializable {
   def concat(as: IList[A]): IList[A] =
     foldRight(as)(_ :: _)
 
+  final def concatLazy(as: IList[A]): IList[A] =
+    foldRightLazy[A, IList[A]](as, (h, tail) => byNeed(ICons(h, tail)))(this)
+
   // no contains; use Foldable#element
 
   def containsSlice(as: IList[A])(implicit ev: Equal[A]): Boolean =
@@ -75,7 +79,7 @@ sealed abstract class IList[A] extends Product with Serializable {
 
   def distinct(implicit A: Order[A]): IList[A] = {
     @tailrec def loop(src: IList[A], seen: ISet[A], acc: IList[A]): IList[A] =
-      src match {
+      src.value match {
         case ICons(h, t) =>
           if(seen.notMember(h)){
             loop(t, seen.insert(h), h :: acc)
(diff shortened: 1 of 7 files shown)


TomasMikula:
What's the difference then (from user's perspective) from `StreamT[Id, A]`?

jbgi:
@TomasMikula it is still a List, with same or better performance characteristics. From what I understand StreamT[Id] is more like CorecursiveList. (?)

TomasMikula:
Seems to me that the `unfold` method

```scala
def unfold[A,B](b: B)(f: B => Option[(A,B)]): StreamT[Id,A]
```

used to create corecursive lists can now be implemented on `IList` with lazy `flatMap`:

```scala
def unfold[A,B](b: B)(f: B => Option[(A,B)]): IList[A] =
  IList[A \/ B](\/-(b)).flatMapLazy({
    case -\/(a) => IList(a)
    case \/-(b) => f(b) match {
      case Some((a, b)) => IList(-\/(a), \/-(b))
      case None => IList()
    }
  })
```

jbgi:
That's true. The main difference is that IList will memoize the whole sequence.

core/src/main/scala/scalaz/IList.scala
```diff
@@ -2,18 +2,19 @@ package scalaz
...
+import IList.{empty, foldRightLazy, single, byNeed}
+import scalaz.std.option
 
 /**
  * Safe, invariant alternative to stdlib `List`. Most methods on `List` have a sensible equivalent
  * here, either on the `IList` interface itself or via typeclass instances (which are the same as
  * those defined for stdlib `List`). All methods are total and stack-safe.
  */
-sealed abstract class IList[A] extends Product with Serializable {
+trait IList[A] extends Product with Serializable {
```
xuwei-k: Why change to `trait` ? There is an advantage for binary-compatibility keepability.

core/src/main/scala/scalaz/IList.scala
```diff
@@ -467,24 +483,61 @@ sealed abstract class IList[A] extends Product with Serializable {
...
+sealed abstract case class ICons[A](head: A, tail: IList[A]) extends StrictIList[A] { self: IList[A] =>
+  override def value: StrictIList[A] = this
+  override def list: IList[A] = this
+}
+object ICons {
+  def apply[A](head: A, tail: IList[A]): IList[A]  = new ICons[A](head, tail) with IList[A] {}
+}
+private final class ByNeed[A](private[this] var list: () => IList[A]) extends IList[A] {
+  override lazy val value = {
+    val value0 = list().value
```
TomasMikula: This line could potentially lead to stack overflow, if there are too many `ByNeed` lists to be evaluated before a `StrictIList` is encountered, right?

core/src/main/scala/scalaz/IList.scala
jbgi: This was to be able to hide sub-typing between `INil`/`ICons` and `IList` to avoid any confusion when pattern matching. But maybe it is overkill. [Reverting back to abstract class](https://github.com/jbgi/scalaz/commit/231946bf5c89cac6cd3c159a21462d4615292e46) is possible.

core/src/main/scala/scalaz/IList.scala
jbgi: yes I guess so. Do you have a a situation in mind where such a pathological nesting could happen?
I guess it could be workaround with a loop.

core/src/main/scala/scalaz/IList.scala
jbgi: @xuwei-k is the binary-compatibility advantage of abstract class real if client code does not implement IList? because otherwise making the trait `sealed` should be enough... maybe?

core/src/main/scala/scalaz/IList.scala
TomasMikula: If I implement `filterLazy` using `flatMapLazy` and then filter out too many elements in a row.

jbgi: @TomasMikula good point. ByNeed is now properly "trampolined".

TomasMikula:
Looking good. The question is whether this should be a separate data structure from `IList`. It can be confusing that `IList#map` is strict, but `Functor[IList].map` is lazy.

Should also `traverseImpl` use `foldRightLazy`? Would it enable short-circuiting `traverse` when the target `Applicative` is e.g. `Option`?

Before it goes in, tests that it is actually lazy, memoizing, short-circuiting and stack-safe would be nice.

jbgi:
@TomasMikula I used `lazyMap` in `Functor` mostly to trigger the tests.
Eventually I was thinking of a `Lazy` tag to kick-in lazy implementation of type classes... ?

TomasMikula:
I guess that would work.

Though I wonder whether there is a place for a strict list in `scalaz`. Personally, I don't care much, because std `List` works for me most of the time anyway. Though this also makes `NonEmptyList` potentially lazy.

PR #1367, opened 2017-04-26
Drop 2.10 support, bump 2.11.x version to 2.11.11, enable -Ypartial-unification

TomasMikula:


xuwei-k:
wait scala-native 0.2 or use 2.11.8 if scala-native build?

https://github.com/scala-native/scala-native/issues/629

TomasMikula:
The source code doesn't compile both with and without `-Ypartial-unification`, so will have to wait.

jbgi:
Along with scala 2.11.11, I think we should use the `"-Xsource:2.12"`compiler flag, to benefits from things like https://github.com/scala/scala/pull/5846.
Maybe also to consider:
- `-Xlint`
- and others flags from https://tpolecat.github.io/2017/04/25/scalac-flags.html

ps: looks like [scala-native 0.2 is on its way](https://github.com/scala-native/scala-native/commit/6300153bd1e3ad4bead4f00fdb403da8e7a29c89).

TomasMikula:
Updated to use `scala-native` 0.2.

Not sure what this is for

```scala
addSbtPlugin("org.scala-native" % "sbt-scalajs-crossproject" % "0.1.0")
```

and whether it needs to be updated to `0.2.0` as well. If yes, what `sbt` command do I run to make it fail?

@jbgi Those are good suggestions, but I just wanted to do the minimal changes necessary to get partial unification, not being ready to address warnings due to `-Xlint` or whatnot.

xuwei-k:
https://github.com/xuwei-k/scalaz/commit/c7c780cfdae56e80a7a1d92f94265034c9798c30

TomasMikula:
Thanks @xuwei-k, added that.

PR #1370, opened 2017-04-27
Add Future.safeStep and safePerformListen	

Atry:


emilypi:
Outdated as of https://github.com/scalaz/scalaz/pull/2037

PR #1371, opened 2017-04-27
Laws for MonadPartialOrder: must be injective monad homomorphism.

TomasMikula:
This is an attempt to give reasonable laws for `MonadPartialOrder` (a.k.a. `|>=|`). At the very least, it should be
 - a _monad homomorphism_ (preserve `point` and `bind`); and
 - _injective_, as the expectation is that translation to the "bigger" monad doesn't lose any information.

Since we provide `MonadPartialOrder` instances for monad transformers, I added the injectivity requirement on `MonadTrans` as well. This shouldn't be constraining in practice, since (I believe) we expect transformers to be injective (not lose information in `liftM`) anyway. (I think that due to parametricity, the only otherwise lawful transformer instances this rules out are the ones that ignore the monad type parameter altogether, such as `type FakeTrans[M[_], A] = Foo[A]`.)

(I also renamed the other two `MonadTrans` laws, as I think I named them inappropriately before.)

Additional minor change: Avoid instantiating the reflexive identity instance when creating `MonadPartialOrder` instance for transformers.

Note: These new `MonadPartialOrder` laws are not used anywhere in tests yet. This is because the only instances we provide are for transformers, and transformer laws already have all the laws of `MonadPartialOrder`. However, if we start providing other instances, such as `List |>=| Option` or `ReaderWriterStateT[F, R, W, S, ?] |>=| WriterT[F, W, ?]`, there will now be laws to test.

### DIFF
--- a/core/src/main/scala/scalaz/MonadTrans.scala
+++ b/core/src/main/scala/scalaz/MonadTrans.scala
@@ -16,13 +16,18 @@ trait MonadTrans[F[_[_], _]] {
   implicit def apply[G[_] : Monad]: Monad[F[G, ?]]
 
   trait MonadTransLaw {
+    // For each monad `G[_]`, `liftM` must be injective monad homomorphism from `G` to `F[G, ?]`.
+
     /** Lifted `point` is `point` of this transformer's monad. */
-    def identity[G[_], A](a: A)(implicit G: Monad[G], FA: Equal[F[G, A]]): Boolean =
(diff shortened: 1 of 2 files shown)


vmarquez:
Just saw this, apologies.  👍 I think this is cool and we should add the other instances like you talked about.

xuwei-k:
https://github.com/scalaz/scalaz/pull/1432

PR #1376, opened 2017-04-29
Remove unnecessary implicit parameter Applicative[S] in Free.suspend

Atry:


Atry:
@xuwei-k Could you review this PR please?

Atry:
Any idea? @xuwei-k @S11001001 @TomasMikula

TomasMikula:
Good catch on the unnecessary `Applicative`!

As for the caching, I would at least localize the `asInstanceOf` more, such as

```scala
@inline private def pointUnit[S]: Free[S, Unit] = pointUnitCache.asInstanceOf[Return[S, Unit]]
```

Atry:
@TomasMikula `pointUnit` done.

PR #1377, opened 2017-04-29
Add FreeContsTBindRec	

Atry:
Allow stack-safe `ContT` mentioned in https://github.com/scalaz/scalaz/pull/1368#issuecomment-297921691

### DIFF
--- a/core/src/main/scala/scalaz/IndexedContsT.scala
+++ b/core/src/main/scala/scalaz/IndexedContsT.scala
@@ -106,7 +106,14 @@ trait IndexedContsTFunctions {
     }
 }
 
-sealed abstract class IndexedContsTInstances2 {
+sealed abstract class IndexedContsTInstances3 {
+  implicit def FreeContsTBindRec[W[_], S[_], R](implicit W0: Cobind[W]): BindRec[ContsT[W, Free[S, ?], R, ?]] =
+    new FreeContsTBindRec[W, S, R] {
+      val W = W0
+    }
+}
+
+sealed abstract class IndexedContsTInstances2 extends IndexedContsTInstances3 {
   implicit def IndexedContsTFunctorRight[W[_], M[_], R, O](implicit W0: Functor[W]): Functor[IndexedContsT[W, M, R, O, ?]] =
     new IndexedContsTFunctorRight[W, M, R, O] {
       implicit val W: Functor[W] = W0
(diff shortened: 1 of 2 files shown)


TomasMikula:
Important point to note is that this is not a general stack-safe `ContT[M[_], R, A]`, but only for the case when `M = Free[S, ?]`. As mentioned in a [follow-up comment](https://github.com/scalaz/scalaz/pull/1368#issuecomment-298197396), a `Monad` instance specialized for the case `M = Free` could be provided as well, and seems sufficient for stack-safety.

Atry:
I added more stack-safe type classes	and tests.

core/src/main/scala/scalaz/IndexedContsT.scala
```diff
@@ -145,10 +167,9 @@ abstract class IndexedContsTInstances extends IndexedContsTInstances0 {
       implicit val M: Functor[M] = M0
     }
 
-  implicit def ContsTMonadPlus[W[_], M[_], R](implicit W0: Comonad[W], M0: PlusEmpty[M]): MonadPlus[ContsT[W, M, R, ?]] =
-    new ContsTMonadPlus[W, M, R] {
+  implicit def FreeContsTMonad[W[_], S[_], R](implicit W0: Comonad[W]): Monad[ContsT[W, Free[S, ?], R, ?]] with BindRec[ContsT[W, Free[S, ?], R, ?]] =
+    new FreeContsTBindRec[W, S, R] with ContsTMonad[W, Free[S, ?], R] {
```
TomasMikula: Doesn't `with ContsTMonad` override `bind` back to the non-stack-safe one?

TomasMikula:
Looking good!

There should be tests demonstrating what is possible now that overflowed the stack before.

core/src/main/scala/scalaz/IndexedContsT.scala
Atry: `ContsTMonad` does not override `bind`. So it is not a problem.

Anyway, I swapped the order of super traits, ensuring that the final instance always pick stack-safe `bind` even if `ContsTMonad` provide another `bind` implementation.

TomasMikula: 👍

TomasMikula:
I like this, but I would like to know other people's opinion on the fact that this is an ad-hoc instance specialized to `Free` and, as a consequence, `MonadTrans[ContT[?[_], R, ?]].apply[Trampoline]` and Monad[ContT[Trampoline, R, ?]]` are monads with different stack behavior.

emilypi:
@Atry on this one, i think we just need to sort out the resolution and we're good to go on this one as well. Though, @TomasMikula does raise a good point. I don't necessarily agree with the idea that we should specialize to `Free`

Atry:
Another approach is creating an opaque type alias of `ContT[Trampoline, ?, ?]`, and provide special type instances for the opaque type alias.

I had implement this idea as a separate library: https://github.com/ThoughtWorksInc/future.scala

PR #1390, opened 2017-05-19
Add extrema, extremaOf, and extremaBy to Foldable

nuttycom:
Single-pass operations to determine the extrema of Foldable values.


core/src/main/scala/scalaz/Foldable.scala
```diff
@@ -228,6 +228,40 @@ trait Foldable[F[_]]  { self =>
...
+        val b = f(a)
+        some((b, b))
+      case (Some(x @ (bmin, bmax)), a) => 
+        val b = f(a) 
+        some(
+          if (Order[B].order(b, bmin) == LT) (b, bmax)
+          else if (Order[B].order(b, bmax) == GT) (bmin, b)
+          else x
+        )
+    }
```
TomasMikula: Just something to consider, using `foldMapLeft1Opt` instead of `foldLeft` would save you some lines of code.

core/src/main/scala/scalaz/Foldable.scala
```diff
@@ -228,6 +228,40 @@ trait Foldable[F[_]]  { self =>
...
+        some(
+          if (Order[B].order(b, bmin) == LT) (b, bmax)
+          else if (Order[B].order(b, bmax) == GT) (bmin, b)
+          else x
+        )
+    } 
+
+  /** The elements (amin, amax) of `fa` withc yield the smallest and largest values of `f(a)`, respectively, or None if `fa` is empty */
+  def extremaBy[A, B: Order](fa: F[A])(f: A => B): Option[(A, A)] =
+    foldLeft(fa, none[((A, A), (B, B))]) {
```
TomasMikula: Just me prematurely optimizing, but using a flat `Tuple4` would save one allocation per iteration.

PR #1392, opened 2017-05-20
Map (==>>) should have a Band instance

vmarquez:
Map (==>>) should have a Band instance for A ==>> B if there is a Band[B].


vmarquez:
Apologies @xuwei-k I put it in the wrong instance place.  Now there should not be a conflict with implicits.

Issue #1374, opened 2017-04-28
Add usage of disjunctions to the example package


abhishek7:
While disjunctions may appear to be straightforward, it would be great to provide a set of simple to relatively complex examples of using disjunctions in `scalaz/example/src/main/scala/scalaz/example`.

A couple of suggestions for the examples could be to start with general `match case` statements, and eventually start working with traits and disjunctions.

abhishek7:
While I am new to both programming in Scala and interacting with the Scalaz itself, would be more than happy to attempt this.

TomasMikula:
Absolutely, give it a go 👍

abhishek7:
@TomasMikula having some trouble compiling the project, do we have any specific set of steps on how to do so for contributors? If not, I can raise a separate issue for this.

TomasMikula:
I usually only compile for JVM. Inside `sbt`, run

```
coreJVM/compile
```

To compile the examples, run

```
example/compile
```

abhishek7:
@TomasMikula, any preferred way we should test our code before opening a PR? Noticed a lot of the examples has `assert` statements within, do we want to have them as part of the `test` package?

TomasMikula:
I view assertions in the examples more as a way of documenting (rather than testing) the behavior.

PR #1395, opened 2017-05-28
Disjunctions example 1374

abhishek7:
Simple example using `\/` in Scalaz, referencing https://github.com/scalaz/scalaz/issues/1374. Happy to update this as time goes on, or if there is anything in particular we'd like to see prior to merging.

### DIFF
--- a/example/src/main/scala/scalaz/example/DisjunctionUsage.scala
+++ b/example/src/main/scala/scalaz/example/DisjunctionUsage.scala
@@ -30,19 +30,11 @@ object DisjunctionUsage extends App {
     case _ => -\/(false)
   }
 
-  // Simple val that is assigned the value of 5.
-  // We'll use this as a parameter for our above function.
   val x = 5
-
-  // Getting the result, expecting -\/("Correct!")
   val expectingCorrect = infixNotation(x)
   assert(expectingCorrect === \/-("Correct!"))
 
-  // Let's try getting false with another val assigned the
-  // value of 6.
   val y = 6
-
-  // Getting the result, expecting \/-(false)
   val expectingFalse = infixNotation(y)
   assert(expectingFalse === -\/(false))
 
@@ -71,11 +63,6 @@ object DisjunctionUsage extends App {
     case _ => false.left
   }
 
-  /*
-     Rather than just assigning another val, we can assert
-     more elegantly:
-  */
   assert(bracketNotation(1) === -\/(false))
   assert(bracketNotation(5) === \/-("Correct!"))
-
 }


TomasMikula:
I don't think we should document standard Scala syntax, such as infix notation for symbolic names.

Things that _might_ be worth demoing:
 - creating a value of `A \/ B`;
 - pattern matching on `A \/ B`;
 - right biased functor instance;
 - ???

I would also avoid writing in comments that which is clear from the code.

abhishek7:
Thanks for the comments, will update the PR with the three things you proposed. It may or may not be worth having as a demo, but if I think of anything else while working on this, I'll add it on.

abhishek7:
@TomasMikula I'm trying to add on some examples around right biased functor instances now. Could you please explain what that is? I was doing some research on my own about this, but I haven't found anything useful yet. Please let me know if there are any resources I can refer to.

TomasMikula:
@puffnfresh 

> `\/.left` and `\/.right` should be used, not the Scala constructors.

After #1645, this should not matter. `-\/` and `\/-` smart constructors now return `\/`.

abhishek7:
@TomasMikula, are there any examples I can refer to for a right-biased functor instance? I asked a question a while back about this here: https://stackoverflow.com/questions/48738404/what-is-functor-bias, but the concept is still a bit unclear to me.

JordanMartinez:
@abhishek7 Given the following object
````scala
case class Box[A] (value: A) {}
````
a Functor will "open" the box, pass the `value` into a function `(A => B)` and put the result of the function into a new Box (`Box[B]`). There is no confusion as to which "value" to pass into the function because the `Box` type only has one possible value that could be passed into the function: `value`

Now, let's say we have a new data class:
````scala
case class Box2[A, B] (valueA: A, valueB: B) {}
````
If we wanted to use a `Functor` on `Box2[A, B]`, should we output `Box2[C, B]` or `Box2[A, C]` since we can only map one of the values? In other words, should we pass `valueA` into a function `(A => C)`? Or should we pass `valueB` into a function `(B => C)`?

A right-biased Functor means we should output `Box2[A, C]` by passing `valueB` into the function `(B => C)`. A left-biased Functor would take the opposite route: output `Box2[C, B] with `valueA` passed into the function (`A => C)`. The bias refers to which of the two types we should use in a function.

Does the answer to your StackOverflow question now make more sense? Given an `Either[A, B]`, which "value" do we pass into the function, the value stored in the `Left[A]` or the value stored in the `Right[B]`? Since an `Either` is usually used to distinguish between an error (the left side) and a correct value (the right side), we typically only want to map the `Either` when the instance is `Right[B]`. Hence, `Either` is a right-biased functor.

Returning to `Box2[A, B]`, if we wanted to map both values at once so that `Box[C, D] is outputted with functions `(A => C)` and `(B => D)`, we would use a `BiFunctor`.

TomasMikula:
@abhishek7 Yeah, this jargon is not very precise.

It doesn't make sense to say that `Either` is a functor, because it doesn't have the right shape. It makes sense to ask whether `Either[A, ?]` is a functor (for any fixed type `A`) or whether `Either[?, B]` is a functor (for any fixed type `B`). The answer is yes in both cases. Once we set our mind to a type constructor of the correct shape (`Either[A, ?]` or `Either[?, B]`), the functor instance for it has no bias. The bias is only in _our preference_ for the former, i.e. `Functor[Either[A, ?]]`, because that's the more common use case (as noted by @JordanMartinez), and also because it works better with type inference (namely with partial unification, which is also "right biased"). Other than that, `Functor[Either[A, ?]]` and `Functor[Either[?, B]]` are equally good functors. We don't provide both implicitly because it would confuse type inference in some situations. So we choose to provide the one that we _usually_ want and that one also plays better with type inference.

abhishek7:
@JordanMartinez @TomasMikula excellent explanations, thank you, I appreciate it! This makes much more sense now.

abhishek7:
@TomasMikula @JordanMartinez @puffnfresh @S11001001 Please take a look at the new PR when you can: https://github.com/scalaz/scalaz/pull/1822

PR #1397, opened 2017-06-01
Feature/map f

vmarquez:
This is the remerge of @justjoheinz's changes and also the suggestion by @TomasMikula for the method in MonadTrans.


core/src/main/scala/scalaz/MonadTrans.scala
```diff
@@ -15,6 +15,9 @@ trait MonadTrans[F[_[_], _]] {
   /** The [[scalaz.Monad]] implied by this transformer. */
   implicit def apply[G[_] : Monad]: Monad[F[G, ?]]
 
+  def mapF[G[_], A, B](fa: F[G, A])(f: A => G[B])(implicit M: Monad[G]): F[G, B] =
+    Monad[F[G, ?]].bind(fa)(a => liftM(f(a)))
+
```
TomasMikula: We have `liftM` and `mapF`, where both `-M` and `-F` refer to type parameter named `G`. This can be confusing :man_shrugging:

vmarquez: @TomasMikula I think this is true, it is confusing.  Seems we would break backwards compatibility either way if we change `mapF` to `mapM` or `liftM` to `liftF`?  I don't have strong feelings.  @xuwei-k ?

tonymorris:
FYI

`liftM` is a hangover from Haskell, in the days before `Applicative` was known about. The `M` there denotes that it is the `Monad` version of `fmap`, contrasted to the one on `Functor`.

You'll also find `liftA` in Haskell, starting at the time `Applicative` was found to be useful.

core/src/main/scala/scalaz/MonadTrans.scala
tonymorris: FYI

`liftM` is a hangover from Haskell, in the days before `Applicative` was known about. The `M` there denotes that it is the `Monad` version of `fmap`, contrasted to the one on `Functor`.

You'll also find `liftA` in Haskell, starting at the time `Applicative` was found to be useful.

TomasMikula: Maybe just switching the names of type parameters `F` and `G` in `MonadTrans` would help a little? And rename the type parameter to `liftM` to `M`?

PR #1400, opened 2017-06-07
invariant functor

vmarquez:
Invariant Functor.

### DIFF
--- a/base/src/main/scala/typeclass/InvariantFunctor.scala
+++ b/base/src/main/scala/typeclass/InvariantFunctor.scala
@@ -1,0 +1,10 @@
+package scalaz
+package typeclass
+
+trait InvariantFunctor[F[_]] {
+  def imap[A, B](ma: F[A])(f: A => B)(g: B => A): F[B]
+}
+
+object InvariantFunctor extends InvariantFunctorFunctions with InvariantFunctorSyntax {
+  def apply[F[_]](implicit F: InvariantFunctor[F]): InvariantFunctor[F] = F
+}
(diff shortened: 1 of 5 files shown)


tonymorris:
:+1:

jdegoes:
All `Functor` are `InvariantFunctor`. Separate pull request?

base/src/main/scala/typeclass/InvariantFunctor.scala
```diff
@@ -0,0 +1,10 @@
+package scalaz
+package typeclass
+
+trait InvariantFunctor[F[_]] {
+  def xmap[A, B](ma: F[A])(f: A => B)(g: B => A): F[B]
```
jdegoes: Since it's called `InvariantFunctor` and not `XFunctor` or `ExpFunctor`, I might suggest `imap` (or possibly just `map` depending on the policy for overloading).

The `(f: A => B)(g: B => A)` may be more usefully represented as `Iso[A, B]` / `<=>`, which composes better and may already be lying around for other purposes.

vmarquez: @jdegoes I agree, imap is a better name, will change. 

Once we have Iso I believe we should also have an additional method that takes an Iso.  This is how we have it in scalaz 7.

tonymorris: In general, `Iso` is a data type and `InvariantFunctor` is a type-class. I think we might (should?) agree to not conflate these two, going forward. That is, use a type-class excuslively so, and same for a data type.

PR #1403, opened 2017-06-12
Phantom functor

edmundnoble:
Phantom-variant functors. All `F[A]`s with lawful `Functor[F]` *and* `Contravariant[F]` have no occurrences of `A` in their definition, and thus the `A` can be freely changed to any other `B` without changing the data within; the `A` is *phantom-variant*. More succinctly, all values of type `F[A]` are also values of type `F[B]` for all `A` and `B`, which is witnessed by `pmap[A, B](fa: F[A]): F[B]`. See http://typelevel.org/blog/2016/02/04/variance-and-functors.html.

Note that this has no laws further than `Functor[F]` and `Contravariant[F]`, which interact with one another. I'm not sure where `pmap` should go otherwise, if this is too large an obstacle to include this type class.

### DIFF
--- a/base/src/main/scala/typeclass/Phantom.scala
+++ b/base/src/main/scala/typeclass/Phantom.scala
@@ -1,0 +1,10 @@
+package scalaz
+package typeclass
+
+trait Phantom[F[_]] {
+  def pmap[A, B](ma: F[A]): F[B]
+}
+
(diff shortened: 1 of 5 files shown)


tonymorris:
:+1:

Very nice.

aloiscochard:
TIL about `PhantomFunctor` ty, looks great :+1:

PR #1404, opened 2017-06-12
Remove macro recompile message

edmundnoble:
The macro recompile message is annoying and verbose, and I think we've sufficiently isolated the syntax macros by putting them in the meta project. This PR removes them.


vmarquez:
👍  I was just looking at these messages yesterday in annoyance wondering why we still had them.    Thanks @edmundnoble!

PR #1406, opened 2017-06-13
Match folder and package structure

edmundnoble:
Was the lack of package-named folders intentional? I believe it breaks tooling (IntelliJ especially) and I don't see the gain; this PR fixes it.


aloiscochard:
I guess the reason is as simple as I did it all with vi, so I never faced the issue! Definitely not intentional.

:+1:

PR #1411, opened 2017-06-17
Add mapConst, tapM and a syntax alias mechanism

edmundnoble:
`tapM` and `mapConst` are similar operations related to `const`. Both provide significant opportunities for optimization when overloaded, especially `tapM` for monads with no side effects (`Function0`, `Function1[C, ?]`, `Identity`), as well as being convenient.

As a side effect of this, I introduced a mechanism for syntactic aliases using phantom types with the same names as aliased methods. Open to any suggestions.

### DIFF
--- a/base/src/main/scala/data/ConstInstances.scala
+++ b/base/src/main/scala/data/ConstInstances.scala
@@ -7,20 +7,24 @@ import FoldableClass._
 import TraversableClass._
 
 trait ConstInstances {
-  implicit def traverse[R]: Traversable[Const[R, ?]] = new TraversableClass[Const[R, ?]] with FoldRight[Const[R, ?]] with Traverse[Const[R, ?]] {
-    def map[A, B](ma: Const[R, A])(f: A => B): Const[R, B] = ma.retag
+  implicit def traverse[R]: Traversable[Const[R, ?]] =
+    new TraversableClass[Const[R, ?]]
+      with FunctorClass.Map[Const[R, ?]]
+      with FoldRight[Const[R, ?]]
+      with Traverse[Const[R, ?]] {
+      def map[A, B](ma: Const[R, A])(f: A => B): Const[R, B] = ma.retag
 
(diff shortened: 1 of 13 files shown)


TomasMikula:
Regarding naming, Scalaz 7 has
 - [`Functor.as`](https://github.com/scalaz/scalaz/blob/fdbf1e3696b03929c39f3cf688460d2169116999/core/src/main/scala/scalaz/syntax/FunctorSyntax.scala#L22) which is the same as `mapConst` (up to by-name parameter); and
 - [`>>!`](https://github.com/scalaz/scalaz/blob/fdbf1e3696b03929c39f3cf688460d2169116999/core/src/main/scala/scalaz/syntax/BindSyntax.scala#L21) which is the same as `tapM`.

edmundnoble:
Which do you prefer? I thought `mapConst` was a bit more obvious if a bit more verbose. I also think that `>>=|` is more self-explanatory than `>>!`, by analogy to `>|`.

TomasMikula:
Agree that `mapConst` is more descriptive. Just pointing out that there's an existing name.

>  I also think that `>>=|` is more self-explanatory than `>>!`, by analogy to `>|`.

But `tapM` isn't really analogous to `mapConst`, is it? `mapConst` ignores the _old_ value, while `tapM` ignores the _new_ value.

Also, I don't understand how "tap" is descriptive of what is going on. Maybe some idiosyncratic meaning of the word "tap"?

edmundnoble:
I was going for `tap` by analogy to `unsafeTap`, `tapM` is a sort of "`safeTap`". That's a good point though, it's not exactly analogous to `mapConst`; it's more `flatMap(a => mapConst(f(a))(a))`. What do you think of `>>=!` as a symbolic alias? I am a little apprehensive about `>>!` because it doesn't seem very intuitive, if `!` denotes something being done purely for the effects it's odd because `<<` is already a sort of "`>>` for the effects".

TomasMikula:
Another option for `tapM`: `<<|` (in analogy to `tap` being `<|`). But I don't find it or any of `>>=|`, `>>!`, `>>=!` self-explanatory. And neither `>|`, for that matter. Consequently, I don't care all that much.

PR #1412, opened 2017-06-17
Add IdentitySyntax

edmundnoble:
Couple of ops for all values, tastefully selected from Scalaz 7: `unsafeTap` and `??` are gone, replaced `doWhile` and `whileDo` with `unfold` which is easier to understand, renamed `squared` to `diagonal` for less ambiguity with math.


TomasMikula:
If monoid's  identity element is now called `empty`, then it should be `matchOrEmpty` instead of `matchOrZero`.

Also, the "match" part doesn't quite fit, IMO. What about `imageOrEmpty`?

__Aside:__ Do people find `empty` a better name than `zero`? I can see that `empty` fits better with `append`, but if we are going to change names, wouldn't terminology that doesn't suggest a specific use be better? Such as `compose` and `identity`, or `combine` and `neutral`?

edmundnoble:
What do you think about `applyOrEmpty`? Also I would prefer `combine` and `zero`, personally.

TomasMikula:
`a.applyOrEmpty(f)` would not be consistent with the meaning of `a.apply(f)`.

base/src/main/scala/data/IdentitySyntax.scala
```diff
@@ -0,0 +1,42 @@
...
+    * @return the result of pf(self) if defined, otherwise the the empty element of `B`.
+    */
+  def applyOrEmpty[B: Monoid](pf: PartialFunction[A, B]): B =
+    pf.lift(self) match {
+      case None    => Monoid[B].empty
+      case Some(x) => x
+    }
+
+  /** Repeatedly apply `f`, seeded with `self`, until a `B` is returned. */
+  def unfold[B](f: A => A Either B): B = {
```
jdegoes: Should be `\/`.

base/src/main/scala/data/IdentitySyntax.scala
```diff
@@ -0,0 +1,42 @@
...
+  def diagonal: (A, A) =
+    (self, self)
+
+  /** Returns `Unit`. */
+  def ignore: Unit = ()
+
+  /**
+    * @return the result of pf(self) if defined, otherwise the the empty element of `B`.
+    */
+  def applyOrEmpty[B: Monoid](pf: PartialFunction[A, B]): B =
```
jdegoes: Do you think this would be better added as syntax on `PartialFunction`?

base/src/main/scala/data/IdentitySyntax.scala
```diff
@@ -0,0 +1,42 @@
+package scalaz
+package data
+
+import annotation.tailrec
+import scalaz.typeclass.{Applicative, Monoid}
+import language.implicitConversions
+
+final class IdOps[A](val self: A) extends AnyVal {
+  /** Applies the provided function to `self`. The Thrush combinator. */
+  def |>[B](f: A => B): B =
```
jdegoes: If we use this, as opposed to, say, `<#>` or `&`, we should probably also think about `<|` where appropriate (not here).

base/src/main/scala/typeclass/ApplicativeSyntax.scala
```diff
@@ -10,5 +10,13 @@ trait ApplicativeSyntax {
 object ApplicativeSyntax {
   class OpsA[A](a: A) {
     def pure[F[_]](implicit F: Applicative[F]): F[A] = F.pure(a)
+
+    /**
+      * If the provided partial function is defined for `self` run this,
+      * otherwise lift `self` into `F` with the provided [[scalaz.typeclass.Applicative]].
+      */
+    def visit[F[_]](p: PartialFunction[A, F[A]])(implicit F: Applicative[F]): F[A] =
```
jdegoes: Same question as above: would this be better as syntax on `PartialFunction`?

edmundnoble:
@jdegoes I addressed your and @TomasMikula's feedback by just removing `applyOrEmpty` and using `\/`.

base/src/main/scala/data/IdentitySyntax.scala
jdegoes: It would be defined on `FunctionOps`.

base/src/main/scala/data/IdentitySyntax.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  /** Repeatedly apply `f`, seeded with `self`, until a `B` is returned. */
+  def unfold[B](f: A => A \/ B): B = {
+    @tailrec
+    def loop(value: A): B = f(value) match {
+      case \/-(r) => r
+      case -\/(l) => loop(l)
+    }
+    loop(self)
+  }
```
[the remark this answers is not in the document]
jdegoes: `cycle`, `recycle`, or `repeat` maybe.

base/src/main/scala/data/IdentitySyntax.scala
hrhino: Oh, `def <|(a: A): B = self.apply(a)`, not `tap` (like in scalaz 7)? That makes more sense.

PR #1416, opened 2017-06-26
Add FunctionH

edmundnoble:
Just a shot in the dark: I named this `FunctionH` instead of NaturalTransformation because technically speaking, not all natural transformations are `FunctionH`s, even though all `FunctionH`s are natural transformations. Bikeshedding very welcome. I also added a bit of sugar for `FunctionH` creation, courtesy of a trick @alexknvl thought of; if it's considered unnecessary by most given the `Lambda` syntax (despite that it doesn't work in Intellij) I wouldn't be averse to dropping it.

### DIFF
--- a/base/src/main/scala/scalaz/data/FunctionH.scala
+++ b/base/src/main/scala/scalaz/data/FunctionH.scala
@@ -1,0 +1,54 @@
+package scalaz
+package data
+
+import language.implicitConversions
+
+/** A universally quantified function, usually written as `F ~> G`,
+  * for symmetry with `A => B`.
+  *
+  * All `FunctionH[F, G]` values, with `F` and `G` functors, form functor
+  * homomorphisms (natural transformations) from `F` to `G` by a free theorem.
+  */
+trait FunctionH[-F[_], +G[_]] {
+  self =>
+  def apply[A](fa: F[A]): G[A]
+
+  def compose[E[_]](f: E ~> F): E ~> G = λ[E ~> G](
+    ea => self(f(ea))
+  )
+
+  def andThen[H[_]](f: G ~> H): F ~> H =
+    f compose self
+}
+
+trait FunctionHFunctions {
+  /** A universally quantified identity function */
+  def id[F[_]] =
+    λ[F ~> F](fa => fa)
+
+  /** Reify a `NaturalTransformation`. */
+  implicit def natToFunction[F[_], G[_], A](t: F ~> G): F[A] => G[A] = fa => t(fa)
+
+  private object MkFunctionH {
+    val unsafe: MkFunctionH = new MkFunctionH
+  }
+
+  /**
+    * Use to create a FunctionH from a quantified method:
+    * `FunctionH.mk[List, Option](_.headOption)`
+    */
+  def mk: MkFunctionH = MkFunctionH.unsafe
+}
+
+// credit to Alex Konovalov for this trick: undefined type members
+// with unstable prefixes, to encode universal quantification
+private[data] final class MkFunctionH(val dummy: Boolean = true) extends AnyVal {
+  type T
+
+  def apply[F[_], G[_]](f: F[T] => G[T]): FunctionH[F, G] = new FunctionH[F, G] {
+    def apply[A](fa: F[A]): G[A] = f(fa.asInstanceOf[F[T]]).asInstanceOf[G[A]]
+  }
+}
+
+object FunctionH extends FunctionHFunctions
+
--- a/base/src/main/scala/scalaz/Prelude.scala
+++ b/base/src/main/scala/scalaz/Prelude.scala
@@ -26,6 +26,7 @@ trait Prelude  extends data.DisjunctionFunctions
   type Profunctor[F[_,_]] = typeclass.Profunctor[F]
   type Category[=>:[_,_]] = typeclass.Category[=>:]
   type InvariantFunctor[F[_]] = typeclass.InvariantFunctor[F]
+  type ~>[F[_], G[_]] = FunctionH[F, G]
 
   def Applicative[F[_]](implicit F: Applicative[F]): Applicative[F] = F
   def Apply[F[_]](implicit F: Apply[F]): Apply[F] = F
@@ -73,7 +74,7 @@ trait Prelude  extends data.DisjunctionFunctions
 
   implicit def PstrongOps[P[_,_], A, B](pab: P[A, B])(implicit P: Strong[P]): StrongSyntax.Ops[P, A, B] =
     new StrongSyntax.Ops(pab)
-   
+
   //InvariantFunctorSyntax
   implicit def InvariantFunctorOps[F[_], A](fa: F[A])(implicit F: InvariantFunctor[F]): InvariantFunctorSyntax.Ops[F, A] =
     new InvariantFunctorSyntax.Ops(fa)
--- a/base/src/main/scala/scalaz/package.scala
+++ b/base/src/main/scala/scalaz/package.scala
@@ -1,2 +1,8 @@
-package object scalaz extends scalaz.BaseHierarchy
+import scalaz.data.FunctionH
+
+package object scalaz extends scalaz.BaseHierarchy {
+  /** A [[scalaz.data.FunctionH]][F, G]. */
+  type ~>[-F[_], +G[_]] = FunctionH[F, G]
+
+}
 


TomasMikula:
👍 for the name `FunctionH`.

The `MkFunctionH` trick seems unlikely to cause any trouble, so I would keep it even though it doesn't seem 100% safe (?).

That said, `FunctionH` is still just a special case of a polymorphic _value_. If we could get the same level of syntactic convenience and performance with modeling `FunctionH` as a polymorphic value, I would try to go that way. I started work on syntax in non/kind-projector#54. Alternatively, the same trick as in this PR could be used to create `Forall`s.

base/src/main/scala/scalaz/data/FunctionH.scala
```diff
@@ -0,0 +1,54 @@
...
+  /**
+    * Use to create a FunctionH from a quantified method:
+    * `FunctionH.mk[List, Option](_.headOption)`
+    */
+  def mk: MkFunctionH = MkFunctionH.unsafe
+}
+
+// credit to Alex Konovalov for this trick: undefined type members
+// with unstable prefixes, to encode universal quantification
+private[data] final class MkFunctionH(val dummy: Boolean = true) extends AnyVal {
```
aloiscochard: Is this default parameter really required? Seems like stale code to me.

base/src/main/scala/scalaz/data/FunctionH.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  private object MkFunctionH {
+    val unsafe: MkFunctionH = new MkFunctionH
+  }
+
+  /**
+    * Use to create a FunctionH from a quantified method:
+    * `FunctionH.mk[List, Option](_.headOption)`
+    */
+  def mk: MkFunctionH = MkFunctionH.unsafe
```
aloiscochard: Could it be simply named `apply`? so the syntax would be `FunctionH[A, B](...)`.

base/src/main/scala/scalaz/data/FunctionH.scala
TomasMikula: An `AnyVal` has to have a parameter.

base/src/main/scala/scalaz/data/FunctionH.scala
TomasMikula: I don't think Scala will expand `FunctionH[A, B](...)` to `FunctionH.apply.apply[A, B](...)`

base/src/main/scala/scalaz/data/FunctionH.scala
aloiscochard: ooooh, okay so that's the trick then.

I would prefer a `Unit` than a `Boolean` here but it is really bikeshedding I guess.

aloiscochard:
What would be an unsafe use of the syntax? I could not work one out.

base/src/main/scala/scalaz/data/FunctionH.scala
aloiscochard: Actually re-reading the code it seems like the doc is wrong, the actual syntax is in fact `mk[A, B](...)`, as the method is directly at the root of the trait, or am I missing something here?

Then we could simply find a better name than `mk` without wrapping it in an object and get a syntax similar to what I was saying.

base/src/main/scala/scalaz/data/FunctionH.scala
aloiscochard: Using `Unit` would have the advantage to make it clearer that this value have no real uses.

TomasMikula: I think there's some scalac bug with `Unit` in this case 😄

TomasMikula:
I think if you do

```scala
val maker = FunctionH.mk
```

now you have access to `maker.T`, could create a non-parametric instance of `F[T] => G[T]` and pass it to `maker.apply`.

Haven't tried it, but I base it on a suspicion that the fact that `MkFunctionH` is `private[data]` doesn't really matter.

base/src/main/scala/scalaz/data/FunctionH.scala
TomasMikula: https://github.com/scala/bug/issues/9240

edmundnoble:
@TomasMikula You are correct; if you can make two FunctionH's to mention the same `T` "skolem" from the same `MkFunctionH` then you get bugs.

TomasMikula:
@edmundnoble I had this example in mind:

```scala
val maker = FunctionH.mk
val f: Id ~> Const[maker.T, ?] = maker.apply[Id, Const[maker.T, ?](a => Const(a))
f[Int](5) // BOOM!
```

However, this example is prevented if we prevent `F` and `G` from mentioning `maker.T`, which can be achieved if `F`, `G` have to be specified before `MkFunctionH` even exists, which in turn is the case if `MkFunctionH` is parameterized by them.

Can you come up with a failing example even after this proposed change (i.e. after moving `F` and `G` parameters from `MkFunctionH.apply` to `MkFunctionH`)? I don't see how

> two `FunctionH`'s to mention the same `T` "skolem" from the same `MkFunctionH`

causes a problem.

edmundnoble:
Ah, that sounds like a good fix. What I was saying amounted to the same thing as your example.

jdegoes:
@edmundnoble Sorry this has taken so long to get around to. Can you please fix merge conflicts?

@TomasMikula Do you have any more comments on the safety of the helper syntax for this?

TomasMikula:
If the suggestion is incorporated, I'm convinced the syntax is safe, and I have used the same trick in `Forall`.

But I would like us to consider whether to have this dedicated representation of polymorphic functions, or whether to represent polymorphic functions as polymorphic values (`Forall`):

```scala
type ~>[F[_], G[_]] = ∀[λ[α => F[α] => G[α]]]
```

There are tradeoffs to both approaches. I will try to summarize the ones I'm aware of.

### Creation

The `mk` syntax trick in this PR and in `Forall` is basically the same.

Additionally, `λ` syntax from `kind-projector` can be used to create `FunctionH`s from this PR. However, that syntax doesn't generalize to [`BiNaturalTransformation`](https://github.com/scalaz/scalaz/blob/b2f656aaad644f2ad1d84cef7be7e70ffd2f8ba5/core/src/main/scala/scalaz/NaturalTransformation.scala#L57) and other variations of polymorphic functions.

None of these syntaxes are applicable when the polymorphic implementation needs to refer to the type argument. If that is the case, we have to fall back to `new (F ~> G) { def apply[A](fa: F[A]) = ??? }`.

[P∀scal](https://github.com/TomasMikula/pascal) provides syntax for polymorphic values encoded as objects with one polymorphic nullary method. It works for any number of type parameters of any kinds, and allows to refer to type arguments from the body.

P∀scal syntax applies directly to `Forall` from scalaz7. In scalaz8, encoding of `Forall` is different (internally, `type Forall[F] = F[Any]`). However, P∀scal can be used to create [`Forall.Prototype`](https://github.com/scalaz/scalaz/blob/7428dac3328d8f20f38644c5b5f3769828892d69/base/shared/src/main/scala/scalaz/data/Forall.scala#L13-L15), which in turn [can be used to create `Forall`](https://github.com/scalaz/scalaz/blob/7428dac3328d8f20f38644c5b5f3769828892d69/base/shared/src/main/scala/scalaz/data/Forall.scala#L19).

This could be considered as a slight advantage of the `Forall`-encoded polymorhic function.

### Application

Given `φ: F ~> G` from this PR, applying it to a value `fa: F[A]` is as easy as

```scala
φ.apply[A](fa)
```

which can be shortened to just

```scala
φ(fa)
```

To apply the `Forall`-encoded natural transformation `φ`, the full spelling is

```scala
φ.apply[A].apply(fa)
```

The first `apply` is instantiation of the universally quantified function to a specific type, `A`, and amounts to no more than `asInstanceOf`. The second `apply` is then just invoking the resulting `Function1`.

The application can be shortened to

```scala
φ[A](fa)
```

or

```scala
φ.apply(fa)
```

but we can't get to `φ(fa)` as with `FunctionH` from this PR. We could, however, provide special syntax, such as `φ $ fa`.

This is a slight downside of the `Forall`-encoded polymorphic function.

### Subtyping

For both `FunctionH` from this PR and `Forall`-encoded natural transformation, one can write a function

```scala
(F ~> G) => (F[A] => G[A])
```
for any type `A`.

With `Forall`-encoded natural transformation, however, we get something stronger:

```scala
(F ~> G) <~< (F[A] => G[A])
```
for any type `A`.

Also consider binatural transformation encoded as

```scala
type ~~>[F[_, _], G[_, _]] = ∀∀[λ[(α, β) => F[α, β] => G[α, β]]]
```

We can have, for example

```scala
(F ~~> G) <~< (F[A, ?] ~> G[A, ?])
```

that is, we can get a natural transformation from binatural transformation by fixing one type parameter, and this operation is a no-op.

Or we can have

```scala
(F ~~> G) === ∀[λ[α => F[α, ?] ~> G[α, ?]]]
          === ∀[λ[β => F[?, β] ~> G[?, β]]]
```

i.e. view a binatural transformation as a (further) universally quantified natural transformation, and switching between these views is a no-op.

This is why I would lean towards `Forall`-encoded polymorphic functions.

jdegoes:
@TomasMikula I agree. So what do we need in a new PR?

TomasMikula:
I sneaked in a [type alias](https://github.com/scalaz/scalaz/blob/4318664f571f97b98b049f9ad8e634281b24dee4/base/shared/src/main/scala/scalaz/data/package.scala#L18-L19) in the type-aligned commit. 😈 😉 

Dedicated syntax (e.g. for composition, application, application to universally quantified value), as well as a companion object, are missing.

edmundnoble:
@TomasMikula Sneaky bugger 😝 I'm agreed then; I can just rewrite this PR into the utilities that would be required on top of Forall-based FunctionH. My only comment is special application syntax does seem like a fairly large hurdle of non-generalization, but I think passing the type argument is mostly an acceptable bit of syntax especially given Λ in Pascal. Either way it's miles better than kind projector right now.

TomasMikula:
Well, I just needed _some_ sort of natural transformation there. If we agree here otherwise, we can change the one used in type-aligned data types.

There's also the possibility for `FunctionH` being an abstract type, privately a type alias
```scala
type FunctionH[F[_], G[_]] = ∀[λ[α => F[α] => G[α]]]
```
that wouldn't be visible to the user. (Though there are performance concerns with such approach, see #1450). Then we could provide application syntax as convenient as with the specialized `FunctionH` in this PR, since it would not conflict with the syntax for `Forall`. In addition, we would provide a `Leibniz` instance `FunctionH[F, G] === ∀[λ[α => F[α] => G[α]]]` for all `F[_]`, `G[_]`, to get the benefits of interaction of `Forall` with subtyping.

Atry:
`FunctionH ` requires https://github.com/scala/bug/issues/9916 or https://github.com/scala/bug/issues/9897 , or the syntax would be too trivial.

edmundnoble:
Closing in favor of a proper PR here.

PR #1417, opened 2017-06-28
Unboxed Forall

TomasMikula:
Define `Forall[F[_]]` as an abstract type, privately implemented as

```scala
type Forall[F[_]] = F[Any]
```

A type alias is provided

```scala
type ∀[F[_]] = Forall[F]
```

### Creating instances

Since `Forall[F]` is represented as `F[Any]`, care needs to be taken so that only truly parametric instances can be created. Two ways to create instances are provided:

 1. From a "prototype", which is a Scalaz7-style universally quantified value:

    ```scala
    def from[F[_]](p: Prototype[F]): Forall[F]

    trait Prototype[F[_]] {
      def apply[A]: F[A]
    }
    ```

    Prospectively, there could be convenient syntax for creating `Prototype`s via non/kind-projector#54.

    Note that creation via a prototype requires an allocation of a short-lived object (the prototype), but the resulting `Forall[F]` value is unboxed. (Perhaps that one allocation could be eliminated by the optimizer?)

 2. Via the same trick (due to @alexknvl) as in #1416:

    ```scala
    val l1: Forall[List] = Forall.of[List](Nil)

    // or

    val l2: Forall[List] = Forall.mk[Forall[List]].from(Nil)
    ```

    The seemingly longer `mk[...].from(...)` syntax becomes more succinct in some cases:

    ```scala
    def listSemigroup[A]: Semigroup[List[A]] = ???

    type Plus[F[_]] = Forall[λ[A => Semigroup[F[A]]]]

    val listPlus: Plus[List] = Forall.of[λ[A => Semigroup[F[A]]]](listSemigroup)

    // vs

    val listPlus: Plus[List] = Forall.mk[Plus[List]].from(listSemigroup)
    ```

### Using instances

From user's perspective `Forall[F]` is completely abstract, but values are enriched with `apply[A]` method:

```scala
class ForallOps[F[_]](val a: Forall[F]) extends AnyVal {
  def apply[A]: F[A] = ???
}
```

One then uses this method to get an instance specialized for a specific type:

```scala
listPlus[Int].append(List(1, 2), List(3, 4)) // List(1, 2, 3, 4)
```

---

See also `ForallUsage.scala` on how `Forall` can be used to represent `NaturalTransformation`.

### DIFF
--- a/base/shared/src/main/scala/scalaz/Prelude.scala
+++ b/base/shared/src/main/scala/scalaz/Prelude.scala
@@ -6,6 +6,7 @@ import data._
 import scala.language.implicitConversions
 
 trait Prelude  extends data.DisjunctionFunctions
+                  with data.ForallSyntax
                   with data.IdentityTypes
                   with data.MaybeFunctions
                   with typeclass.BindFunctions
@@ -97,6 +98,11 @@ trait Prelude  extends data.DisjunctionFunctions
   type Identity[A] = data.Identity[A]
   type Maybe[A] = data.Maybe[A]
   type Forget[A, B, C] = data.Forget[A, B, C]
+
+  val forall = data.forall
+  type Forall[F[_]] = forall.Forall[F]
+  type ∀[F[_]] = Forall[F]
+
 }
 
 object Prelude extends Prelude
--- a/base/shared/src/main/scala/scalaz/Prelude.scala
+++ b/base/shared/src/main/scala/scalaz/Prelude.scala
@@ -7,6 +7,7 @@ import scala.language.implicitConversions
 
 trait Prelude  extends data.DisjunctionFunctions
                   with data.ForallSyntax
+                  with data.Forall2Syntax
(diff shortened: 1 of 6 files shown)


aloiscochard:
This is beautiful work, I :heart: it.

What about adding the `$` syntax directly in `base`? seems pretty useful to me.

Should we add this in default `Prelude`? I think yes.

TomasMikula:
In this PR, `~>` is used just as an example. We can consider `$` syntax when adding natural transformation. Though there might be confusion about preference, since `a $ b $ c` in Scala is `(a $ b) $ c`, whereas in Haskell it is `a $ (b $ c)`. I used `$` because `Forall` already has an `apply` method.

Yeah, I will add it to `Prelude`.

TomasMikula:
Added to Prelude. Maybe there's a better organization of the code?

aloiscochard:
Looks great!

PR #1422, opened 2017-06-29
Series8/feature/forget fix

vmarquez:


aloiscochard:
LGTM, I think it's worth rebasing and squashing all commits into a single one before merging (that will remove the merge commits).

PR #1424, opened 2017-07-17
comonad and cobind

vmarquez:


base/src/main/scala/scalaz/Prelude.scala
```diff
@@ -39,6 +41,8 @@ trait Prelude  extends data.DisjunctionFunctions
   def Strong[P[_,_]](implicit P: Strong[P]): Strong[P] = P
   def Category[=>:[_,_]](implicit P: Category[=>:]): Category[=>:] = P 
   def InvariantFunctor[F[_]](implicit F: InvariantFunctor[F]): InvariantFunctor[F] = F
+  def Comonad[F[_]](implicit F: Comonad[F]): Comonad[F] = F
+  def Cobind[F[_]](implicit F: Cobind[F]): Cobind[F] = F
```
aloiscochard: It's not only part of this PR, but as the other code is also from you (https://github.com/vmarquez/scalaz/commit/9f7b59c7119a6228ef99bb4c99a361a9737bbca1) I take the initiative to mention this here.

Could you please order the added `type`/`def` in alphebetical order? don't wanna sound spreading my OCD, but I really think that keeping alphabetical order here will make maintainance easier.

base/src/main/scala/scalaz/Prelude.scala
```diff
@@ -78,6 +82,13 @@ trait Prelude  extends data.DisjunctionFunctions
   implicit def InvariantFunctorOps[F[_], A](fa: F[A])(implicit F: InvariantFunctor[F]): InvariantFunctorSyntax.Ops[F, A] =
     new InvariantFunctorSyntax.Ops(fa)
   
+  implicit def CobindOps[F[_], A](fa: F[A])(implicit F: Cobind[F]): CobindSyntax.Ops[F, A] =
+    new CobindSyntax.Ops(fa)
+
+  implicit def ComonadOps[F[_], A](fa: F[A])(implicit F: Comonad[F]): ComonadSyntax.Ops[F, A] =
+    new ComonadSyntax.Ops(fa)
+
```
aloiscochard: Same here, about alphabetical ordering.

vmarquez: @aloiscochard apologies, will do!

Issue #1292, opened 2016-11-11
which optimize option use in Scala 2.12


xuwei-k:
http://scala-lang.org/news/2.12.0#new-optimizer

```
$ scala -version
Scala code runner version 2.12.0 -- Copyright 2002-2016, LAMP/EPFL and Lightbend, Inc.
$ scala -opt:help
Enable optimizations
  unreachable-code     Eliminate unreachable code, exception handlers guarding no instructions, redundant metadata (debug information, line numbers).
  simplify-jumps       Simplify branching instructions, eliminate unnecessary ones.
  compact-locals       Eliminate empty slots in the sequence of local variables.
  copy-propagation     Eliminate redundant local variables and unused values (including closures). Enables unreachable-code.
  redundant-casts      Eliminate redundant casts using a type propagation analysis.
  box-unbox            Eliminate box-unbox pairs within the same method (also tuples, xRefs, value class instances). Enables unreachable-code.
  nullness-tracking    Track nullness / non-nullness of local variables and apply optimizations.
  closure-invocations  Rewrite closure invocations to the implementation method.
  inline-project       Inline only methods defined in the files being compiled. Enables unreachable-code.
  inline-global        Inline methods from any source, including classfiles on the compile classpath. Enables unreachable-code.
  l:none               Disable optimizations. Takes precedence: `-opt:l:none,+box-unbox` / `-opt:l:none -opt:box-unbox` don't enable box-unbox.
  l:default            Enable default optimizations: unreachable-code.
  l:method             Enable intra-method optimizations: unreachable-code,simplify-jumps,compact-locals,copy-propagation,redundant-casts,box-unbox,nullness-tracking,closure-invocations.
  l:project            Enable cross-method optimizations within the current project: l:method,inline-project.
  l:classpath          Enable cross-method optimizations across the entire classpath: l:project,inline-global.

S11001001:
Hmm, `l:method` seems like a winner.  Only one I'm not sure about there is `compact-locals`.

Until scala.meta arrives and gives us proper inlining (and a proper way to offer it to users), not sure that's a hole worth going down.  If we want to try anyway, `l:project` and a sprinkling of `@inline` in "obvious" places (to be replaced with meta's `inline` keyword later).

S11001001:
Merged #1338.

PR #1426, opened 2017-07-18
switch to scala 2.12 and add scaffolding for scala.js

jdegoes:
A bunch of prep work to ease the painful situation I'm in presently with scalaz.effect.

 - [X] Add Scala.js compiler plug-in
 - [X] Reorganize for scala.js support
 - [X] Modify build to build JS versions
 - [X] Add sbt script for reproducible builds
 - [X] Upgrade Scala to 2.12.x as default
 - [X] Add optimization flags that have useful performance benefits


.gitignore
```diff
@@ -1,2 +1,3 @@
 tags
 target
+.DS_Store
```
aloiscochard: That should not be here, you can include such OS specific exclude in your global gitignore, ty

project/ScalazBuild.scala
```diff
@@ -4,13 +4,12 @@ import Keys._
...
-  def module(prjName: String) = Project(
-    id = prjName,
-    base = file(prjName)).settings(
+  def stdSettings(prjName: String) = Seq(
     name := s"scalaz-$prjName",
-    scalaVersion := "2.11.11",
+    scalaVersion := "2.12.2",
     scalacOptions ++= Seq("-feature","-deprecation", "-Xlint", "-language:higherKinds",
-                          "-Ydelambdafy:method", "-Ypartial-unification", "-target:jvm-1.8"),
+                          "-Ydelambdafy:method", "-Ypartial-unification", "-target:jvm-1.8", "-opt:l:project",
```
aloiscochard: I find suspicious that `-opt:l:project` is required here, what is the impact on performance?

Inlining is known to have cause issues in the best in users library/projects.

.gitignore
jdegoes: Sounds good. Should I remove `tags` as well, while I'm in there?

project/ScalazBuild.scala
xuwei-k: https://github.com/scalaz/scalaz/issues/1292

jdegoes: It's definitely measurable. However, I can manually inline the critical section if necessary, which will achieve most of the gains. The optimizer will only inline methods so annotated (`@inline`), and it looks more carefully engineered than 2.11.x. 

Let me know how to handle it.

build.sbt
```diff
@@ -1,28 +1,57 @@
+import org.scalajs.sbtplugin.ScalaJSPlugin
+import org.scalajs.sbtplugin.ScalaJSPlugin.autoImport._
```
xuwei-k: maybe these imports unnecessary

.gitignore
aloiscochard: Yep indeed! ty

project/ScalazBuild.scala
aloiscochard: It would interesting to have opinion of @retronym on such thing, as he knows how such a library is crafted and he is probably well aware of the current state of the inliner.

I don't feel qualified enough to make a decision here, I'm just pushing on the safe side due to bad memories about inlining... But it would be a shame to miss such improvement if they are now properly implemented... might be worth trying initially... and can go back later...

Thanks Kenji for the pointer!

aloiscochard:
Great effort BTW, I wish I did start this way.

project/ScalazBuild.scala
jdegoes: I left the inline pragma for now. However, in the future, if we decide to remove it, I can manually inline critical section code.

build.sbt
jdegoes: I deleted them, turns out they were not required.

.gitignore
jdegoes: I deleted everything but `target`.

jdegoes:
@aloiscochard @xuwei-k All right, this is ready for merge.

jdegoes:
@aloiscochard @xuwei-k Anything else I should change before this gets merged?

aloiscochard:
LGTM! great stuff :)

jdegoes:
@aloiscochard Would you mind merging? I don't have permissions. Thanks!

project/ScalazBuild.scala
jbgi: it looks like `opt:l:project` is now deprecated: https://github.com/scala/scala/pull/5964

jdegoes:
@jbgi Great. So now you have to inline in two places: the code, and the build file. 🙄 😆 

I think in my next PR I'll just manually inline...

PR #1427, opened 2017-07-24
Add `Ap` data type: lowers Apply to Semigroup, Applicative to Monoid

puffnfresh:


core/src/main/scala/scalaz/Ap.scala
```diff
@@ -0,0 +1,41 @@
+package scalaz
+
+/** Derive a Semigroup or Monoid instance from an Apply or Applicative.
+ *
+ * https://hackage.haskell.org/package/reducers-3.12.1/docs/Data-Semigroup-Applicative.html#t:Ap
+ */
+case class Ap[F[_], A](f: F[A])
```
gwils: Why not `final case class`?
I'm not sure whether scalaz has a policy on this or not.

tests/src/test/scala/scalaz/ApTest.scala
```diff
@@ -0,0 +1,8 @@
+package scalaz
+
```
xuwei-k: maybe missing import

```scala
import scala.std.list._
```

```
[error] /home/travis/build/scalaz/scalaz/tests/src/test/scala/scalaz/ApTest.scala:5: could not find implicit value for parameter F: scalaz.Semigroup[scalaz.Ap[List,A]]
[error]     def semigroup[A: Monoid] = Semigroup[Ap[List, A]]
[error]                                         ^
```

puffnfresh:
Thanks for the additional tests @xuwei-k :+1:

PR #1429, opened 2017-07-27
Add digitsFrom{Int, Long, BigInt}: extract digits of a number,

jbgi:
also use direct array lookup to speedup some existing Digit methods.


edmundnoble:
Nice.

PR #1430, opened 2017-08-01
adding missing band instance for std. map

vmarquez:
Thought this might be useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ matrix:
   - scala: 2.11.11
     jdk: oraclejdk8
     sudo: required
+    dist: trusty
     before_install:
     - curl https://raw.githubusercontent.com/scala-native/scala-native/v0.3.1/bin/travis_setup.sh | bash -
     script:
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/scalaz/MapTest.scala
```diff
@@ -1010,6 +1010,7 @@ object MapTest extends SpecLite {
   checkAll(zip.laws[IntMap])
   checkAll(bifoldable.laws[==>>])
   checkAll(FoldableTests.anyAndAllLazy[IntMap])
+  checkAll(band.laws[Int ==>> ISet[Int]])
```
xuwei-k: already exists here

https://github.com/scalaz/scalaz/blob/cd13e74e346b4da23d78bf109d5c3bdc77b0ac97/tests/src/test/scala/scalaz/MapTest.scala#L974

vmarquez:
@xuwei-k oops I will remove that one.

puffnfresh:
Isn't it only a Band when the value is one too? ISet has a Band, so passes test but looks like "String ==>> String" type checks but shouldn't have a Band instance right?

vmarquez:
You are right @puffnfresh, I have fixed that.

vmarquez:
@xuwei-k not sure why the build is failing, do I need to fix something else?

xuwei-k:
fixed in https://github.com/scalaz/scalaz/commit/c3f2fd331ddc46eb700013f292d7e773efd93179
please rebase and squash commits

vmarquez:
@xuwei-k thanks! I have merged and rebased again.

xuwei-k:
merged https://github.com/scalaz/scalaz/commit/7a449affd22daeef953e9a2c11e3579fb19e13c4

PR #1432, opened 2017-08-10
Delete MonadPartialOrder

puffnfresh:


tonymorris:
:+1000000:

puffnfresh:
#1371

TomasMikula:
> Writer (Any,Any) to itself can flip the order

Is this really a monad homomorphism?

> keep one or the other and duplicate,
> reset to False

These are not injective.

Still, there _can_ be "a whole boatload" of (even injective) monad homomorphisms from `F` to `G`. Just like there is a whole boatload of Monoid[Int]. There, we just choose one as canonical, and provide others via newtyping.

A more practical problem is possibly ambiguous implicits: multiple ways to implicitly derive the same instance (because of diamond or associativity of the transitive property). This, however, is a technical problem, which doesn't disqualify MonadPartialOrder as a concept.

aaronvargo:
An injective homomorphism is called a monomorphism.

> Still, there can be "a whole boatload" of (even injective) monad homomorphisms from F to G. Just like there is a whole boatload of Monoid[Int]. There, we just choose one as canonical, and provide others via newtyping.

I think the point is that if there's "a whole boatload" then it isn't a partial order. If it's a typeclass and instances are ensured to be unique, then maybe calling it a partial order is fine, but given that scala allows explicit instances it'd be better to name it MonadMonomorphism.

It's (I think) possible to get transitivity by defining a reflexive instance, and then defining further instances via implicits of the form `[F] MonadPartialOrder[F, Foo] => MonadPartialOrder[F, Bar]` in a scato-like hierarchy. Of course, it's essential that these instances all be coherent.

The main question to ask is whether it's worth having this as a typeclass, or if it should just be a regular data type.

ekmett:
> Writer (Any,Any) to itself can flip the order

Is this really a monad homomorphism?

A monad homomorphism is merely required to satisfy

```haskell
f (return a) = return a
f (m >>= k) = f m >>= f . k
```

So, yes. The operations respecting the meaning of tell, etc, is not required, nor in fact is it satisfied by any of the other instances that do things like turn asks into gets or tells into state modifications. Turning `tell (a,b)` into `tell (b, a)` is perfectly admissable.

Even requiring a monomorphism isn't enough, The flip example is a perfectly good monomorphism, and I can similarly take a Reader[Int, A] and build a whole family of monomorphisms that modify the input by scrambling the integers however I want. If I have a lens from X to Y I can run a State[Y,A] action in a State[X,A] monad, but lenses are not unique.

In general, Monomorphisms between monads are not unique, the category isn't thin, so this doesn't belong as a class. Moreover, the way the class is defined at present there is no guarantee that composition of two such embeddings yield the same result.

You can build these sorts of things up as some kinda data type that can be composed, like we do with lenses, and that is a perfectly cromulent way to work with monad (homo|mono|epi)morphisms, and/or their split variants that allow you to retract the embeddings or quotient a given monad through a smaller model, but its just not a thing that makes sense mathematically as a class.

ekmett:
Looks good to me.

aaronvargo:
> In general, Monomorphisms between monads are not unique, the category isn't thin, so this doesn't belong as a class

Monoids aren't unique either. There be many instances of Reader Int ~> Reader Int, but there's still a reasonable instance to choose to be canonical.

> Moreover, the way the class is defined at present there is no guarantee that composition of two such embeddings yield the same result.

This would have to be ensured manually for all instances, and is a better reason to not make it a class.

ekmett:
It is also worth noting that the pile of instances explodes quite drastically and becomes even harder if not impossible to properly handle "uniquely" once you start considering handling monad transformers correctly, or when you start considering that users themselves can write monads. The existing code doesn't even try to do so, but which reader layer promoted to which state layer? How to deal with insertions of additional transformer layers into the stack, etc.

I have no objection to adding a type for describing, composing, and using monad morphisms, I only object to passing them via implicits and calling the result a "partial order" when really it is an ad-hoc subcategory of the category of monad monomorphisms chosen more or less willy-nilly based on a combination of rules about what scala lets us talk about easily and what seems natural to the person writing it at the time that then requires omniscience on the behalf of the instance writer to know about all other instances that already exist in order to define a new one, as there is no way to make or check that these things automatically be transitive, etc.

As values the user can pass around freely and compose themselves, omniscience and "taste" are no longer required.

ekmett:
Speaking to the impossibility once you start trying to handle monad transformers correctly:

It is reasonable to consider StateT[S,M,A] as having two monad homomorphisms:

State[S,_] ~> StateT[S,M,_]

and

M[_] ~> StateT[S,M,_]

The existence of these two embeddings is one of the "common sense" ways of defining StateT.

Unfortunately, M can also be State[S,_], yielding two implicits, if we were to define them.

and also unfortunately we have no guarantee that the former is a monomorphism! Using (M = Proxy) provides a counter-example to the former being a monomorphism.

ekmett:
Speaking to the impossibility of handling composition in the presence of user defined monads / omniscience requirement.

Two libraries can define separate data types. One with a "monad partial order" embedding of M into state. And another with a "monad partial order" embedding from state to N, but there is no "monad partial order" instance embedding M into N.

PR #1433, opened 2017-08-10
Remove implicit functions

puffnfresh:


core/src/main/scala/scalaz/LeibnizF.scala
```diff
@@ -0,0 +1,28 @@
...
+    g: LeibnizF[A, B]
+  ): LeibnizF[A, C] =
+    f.subst[λ[X[_] => LeibnizF[A, X]]](g)
+
+  /** Equality is symmetric */
+  def symm[A[_], B[_]](
+    f: LeibnizF[A, B]
+  )  : LeibnizF[B, A] =
+    f.subst[λ[X[_]=> LeibnizF[X, A]]](refl)
+}
```
S11001001: Using infix `=~=` would make these signatures read more like their underlying implications. nbd

A derived congruence would also be nice:

```scala
def lift[T[_[_], _], A[_], B[_]](f: A =~= B): T[A, ?] =~= T[B, ?] = subst...
```

jedws:
+1

puffnfresh:
@xuwei-k this alright to go into 7.3?

PR #1435, opened 2017-08-12
Delete Boolean and Option methods for side-effects

puffnfresh:


tonymorris:
I agree. However, should they be rewritten to be pure?

puffnfresh:
`unlessM` and `whenM` already exist.

`o.ifNone` would equate to `o.isEmpty.whenM`. Should that be added?

tonymorris:
Yeah I reckon that additional function is useful. Maybe the `when` and `unless` identifiers should be used instead of `whenM` and `unlessM`.

PR #1437, opened 2017-08-15
Redefine scalaz.meta.Ops as a macro bundle

dwijnand:
Removes some boilerplate.

See http://docs.scala-lang.org/overviews/macros/bundles.html

Also upgrades some versions and drops the deprecated sbt.Build trait.

Feedback always welcome.

### DIFF
--- a/meta/shared/src/main/scala/scalaz/meta/Ops.scala
+++ b/meta/shared/src/main/scala/scalaz/meta/Ops.scala
@@ -5,35 +5,31 @@ import scala.reflect.macros.whitebox.Context
 
 // Originally inspired by http://typelevel.org/blog/2013/10/13/spires-ops-macros.html
 
-object Ops {
-  
-  
-  def _f0[R](c: Context): c.Expr[R] = {
-    import c.universe._
-    val (ev, lhs) = unpack(c)
+class Ops(val c: Context) {
+  import c.universe._
+
+  def _f0[R]: Expr[R] = {
+    val (ev, lhs) = unpack
     c.Expr[R](Apply(Select(ev, TermName(c.macroApplication.symbol.name.toString)), List(lhs)))
   }
-  
-  def _f1[A, R](c: Context)(f: c.Expr[A]): c.Expr[R] = {
-    import c.universe._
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -1,9 +1,10 @@
+import Scalaz._
 
 lazy val root = Project(
   id = "root",
   base = file(".")
 ).settings(
-  scalaVersion := "2.12.2"
+  scalaVersion := "2.12.3"
```
fommil: should we maybe create a `project/ProjectPlugin.scala` to allow use of `projectSettings`?

dwijnand: If what you want to achieve is to remove duplication then in this particular case I think even better than using AutoPlugin's projectSettings is using [sbt-travisci](https://github.com/dwijnand/sbt-travisci).

More generally, there are lots of ways to write builds - often down to personal preference/habit or niche scaling issues. So I held back to just version upgrades and deprecated Build, and even those are scope-creeping in this PR ;-)

project/ScalazBuild.scala
```diff
@@ -1,12 +1,12 @@
 import sbt._
 import Keys._
 
-object Scalaz extends Build {
+object Scalaz {
   val testDeps = Seq("org.scalacheck" %% "scalacheck" % "1.13.4" % "test")
 
   def stdSettings(prjName: String) = Seq(
```
fommil: this seems like it could be more easily implemented as the `projectSettings`

aloiscochard:
This is much nicer! Thanks a lot!

I'll create an issue about investigating usage of https://github.com/dwijnand/sbt-travisci, sounds like a great idea to me.

dwijnand:
Rebased onto current series/8.0.x, dropping the changes that have been integrated by merging @jbgi's #1442.

jdegoes:
👍

dwijnand:
Looks like @vmarquez, @TomasMikula and @aloiscochard have been pushing to series/8.0.x.

So review by any of the above, please.

fommil:
was this not merged?

jdegoes:
LGTM. Thanks! 🙏

PR #1443, opened 2017-08-20
[scalaz8] sbt 1.0, lint, fatal warnings, non-deprecated inlining options

jbgi:


jbgi:
followup on #1442

aloiscochard:
This looks great! I'm just curious about how `silent` can be used.

jbgi:
`@silent` hide any warning raised for the annotated expression. it allows to use `-Xfatalwarning` even if there is cases (like for the macro-based syntaxes) where a warning is unavailable (in that case because the annotated parameter is not used)

aloiscochard:
@jbgi oh I see, thanks for explaining.

It feels like it is really a scalac bug here:
- The compiler should inspect the expended code and see that the value is used (once expended)

Anyway, this is the kind of bug I guess we should not expect to be fixed any time soon I guess!

LGTM

PR #1444, opened 2017-08-26
Tuple's Band/Semigroup causes ambiguous implicits, need to swap

vmarquez:
we need to swap band and semigroup  so that they aren't ambiguous implicits.  I noticed this when I was adding SemiLattice but thought I'd fix this in a separate PR.


xuwei-k:
It doesn't make sense to me. Where are you going to add SemiLattice instances?

xuwei-k:
Please add more tests https://github.com/scalaz/scalaz/blob/5bf262da8f8d859b24a38d6da34cb5678f157871/tests/src/test/scala/scalaz/std/TupleTest.scala#L121

vmarquez:
Tuple in particular is confusing because unlike the others, `object tuple extends TupleInstances2`, NOT `TupleInstances0`.   Therefore, I believe SemiLattice is in the right place here.  

@xuwei-k is right though, my original PR had fixed the ambiguous issue with Band but not with semigroup and monoid.  I have updated now, let me know if that makes sense @puffnfresh . 

btw This is hard to show in tests as partial unification masks this issue...

puffnfresh:
@vmarquez ah bugger I'm very confused

emilypi:
hey @puffnfresh - do you still have strong feelings for this one? i'd like to merge if we're good.

PR #1450, opened 2017-09-04
Maybe as an abstract type, privately implemented via Option.

TomasMikula:
This PR exposes `Maybe` as an abstract type, privately implemented as `Option`.

The benefits of this approach are:
 - conversions between `Maybe` and `Option` are identities;
 - `IsCovariant[Maybe]` can be implemented without using any unsafe operations.

A side effect of making `Maybe` an abstract type is that it can't have a companion object, and thus any typeclass instances will be orphan, should anyone care.

---

The following is a description of what has already been merged in #1453.

**`IsCovariant[F[_]]`** is a typeclass that witnesses that `F` is covariant, even though the covariant type parameter annotation has been forgotten:

```scala
trait Module {
  type Maybe[A]
}

object ModuleImpl {
  type Maybe[A] = Option[A]
}

val module: Module = ModuleImpl
```

Here, `module.Maybe` is declared invariant, however, it is in fact covariant. We can preserve the information about covariance via an instance of `IsCovariant[Maybe]`:

```scala
trait Module {
  type Maybe[A]

  implicit def isCovariant: IsCovariant[Maybe]
}
```

`IsCovariant[F]` allows to lift `Liskov[A, B]` into `Liskov[F[A], F[B]]`, and is a safer alternative to indiscriminately lifting `Liskov` for every `Functor`. In the case when the implementation of `F` is actually Scala-covariant, `IsCovariant[F]` instance can be defined without any unsafe casts. For cases when `F` really is invariant, one can still apply the reasoning described in [When can Liskov be lifted?](https://typelevel.org/blog/2014/03/09/liskov_lifting.html) to justify an unsafe cast, but on a case-by-case basis, rather than across the board for all `Functor`s, present and future.

**`IsContravariant`** is analogous, for contravariant type constructors.


TomasMikula:
Alternatively, we could call them `ScalaCovariant`, `ScalaContravariant`.

puffnfresh:
I'd like to see what `ISet` does with this.

jdegoes:
I like the module style of mixing in functionality into `Prelude`. It could help ensure we don't forget useful functionality in `Prelude` and that we don't clash names. A safer `Liskov` from covariant Scala types also makes a lot of sense.

On the particular choice to use `Option` to back `Maybe`, there are a few concerns I'd raise:

 * Orphan type classes, as you note, which will complicate implicit search.
 * The fact that `Maybe` is a `val` will add an additional `invokevirtual` to every access of every function in the `Maybe` "module". I'm not sure if there's a way to eliminate this.
 * If the user can see `Maybe` is `Option`, then the many unsafe / partial functions of `Option` will be visible, which is undesirable.
 * If the user cannot see that `Maybe` is `Option`, then the only means to deconstruct is to `fold`. While suitable for most code, low-level, performance-intensive code will suffer greatly if every inspection of a `Maybe` requires allocation of two closures (which will necessitate storing references in `ObjectRef`, leading to yet more heap allocation and indirection), and then invoking one of these closures virtually. The cost of pattern matching, by comparison, is quite cheap.
 * Related to the above, `@inline` is basically unusable at this point, requiring both code annotations and build instructions to the inliner in *every* project that wishes to inline said code, limiting its usefulness in ameliorating performance problems introduced by indirection.

Other than these observations, there's lots of stuff I think is clearly a win in this PR.

TomasMikula:
@puffnfresh `ISet` won't benefit from this, because `ISet` is not Scala-covariant. If one wants to do Liskov-lifting for `ISet` anyway, one can implement `IsCovariant[ISet]` by forcing a `Liskov`.

I don't see lifting Liskov to ISet being useful, though, and could be dangerous, if the `Order` instances are not compatible.

TomasMikula:
@jdegoes 

> * Orphan type classes, as you note, which will complicate implicit search.

:man_shrugging: 

> * The fact that `Maybe` is a `val` will add an additional `invokevirtual` to every access of every function in the `Maybe` "module". I'm not sure if there's a way to eliminate this.

I don't know either. But if `MaybeImpl` was the only implementation of (sealed) `MaybeModule`, then the compiler would have all the information necessary to eliminate that `invokevirtual`.

> * If the user can see `Maybe` is `Option`, then the many unsafe / partial functions of `Option` will be visible, which is undesirable.

Right. This is invisible to the user.

> * If the user cannot see that `Maybe` is `Option`, then the only means to deconstruct is to `fold`. While suitable for most code, low-level, performance-intensive code will suffer greatly if every inspection of a `Maybe` requires allocation of two closures (which will necessitate storing references in `ObjectRef`, leading to yet more heap allocation and indirection), and then invoking one of these closures virtually. The cost of pattern matching, by comparison, is quite cheap.

Additionally, `fold` doesn't allow for tail-recursion (though not relevant for non-recursive data types like `Maybe`).

But there _is_ a means for pattern matching:

```scala
val ma: Maybe[A]

option(ma) match {
  case Some(a) => ???
  case None    => ???
}
```

`option(ma)` is identity, but not sure if it can be inlined.

> * Related to the above, `@inline` is basically unusable at this point, requiring both code annotations and build instructions to the inliner in *every* project that wishes to inline said code, limiting its usefulness in ameliorating performance problems introduced by indirection.

Do you mean inlining those `invokevirtual`s or those `fold`s?

S11001001:
> I don't know either. But if MaybeImpl was the only implementation of (sealed) MaybeModule, then the compiler would have all the information necessary to eliminate that invokevirtual.

It's probably irrelevant whether there's one implementation, and it's probably also irrelevant whether it's sealed. It's probably relevant whether the `val`'s concrete implementation is statically apparent at the point of call, and whether that `val` has a statically apparent implementation of `option` et al, &c.

[Scala.js does an awesome job in that case.](https://failex.blogspot.com/2017/04/and-glorious-subst-to-come.html#markdown-header-would-you-box-a-javascript-string) In response to @Blaisorblade, I [tried with Scala JVM earlier, which did a terrible job and didn't inline anything](https://failex.blogspot.com/2017/04/the-high-cost-of-anyval-subclasses.html?showComment=1501441994974#c8343570638013278411).

Blaisorblade:
>> then the compiler would have all the information necessary to eliminate that invokevirtual.

> It's probably irrelevant whether there's one implementation, and it's probably also irrelevant whether it's sealed.

You're talking about different things — @TomasMikula's comment is about what a *sufficiently smart* compiler could do statically, and I think `sealed` is relevant there.

But Scalac.JVM is not even remotely "sufficiently smart", because it delegated everything to JIT compilers (which seems to sometimes not work well). Dotty added an optimizer (which IIUC is still being refined)—which isn't "sufficiently smart" like GHC (especially yet), but at least tries something.

Atry:
As far as I know, HotSpot JVM is very good at inline such kind of virtual calls.

It will be replaced to some assembly like following:

```
if (module is a ModuleImpl$) {
  run the inlined method body, which is a no-op
} else {
  looking for vtable for methods on Module
}
```

The condition jump costs less than 5 nanoseconds according to some benchmarks I made many years ago.

Atry:
Also JVM will take advantage from the fact that there is only one loaded class implementation of `Module`. JVM will discard the JIT generated code if the fact is break when new classes are loaded into class loaders.

That is to say, virtual calls to `Function1` usually cost more than `ModuleImpl` because there are more class implementations.

TomasMikula:
> It's probably irrelevant whether there's one implementation

If there's just a single implementation, even when it remains `invokevirtual`, all the call sites will be "monomorphic", which the JVM can take advantage of. From https://wiki.openjdk.java.net/display/HotSpot/VirtualCalls:

> A monomorphic call site is one which optimistically points to the only concrete method that has ever been used at that particular call site. The calling sequence consists of verifying that the current (dynamically determined) receiver is exactly the same type as the previous call. That check is a simple pointer comparison, of a constant versus the InstanceKlass* loaded (via one indirection) from the object header. If the check succeeds, the call is made directly.

I think this is what @Atry is talking about.

Anyway, that `val` could perhaps be changed into an `object`.

Atry:
Yes, the performance of HotSpot JVM is merely affected by `final` or `sealed` keywords.

TomasMikula:
There are mainly two kinds of concerns:
 - performance concerns;
 - implicit instance resolution.

People seem mainly positive about the `IsCovariant`, `IsContravariant` typeclasses though.

I shall split this PR into 2 then:
 1. `IsCovariant`, `IsContravariant`,
 2. `Maybe` as abstract type, privately alias for `Option`.

Blaisorblade:
> Yes, the performance of HotSpot JVM is merely affected by final or sealed keywords.

It's not "sealed" (which the JVM doesn't see): the JVM counts (1) subclasses (if only one, the call can be optimized easily) and (2) actual targets (if only one is used, the call site can still be compiled as monomorphic, but needs the instance check you've shown).

On performance more in general: JIT compilers are good, and this looks like an easy case for JIT compilers, but non-inlined bytecode can get in the way of other optimizations (until inlined) — for instance, Hotspot inlines all methods under 35 bytes of bytecode. Failed optimizations can lead to further optimizations failing. And so on.
Also, interactive applications might spend non-trivial time outside of the optimized steady-state execution.

TomasMikula:
#1453 contains `IsCovariant` and `IsContravariant`. I rebased this PR onto that one.

jbgi:
Nice. I just saw two inconveniences with this implementation of `Maybe`:
 1. no support for deconstructing Maybe in nested patterns (cannot write eg. `case Foo(Some(a)) => `)
 2. exclusive of other implementation trick, like lazy constructor (https://github.com/scalaz/scalaz/pull/1366)

(plus the implicit instances search, but I'd say that `Maybe` instances could be added to `Prelude` to avoid any more specific import)

TomasMikula:
I haven't realized the problem with nested patterns. One can of course define custom extractors, but that breaks exhaustivity checking. 😕

> exclusive of other implementation trick, like lazy constructor (#1366)

I guess you mean this about the specific implementation of `Maybe` via `Option`, not the technique of invariant API/covariant implementation in general?

puffnfresh:
We should provided optics to replace nested matching.

Atry:
I created a project of these kind of opacity types at: https://github.com/ThoughtWorksInc/ZeroCost.scala

However, I would ask you guys have a look at another encoding of zero-cost abstractions of type classes https://github.com/typelevel/general/issues/81

Atry:
The key construct for the new representation is [BoxFactory](https://github.com/ThoughtWorksInc/DesignPattern.scala/blob/9f1a44c/designpattern/src/main/scala/com/thoughtworks/designpattern/covariant.scala#L9)

TomasMikula:
Fighting the problem with implicit instances at scala/scala#6074.

alexknvl:
@TomasMikula 
You can avoid `invokevirtual` if you declare the `val` in java.
```scala
trait BazT {
  type T
  def zero: T
}
final class BazI extends BazT {
  type T = Int
  def zero: T = 0
}
import Types.Arr
class Foo {
  type Arr = Arr.T
  def z: Arr = Arr.zero
}
```

```java
public class Types {
    public static final ArrT Arr = new ArrI();
}
```

```java
public class nutty.Foo {
  public java.lang.Object z();
    Code:
       0: getstatic     #16                 // Field Types.Arr:Lnutty/ArrT;
       3: invokeinterface #21,  1           // InterfaceMethod ArrT.zero:()Ljava/lang/Object;
       8: areturn
```

TomasMikula:
@alexknvl This is interesting! It eliminates the `invokevirtual` for field access.

There's still the `invokeinterface` for the method `zero`. I'm working on a Scala PR that would inline `zero` in case `BazT` is `sealed` and `BazI` is the only concrete subclass of `BazT` and `BazI#zero` has `@inline` annotation. Stay tuned.

Putting these two together, we would eliminate virtual calls altogether.

TomasMikula:
[Here](https://github.com/scala/scala/pull/6088) is the PR. Unfortunately, @sjrd immediately spotted a problem with binary compatibility / separate compilation, so will need to work harder on that to happen.

TomasMikula:
The [opaque types proposal](http://docs.scala-lang.org/sips/opaque-types.html) brings a lot of hope for this approach.

alexknvl:
I found a new encoding for newtypes that solves the implicit resolution problem (does require some careful `asInstanceOf`ing though):

```scala
type EitherT[F[_], E, A] = EitherT.Type[F, E, A]
object EitherT {
  type Base$1
  trait Tag$1 extends Any
  type Type[F[_], E, A] <: Base$1 with Tag$1
  
  @inline final def wrap[F[_], E, A](e: Either[E, F[A]]): Type[F, E, A] = e.asInstanceOf[Type[F, E, A]]
  @inline final def unwrap[F[_], E, A](e: Type[F, E, A]): Either[E, F[A]] = e.asInstanceOf[Either[E, F[A]]]

  implicit def functor[F[_], E](implicit F: Functor[F]): Functor[({type L[X] = Type[F, E, X]})#L] = new Functor[({type L[X] = Type[F, E, X]})#L] {
    def map[A, B](fa: Type[F, E, A])(f: A => B): Type[F, E, B] = wrap(unwrap(fa).right.map(F.map(_)(f)))
  }
}

// Assuming that we have syntax implicits in scope
println(EitherT.wrap(Right[String, Option[Int]](Some(1))).map(_ + 1))
```

```scala
  type Base$1 
  // abstract, so it will be treated as universally / existentially quantified
  trait Tag$1 extends Any
  // <: Any is important for AnyVal types
  type Type[F[_], E, A] <: Base$1 with Tag$1
  // will be represented as Base$1 but for the purposes of implicit resolution the compiler
  // will use Tag$1 as well
```

https://github.com/alexknvl/newtypes implements it as a macro but it depends on the very latest version of scalameta, so it is 2.12.3 only :smile:

TomasMikula:
This sure takes some creativity! I keep being amazed and terrified at the same time by the things people come up with in Scala. 😄 

I'm also surprised that `(e: Either[E, F[A]]).asInstanceOf[Type[F, E, A]]` is safe when `Type[F[_], E, A]` is a subtype of `trait Tag$1`. I guess scalac just forgoes any checking in this case? Could this be considered a bug?

alexknvl:
@TomasMikula Not necessarily. `A with B` usually erases to `A` unless `A` is one of `Any`, `AnyRef` or `AnyVal` AFAIK (in which case it erases to `B`). Since `B` here does not have any methods on it, scalac will never attempt to cast to it. Ideally you also want to make `Tag$1` private but since it has some special characters on it I doubt anyone will think it is safe to use :smile:

alexknvl:
There are some corner cases where it is not safe and extra care must be taken. For instance, `Array[Int]` and `Array[Int with Tag]` will erase to two different types `int[]` and `Object[]`. In order to ensure type-safety we need to make scalac think that the type inside `Array` is polymorphic (hence that existential type and inheriting `Tag` from `Any`), then it will generate calls to `ScalaRuntime.array_get` instead of trying to use `Object[]` methods.

alexknvl:
Cast checking is necessary only to get through the java bytecode verifier, so scalac does the minimum possible amount of work there.

```scala
trait A { def bar: Int }
trait B { def blah: Int }
val baz: Any = ...
val foo: A with B = baz.asInstanceOf[A with B]
foo.bar + foo.blah
```
erases to
```scala
trait A { def bar: Int }
trait B { def blah: Int }

val baz: Object = ...
val foo: A = baz.asInstanceOf[A]
foo.bar + (foo.asInstanceOf[B]).blah
```

jdegoes:
@alexknvl Very clever. Nice work.

jdegoes:
@TomasMikula Could this same approach be used to encode `Maybe[A]` as nullable (and therefore necessarily boxed) `A`?

TomasMikula:
You mean representing `Empty` as `null`? How would you then represent `Just(null)`?

jdegoes:
> You mean representing Empty as null? How would you then represent Just(null)?

It's a pretty consistent philosophy in Scalaz we only design for correct code (for example, `IO.point`'s behavior is undefined if you misuse the method for effectful code, which allows significant increases in performance). If the user is using null, their code is broken and all bets are off. Of course, you can guard against this case with a runtime exception on construction.

TomasMikula:
So we are allowed to exploit `null`, but our users aren't?

jdegoes:
> So we are allowed to exploit null, but our users aren't?

That's correct. You can't write a principled FP library if you do not assume basic things, such as: user will not be executing effects (e.g. in a function passed to `fmap`), user will not be passing null values into the library, user will not be throwing exceptions (unless the intent is to crash the program) in supposedly pure functions, and so forth.

Blaisorblade:
@jdegoes What then about `Maybe[Maybe[A]]` and (IIUC) `Just(Empty)`?

jdegoes:
@Blaisorblade I think this example shows you can't use a fixed encoding (such as literally `type Maybe[A] = A`) and you need runtime type checks. That's probably enough to kill the idea because it will push type tags through all polymorphic code. Which is non-composable.

Blaisorblade:
No good way to encode type families, uh? I'd have loved `Maybe[A] = if <A isn't Maybe> then A else <insert ADT here>`. Of course, this would forbid using `null` to implement other types.

> So we are allowed to exploit null, but our users aren't?

I'd say, whoever writes unsafe code has to prove it's in fact safe, just like @alexknvl 's use of casts. But correct casts are encapsulated, while `null` isn't so it's trickier.

TomasMikula:
> Of course, you can guard against this case with a runtime exception on construction.

That would mean _our_ code was not parametric.

Blaisorblade:
>> Of course, you can guard against this case with a runtime exception on construction.
> That would mean our code was not parametric.

Parametricity is a statement about observable behavior, not about how that behavior is implemented—though well-behaved terms are automatically parametric.
Checking against `null` isn't a behavior that is observable by users, unless a user cheats in the first place—and since Scalaz users shouldn't cheat, being parametric only for non-cheating code seems acceptable.

TomasMikula:
What if another library applied the same reasoning and decided to represent, say, an empty list as `null`? This would be completely opaque to the user, and the poor user would try to wrap such empty list (`null`) in our `Maybe`. We can't blame the user for cheating here. What gives _us_ the privilege to exploit `null`, while others (including other libraries) are denied this privilege?

Blaisorblade:
> What gives us the privilege to exploit null, while others (including other libraries) are denied this privilege?

If there aren't other technical constraints, that's a social choice. Would other libraries leave you that privilege? Is Scalaz as fundamental as, say, GHC's implementation?

I suspect Scalaz's default contract is to NOT use cheats that leak and restrict the API and NOT rely on such social contracts, though this seems a grey area.

jdegoes:
> No good way to encode type families, uh? I'd have loved Maybe[A] = if <A isn't Maybe> then A else <insert ADT here>.

If there's a way to do it (without type tags), I don't know how.

edmundnoble:
This argument sounds like it's very important to have, but in this particular issue non-parametric `Maybe` is not worth the cost versus https://github.com/sjrd/scala-unboxed-option.

jdegoes:
@edmundnoble Agreed.

jdegoes:
@TomasMikula There are still a few outstanding issues, but perhaps we should merge this in after any quick fixes and I can open up tickets for the remainder?

TomasMikula:
I will resolve the conflicts. But maybe we can wait for Scala 2.12.4 (which should be around the corner) and then avoid mixing implicit instances into Prelude, since then we can put them in the companion object of `MaybeModule` and they will be resolved automatically.

jdegoes:
@TomasMikula Sounds good! 👍

TomasMikula:
OK, updated. Also added extractors for pattern matching.

jdegoes:
Looks great! I think any additional concerns we can address on followup PRs / tickets.

TomasMikula:
Although `type Maybe[A] = A`, representing the empty case as `null`, can't represent `Maybe[Maybe[A]]`, the problem doesn't translate to `Maybe2`. We could have

```scala
type Maybe2[A, B] = (A, B)

def just2[A, B](ab: (A, B)): Maybe2[A, B] = {
  require(ab != null)
  ab
}
def empty2[A, B]: Maybe2[A, B] = null
```

TomasMikula:
Yeah, there's an improvement only if you already had a `Tuple2` allocated anyway.

I kept `Option2` package private because I thought people wouldn't like covariant types in public API. I personally am fine with making it public.

TomasMikula:
The lack of exhaustive pattern matching for opaque types _is_ annoying. Here's an idea, I'm not sure if it's feasible.

We have

```scala
trait MaybeModule {
  type Maybe[A]

  object Just {
    def unapply[A](ma: Maybe[A]): Option[A] = toOption(ma)
  }

  object Empty {
    def unapply[A](ma: Maybe[A]): Boolean = toOption(ma).isEmpty
  }
}
```

I imagine a `switch` macro

```scala
trait MaybeSyntax {
  implicit final class MaybeOps[A](ma: Maybe[A]) {
    def switch[B](f: Maybe[A] => B): B = meta.switch[(Maybe.Just.type, Maybe.Empty.type)]
  }
}
```

where the tuple type argument (or could be `HList`) is the exhaustive list of extractors.

The usage would look quite naturally

```scala
val ma: Maybe[A]

ma switch {
  case Just(a) => ???
  case Empty() => ???
}
```

The `switch` macro would check that all declared extractors are used and rewrite the whole expression to a series of `if` statements (to allow for tailrec optimization).

Again, I'm not sure it is doable with macros. It also wouldn't support nested patterns.

PR #1453, opened 2017-09-04
IsCovariant, IsContravariant.

TomasMikula:
See #1450.

### DIFF
--- a/base/shared/src/main/scala/scalaz/typeclass/IsContravariantClass.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/IsContravariantClass.scala
@@ -1,0 +1,56 @@
+package scalaz
+package typeclass
+
+import Liskov.<~<
+
+trait IsContravariantClass[F[_]] extends IsContravariant[F] {
+  final def isContravariant: IsContravariant[F] = this
+}
+
+object IsContravariantClass {
+
+  trait SubstCv[F[_]] extends Alt[SubstCv[F]] { self: IsContravariant[F] =>
+    override def substCv[G[+_], A, B](g: G[F[B]])(implicit ev: A <~< B): G[F[A]]
+ 
+    override def substCt[G[-_], A, B](g: G[F[A]])(implicit ev: A <~< B): G[F[B]] = {
+      type H[+a] = G[a] => G[F[B]]
+      substCv[H, A, B](identity[G[F[B]]]).apply(g)
+    }
+ 
+    override def liftLiskov[A, B](implicit ev: A <~< B): F[B] <~< F[A] =
+      substCv[F[B] <~< +?, A, B](Liskov.refl[F[B]])
+ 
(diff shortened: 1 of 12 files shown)


base/shared/src/main/scala/scalaz/typeclass/IsContravariantClass.scala
```diff
@@ -0,0 +1,56 @@
+package scalaz
+package typeclass
+
+import Liskov.<~<
+
+trait IsContravariantClass[F[_]] extends IsContravariant[F] {
+  final def isCovariant: IsContravariant[F] = this
```
S11001001: `isContravariant`?

base/shared/src/main/scala/scalaz/typeclass/LiskovInstances.scala
```diff
@@ -7,9 +7,6 @@ import Liskov.<~<
 
 trait LiskovInstances {
   /**We can witness equality by using it to convert between types */
-  implicit def witness[A, B](lt: A <~< B): A => B = {
-    type f[-X] = X => B
-    lt.subst[f](identity)
-  }
+  implicit def witness[A, B](lt: A <~< B): A => B = lt(_)
```
S11001001: Hmm. If you sign this `implicit def witness[A, B](a: A)(implicit lt: A <~< B): B = lt(a)` instead, does this compile?

```scala
def qqq[A, B](implicit lt: A <~< B): A => B = a => a
```

At the very least, an `implicit` keyword is missing here.

base/shared/src/main/scala/scalaz/typeclass/IsContravariantClass.scala
TomasMikula: Yes, thanks!

base/shared/src/main/scala/scalaz/typeclass/LiskovInstances.scala
TomasMikula: Don't know, will try.

> At the very least, an implicit keyword is missing here.

I guess @aloiscochard meant this as an implicit conversion of `<~<` to function, not to provide an implicit functions.

TomasMikula: It compiles. That might make more sense. Note, however, that I didn't touch the signature, just the implementation.

S11001001: To a certain degree, those are the same thing now. I'm just not sure to what degree.

TomasMikula: @aloiscochard What do you think about changing `witness` to what @S11001001 proposed, i.e.

```scala
implicit def witness[A, B](a: A)(implicit lt: A <~< B): B = lt(a)
```

Passing `A <~< B` to a function that requires `A => B` could be achieved more simply by `Liskov` extending `Function1`.

PR #1454, opened 2017-09-06
Interaction of universal quantification and subtyping.

TomasMikula:
Added three rules:

```scala
∀[F] <: F[A]

A <: ∀[λ[α => A]]

F[_] <: G[_]   implies   ∀[F] <: ∀[G]
```

These correspond, respectively, to rules (4), (5) and (3) from
http://www.sciencedirect.com/science/article/pii/S0890540101929505

---

I wanted to add another one,

```scala
∀α.G[F[α]] <: G[∀α.F[α]]  // for G covariant and F any-variant
```

which would correspond to rule (6) from the linked paper and to quantifier distributivity in logic over any covariant position, but couldn't come up with proper justification for it. I still think it's valid in Scala, just pending a good argument why.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/Forall.scala
+++ b/base/shared/src/main/scala/scalaz/data/Forall.scala
@@ -1,6 +1,8 @@
 package scalaz
 package data
 
+import typeclass.Liskov
+import typeclass.Liskov.<~<
 import scala.language.implicitConversions
 
 trait ForallModule {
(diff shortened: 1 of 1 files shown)


base/shared/src/main/scala/scalaz/data/Forall.scala
```diff
@@ -20,6 +22,18 @@ trait ForallModule {
 
   def mk[X](implicit u: Unapply[X]): MkForall[u.F] = of[u.F]
 
+  def instantiation[F[_], A]: ∀[F] <~< F[A]
```
aloiscochard: Nice!

aloiscochard:
:+1:

jdegoes:
@TomasMikula Thanks!

PR #1455, opened 2017-09-06
[WIP] Sketch of IList with invariant interface and covariant implementation.

TomasMikula:
This is in the same spirit as `Maybe` in #1450.

Things to note:
 - `uncons` is just `Fix.unfix`, which is identity.
 - `IsCovariant[IList]` instance is created without (directly) using any unsafe operations (like `asInstanceOf` or `Liskov.unsafeForce`).
 - Pattern matching is available (via `uncons(as) match { ... }`; though this can't be used in nested patterns).

The major problem, as with #1450, is the lack of companion object to look up typeclass instances.

To match the memory efficiency of `scala.List`, we would have to switch to a specialized implementation of `Option2[A, B]` instead of `Option[(A, B)]` to avoid the extra `Tuple2` instance.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/Forall.scala
+++ b/base/shared/src/main/scala/scalaz/data/Forall.scala
@@ -12,6 +12,7 @@ trait ForallModule {
 
   trait Prototype[F[_]] {
     def apply[A]: F[A]
+    final def make: ∀[F] = from(this)
   }
 
   def specialize[F[_], A](f: ∀[F]): F[A]
--- a/base/shared/src/main/scala/scalaz/data/Forall2.scala
+++ b/base/shared/src/main/scala/scalaz/data/Forall2.scala
@@ -10,6 +10,7 @@ trait Forall2Module {
 
   trait Prototype[F[_, _]] {
     def apply[A, B]: F[A, B]
+    final def make: ∀∀[F] = from(this)
   }
 
   def specialize[F[_, _], A, B](f: ∀∀[F]): F[A, B]
--- a/base/shared/src/main/scala/scalaz/data/IList.scala
+++ b/base/shared/src/main/scala/scalaz/data/IList.scala
@@ -1,0 +1,31 @@
+package scalaz
+package data
+
+import Prelude._
+import typeclass.IsCovariantClass
+import typeclass.Liskov.<~<
+
+trait IListModule {
+  type IList[A]
+
+  def uncons[A](as: IList[A]): Maybe2[A, IList[A]]
+
+  implicit def isCovariantInstance: IsCovariant[IList]
+}
+
+private[data] object IListImpl extends IListModule {
+  type IList[A] = Fix[Maybe2[A, ?]]
+
+  def uncons[A](as: IList[A]): Maybe2[A, IList[A]] =
+    Fix.unfix[Maybe2[A, ?]](as)
+
+  implicit val isCovariantInstance: IsCovariant[IList] =
+    new IsCovariantClass[IList] with IsCovariantClass.LiftLiskov[IList] {
+
+      override def liftLiskov[A, B](implicit ev: A <~< B): IList[A] <~< IList[B] = {
+        type <~~<[F[_], G[_]] = ∀.Prototype[λ[α => F[α] <~< G[α]]]
+        val ev1 = Λ[α](Maybe2.isCovariant_1[α].liftLiskov[A, B]): Maybe2[A, ?] <~~< Maybe2[B, ?]
+        Fix.liftLiskov[Maybe2[A, ?], Maybe2[B, ?]](ev1.make)(Maybe2.isCovariant_2[A])
+      }
+    }
+}
--- a/base/shared/src/main/scala/scalaz/data/Maybe2.scala
+++ b/base/shared/src/main/scala/scalaz/data/Maybe2.scala
@@ -1,0 +1,65 @@
+package scalaz
+package data
+
+import typeclass.IsCovariant
+
+sealed trait Maybe2Module {
+  /**
+   * Isomorphic to `Maybe[(A, B)]`, but avoids allocating a `Tuple2` isntance.
+   */
+  type Maybe2[A, B]
+
+  def just2 [A, B](a: A, b: B): Maybe2[A, B]
+  def empty2[A, B]            : Maybe2[A, B]
+
+  implicit def isCovariant_1[B]: IsCovariant[Maybe2[?, B]]
+  implicit def isCovariant_2[A]: IsCovariant[Maybe2[A, ?]]
+
+  object Just2 {
+    def unapply[A, B](m: Maybe2[A, B]): Just2Extractor[A, B] = new Just2Extractor(toOption2(m))
+  }
+  object Empty2 {
+    def unapply[A, B](m: Maybe2[A, B]): Boolean = toOption2(m).isEmpty
+  }
+
+  private[data] def fromOption2[A, B](o: Option2[A, B]): Maybe2[A, B]
+  private[data] def toOption2  [A, B](m: Maybe2[A, B]): Option2[A, B]
+}
+
+final class Just2Extractor[A, B] private[data] (private val value: Option2[A, B]) extends AnyVal {
+  def isEmpty: Boolean = value.isEmpty
+  def get: Just2Extractor[A, B] = this
+  def _1: A = value._1
+  def _2: B = value._2
+}
+
+private[data] object Maybe2Impl extends Maybe2Module {
+  type Maybe2[A, B] = Option2[A, B]
+
+  def just2 [A, B](a: A, b: B): Maybe2[A, B] = Some2(a, b)
+  def empty2[A, B]            : Maybe2[A, B] = None2
+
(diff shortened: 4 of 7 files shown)


TomasMikula:
I also added the [P∀scal compiler plugin](https://github.com/TomasMikula/pascal) to help with creation of universally quantified values. It is used [here](https://github.com/scalaz/scalaz/pull/1455/files#diff-4232f28a295ca4a0d6a08e36f92e5255R29).

jdegoes:
From a performance perspective, my main concern with this, which is my main concern with fixed point data types in general, is that it literally doubles the overhead of construction and desconstruction, both from a perspective of function calls, and from a perspective of heap allocation / churn / pressure. My own feeling is that fixed point data types are best used in cases where the benefits of fixed points (including ADT sharing & dynamism, smaller code base, etc.) overwhelm the very real and measurable performance drawbacks of the type.

TomasMikula:
> my main concern with fixed point data types in general, is that it literally doubles the overhead of construction and desconstruction [...] from a perspective of heap allocation / churn / pressure.

Notice that the implementation of `Fix` doesn't box, so memory efficiency should be the same as with dedicated implementation.

TomasMikula:
Assuming a dedicated implementation of `Option2[A, B]` instead of `Option[(A, B)]`, that is.

👍 for benchmarking different designs.

jdegoes:
@TomasMikula Excellent point, 👍 for the dedicated implementation of `Option2`!

base/shared/src/main/scala/scalaz/data/Fix.scala
```diff
@@ -0,0 +1,27 @@
...
+  /* functions */
+
+  def   fix[F[_]](f: F[Fix[F]]):   Fix[F]
+  def unfix[F[_]](f:   Fix[F] ): F[Fix[F]]
+
+  def liftLiskov[F[_], G[_]](ev: ∀[λ[α => F[α] <~< G[α]]])(implicit F: IsCovariant[F]): Fix[F] <~< Fix[G]
+}
+
+private[data] trait FixImpl extends FixModule {
+  type Fix[F[_]] = F[Any]
```
S11001001: [This can be safer.](https://github.com/TomasMikula/scalaz/compare/IList...S11001001:IList+Fix)

TomasMikula: Nice! But I wanted to make `data.Fix` an `object` instead of a `val`, due to performance concerns with `invokevirtual` raised in #1450, where I used a `val`.

S11001001: The problem with that approach is, as of e5566352

```scala
scala> List(1, 2, 3): scalaz.data.Fix[List]
res0: scalaz.data.Fix[List] = List(1, 2, 3)

scala> def conforms[A, B >: A] = 'yep
conforms: [A, B >: A]=> Symbol

scala> conforms[scalaz.data.IList[String], scalaz.data.IList[Any]]
res2: Symbol = 'yep

scala> conforms[Option[(Int, Any)], scalaz.data.IList[Int]]
res6: Symbol = 'yep
```

TomasMikula: Ah, thanks for checking this! I had a suspicion we can't trust Scala with keeping `private` things private. I will incorporate your suggestion then.

TomasMikula:
Factored out `Fix` into a separate PR, #1472.

Added `Option2`.

base/shared/src/main/scala/scalaz/data/Option2.scala
```diff
@@ -0,0 +1,12 @@
+package scalaz
+package data
+
+/**
+ * Isomorphic to `Option[(A, B)]`, but avoids allocating a `Tuple2` instance.
+ *
+ * Covariance is intentional. For invariant version, consider implementing `Maybe2`.
+ */
+sealed abstract class Option2[+A, +B]
```
jdegoes: Is a covariant `Option2` strictly needed to implement `IsCovariant[IList]`? Seems like you could just implement `IsCovariant2[Maybe2]` or something similar.

TomasMikula: You could do that, but the problem then shifts to how to implement `IsCovariant2[Maybe2]`. You would have to use `asInstanceOf` or equivalent (e.g. `Liskov.unsafeForce`).

jdegoes: Maybe this should trigger a re-discussion of when / how to use variance in Scalaz 8?

TomasMikula: We could also keep `Option2` package private, and expose `Maybe2` as an invariant type alias for `Option2` (just like `Maybe` in #1450). That way, we safely implement `IsCovariant[Maybe2]` via deferring to covariant `Option2`.

jdegoes: I like that, since it means we don't have to make any decisions on variance (defer all the things, the mantra of FP 😆). It would be nice to keep it in the same file to not clutter up the file structure.

TomasMikula: Aah, we still have to keep `Option2` public in order to support pattern matching on `Maybe2` by conversion to `Option2`.

jdegoes: Can't we just have unapply on a `Maybe2` object?

TomasMikula: We can, but what does it unapply to?

jdegoes: A tuple. Which involves additional boxing, sadly.

TomasMikula: Updated to use `Maybe2`, but still kept `Option2` public.

jdegoes: Something like this would work, wouldn't it?

```scala
object Some2 {
  def apply[A, B](v: Maybe2[A, B]): Option[(A, B)] = ???
}
object None2 {
  def apply[A, B](v: Maybe2[A, B]): Option[Unit] = ???
}
```

Then a user can do:

```scala
maybe2 match {
  case Some2(a, b) => ???
  case None2() => ???
}
```

TomasMikula: You mean `unapply`. Yeah, that would work, but loses exhaustivity checking and allocates a `Tuple2`.

TomasMikula: Scala is like a hydra: working around one problem you hit another two.

jdegoes: I think I found a way to do it:

```
sealed trait Maybe2[A, B]
final case class Some2[A, B](_1: A, _2: B) extends Product2[A, B] with Maybe2[A, B]
final case class None2[A, B]() extends Maybe2[A, B]
object Maybe2 {
  def some[A, B](a: A, b: B): Maybe2[A, B] = new Some2(a, b)
  def none[A, B]: Maybe2[A, B] = new None2[A, B]()
}
object Some2 {
  def unapply[A, B](m: Maybe2[A, B]): Option[Product2[A, B]] = m match {
    case s : Some2[_, _] => Some(s: Product2[A, B])
    case _ => None
  }
}
object None2 {
  def unapply[A, B](m: Maybe2[A, B]): Option[Unit] = m match {
    case s @ Some2(_, _) => None
    case _ => Some(())
  }  
}
Maybe2.some("a", "b") match { case Some2(a, b) => a + b }
```

This does not address exhaustivity. I think there is an SI ticket for that somewhere.

TomasMikula: Here you would again run into trouble of how to define `IsCovariant`, but you got me thinking in the right direction (minus exhaustivity).

TomasMikula:
Keeping the covariant `Option2` private. Pattern matching on `Maybe2` demonstrated in `Maybe2Usage`. We avoid boxing in pattern matching, but don't have exhaustivity checking.

jdegoes:
This looks very promising now. The question is what to do about exhaustivity checking. I found an old resolved ticket on the Scala issue tracker, which seemed relevant, but clearly isn't. It seems to me there's enough information to statically determine exhaustivity.

TomasMikula:
I don't see how the compiler could determine exhaustivity, unless it analyzes the code of the extractors.

TomasMikula:
As a curiosity, I wrote a proof of concept that representation of `List[A]` as `Fix[Maybe2[A, ?]]` also translates to analogous representation of type-aligned lists; we just need a type-aligned version of `Fix`.

```scala
type AFix[F[_[_, _], _, _], A, B] = F[AFix[F, ?, ?], A, B]

type AList[F[_, _], A, B] = AFix[AMaybe2[F, ?[_, _], ?, ?], A, B]
```

where (roughly)

```scala
type AMaybe2[F[_, _], G[_, _], A, B] = Maybe[(F[A, X], G[X, B]) forSome { type X }]
```

Full implementation is in my [AList-via-AFix branch](https://github.com/TomasMikula/scalaz/tree/AList-via-AFix).

jdegoes:
@TomasMikula Very nice!

jdegoes:
@TomasMikula I'm going to merge this and track exhaustivity in a separate ticket. Excellent work, as always.

PR #1456, opened 2017-09-07
Ephemeral stream consistency

### LINKED ISSUE #1446: EphemeralStream.fold{Left,Right} have the wrong type signature
```
def foldLeft[B](z: => B)(f: (=> B) => (=> A) => B): B = ...
```

should be `foldl` (which is more compatible), similarly for `Right`.

This causes the following code to fail, which is inconsistent with `Foldable`'s API

```
scala> EphemeralStream.range(1, 10000).foldLeft(0L)(_ + _)
<console>:40: error: missing parameter type for expanded function ((x$1: <error>, x$2) => x$1.$plus(x$2))
       EphemeralStream.range(1, 10000).foldLeft(0L)(_ + _)
                                                    ^
<console>:40: error: missing parameter type for expanded function ((x$1: <error>, x$2: <error>) => x$1.$plus(x$2))
       EphemeralStream.range(1, 10000).foldLeft(0L)(_ + _)
                                                        ^
```

This has a trivial fix but it breaks the binary and source API so I don't know what to do.

fommil:
close #1446 

I also added jenv config and enough heap to compile the project in `.jvmopts` (used by the `sbt` launch script).


fommil:
who should I be hassling to merge? :smile:

jdegoes:
I am happy to if no one else does. I've never merged anything here before so not 100% sure on the etiquette. 😆

PR #1457, opened 2017-09-07
Polymorphic singletons

TomasMikula:
Use polymorphic singleton values in order to avoid either of
 - allocating a new instance for each type argument; or
 - using `asInstanceOf` to cast the singleton object to the desired type.

### Approach A

```scala
case class Empty[A]() extends Maybe[A]
```

👎  Allocates a new `Empty[T]` instance for each type `T`.

### Approach B

```scala
case object Empty extends Maybe[Nothing]
```

👎  Use `Empty.asInstanceOf[Maybe[T]]` to get an empty instance of `Maybe[T]`. (Note that `Maybe` is assumed to be invariant.)

### Approach C (this PR)

```scala
class Empty[A] private[Maybe]() extends Maybe[A]
val Empty: ∀[Empty] = ∀.of[Empty].from(new Empty)
```

To get an empty instance of `Maybe[T]`, use

```scala
Empty[T]
```

which doesn't allocate any new objects (assuming zero-cost syntax that adds the `.apply[T]` method to `∀[Empty]`).

👍  Avoids both downsides of approaches A and B.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/Forall.scala
+++ b/base/shared/src/main/scala/scalaz/data/Forall.scala
@@ -37,7 +37,10 @@ trait ForallModule {
     implicit def unapply1[G[_], H[_]]: Unapply[∀[λ[α => G[H[α]]]]] { type F[A] = G[H[A]] } =
       new Unapply[∀[λ[α => G[H[α]]]]] { type F[A] = G[H[A]] }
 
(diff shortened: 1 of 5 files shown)


jdegoes:
This looks really amazing!

> which doesn't allocate any new objects (assuming zero-cost syntax that adds the .apply[T] method to ∀[Empty]).

Do we know to what extent this is true, or can be made true, e.g. with macros?

jdegoes:
@TomasMikula This looks great and should be part of our Scalaz 8 design guide. Thanks for your work on this!

PR #1458, opened 2017-09-09
Type-aligned data types.

TomasMikula:
- **`AMaybe[F[_, _], A, B]`:** Similar to `Maybe[F[A, B]]`, but the empty case witnesses type equality between `A` and `B`.
 - **`APair[F[_], G[_]]`:** Isomorphic to `(F[A], G[A]) forSome { type A }`, but more robust with respect to type inference.
 - **`AList[F[_, _], A, B]`:** Type-aligned list. Empty list witnesses type equality between `A` and `B`.
 - **`AList1[F[_, _], A, B]`:** Non-empty type-aligned list.
 - **`ACatenable1[F[_, _], A, B]`:** Non-empty type-aligned sequence with _O(1)_ `cons`, `snoc` and concatenation.

All sequence data types (`AList`, `AList1`, `ACatenable1`) are folded in balanced binary fashion (analogous to what @S11001001 did in #1022).

The [P∀scal compiler plugin](https://github.com/TomasMikula/pascal) was added to help with creation of some polymorphic values.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/ACatenable1.scala
+++ b/base/shared/src/main/scala/scalaz/data/ACatenable1.scala
@@ -1,0 +1,68 @@
+package scalaz
+package data
+
+import scala.annotation.tailrec
+import scalaz.typeclass.Compose
+
+/**
+ * Non-empty type-aligned sequence represented as a (non-balanced) binary tree,
+ * supporting O(1) addition to either end and O(1) concatenation.
+ */
+sealed abstract class ACatenable1[=>:[_, _], A, B] {
+  import ACatenable1._
+
+  def compose[Z](that: ACatenable1[=>:, Z, A]): ACatenable1[=>:, Z, B] =
+    Chain(that, this)
+
+  def <<<[Z](that: ACatenable1[=>:, Z, A]): ACatenable1[=>:, Z, B] =
+    this compose that
+
+  def >>>[C](that: ACatenable1[=>:, B, C]): ACatenable1[=>:, A, C] =
+    that compose this
+
+  def :+[C](f: B =>: C): ACatenable1[=>:, A, C] =
+    Chain(this, Lift(f))
+
+  def +:[Z](f: Z =>: A): ACatenable1[=>:, Z, B] =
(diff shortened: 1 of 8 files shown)


vmarquez:
This looks great!  It's a lot to take of code, so I'll continue looking it over this week.

TomasMikula:
I would say just parse through the signatures. The types make it almost impossible to get the implementation wrong. (Except that one method in `PreComposeBalancer` where I work with raw integers. 😱 )

The signatures should correspond to same-name methods on ordinary structures. For example

```scala
List[T] {
  def foldLeft[U](u: U)(φ: (U, T) => U): U
}

AList[F[_, _], A, B] {
  def foldLeft[G[_]](g: G[A])(φ: RightAction[G, F]): G[B]
  // with RightAction expanded, using pseudo syntax
  def foldLeft[G[_]](g: G[A])(φ: ∀α.∀β.((G[α], F[α, β]) => G[β])): G[B]
}
```

If in `AList#foldLeft` we set

```scala
type F[A, B] = T
type G[A] = U
```

we get exactly the `foldLeft` signature from `List`.

jdegoes:
This looks really great to me. We may want to standardize on some terminology later (`foldl` / `foldLeft`, etc.).

TomasMikula:
Yeah, there might be some discrepancies from standard signatures for the names. Didn't look too hard into that.

TomasMikula:
There will be an important optimization coming if/when scala/scala#6065 is merged:

Arguments to tail-recursive methods are currently wrapped in an `APair`, only to be unwrapped as the first thing in the method body:

```scala
@tailrec def go(p: APair[F, G]): T = {
  val (f, g) = (p._1, p._2)
  ...
}
```

This is because `scalac` currently refuses to optimize tailcalls with varying type arguments. This is precisely what the mentioned PR addresses.

TomasMikula:
The PR for parametric tailcalls has just been merged, so it will be available in Scala 2.12.4. But we won't be able to use them while we support 2.11. 😕

TomasMikula:
Oh, didn't realize that, cool!

PR #1461, opened 2017-09-12
Remove implicit Unapply instances that under -Ypartial-unification are covered by another case.

TomasMikula:
Remove implicit `Unapply` instances that under `-Ypartial-unification` are covered by another case.

Add tests that the cases covered by the removed instances are still covered.

Where the clients previously depended on an instance that is now removed, they will be required to enable `-Ypartial-unification`.

### DIFF
--- a/tests/src/test/scala/scalaz/UnapplyTest.scala
+++ b/tests/src/test/scala/scalaz/UnapplyTest.scala
@@ -3,6 +3,9 @@ package scalaz
 import Leibniz.===
 
 object UnapplyTest extends SpecLite {
+  def teq[A[_], B[X] >: A[X] <: A[X]]: Unit = ()
+  def teq2[A[_, _], B[X, Y] >: A[X, Y] <: A[X, Y]]: Unit = ()
+
   object unapply {
     val ue = Unapply[Monad, Int \/ String]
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/scalaz/UnapplyTest.scala
```diff
@@ -22,6 +22,102 @@ object UnapplyTest extends SpecLite {
     Unapply2[Arrow, Kleisli[NonEmptyList, Int, String]].TC: Arrow[Kleisli[NonEmptyList, ?, ?]]
   }
 
+  object unapply3 {
+    implicit def functorInstance[A]: Functor[λ[α => A]] = ???
+    implicitly[Unapply[Functor, String]]
```
S11001001: This can be written without the `implicitly[...]` wrapping, because `object Unapply` has an appropriate `apply`; see e.g. [`val O` in README](https://github.com/scalaz/scalaz/tree/v7.3.0-M16#standalone-type-class-usage).

Also, [testing `M` might be valuable](https://github.com/TomasMikula/scalaz/compare/Unapply...S11001001:Unapply).

TomasMikula: Thanks, I added your commit to this PR.

Issue #1452, opened 2017-09-04
Foldable.{toMap, groupBy}


fommil:
It would be useful to have something equivalent to stdlib `groupBy` on `Foldable`

Maybe a bit hacky, but we can do this from `splitBy` and `fromFoldableWithKey`, example

```scala
scala> ==>>.fromFoldableWithKey(IList("foo", "bar", "bar", "faz", "gaz", "baz").splitBy(_.charAt(0)))((a, b1, b2) => b1 |+| b2)
res: Char ==>> scalaz.NonEmptyList[String] = Bin(f,NonEmpty[faz,foo],Bin(b,NonEmpty[baz,bar,bar],Tip,Tip),Bin(g,NonEmpty[gaz],Tip,Tip))
```

perhaps with an intermediate `Foldable.toMap` taking the `Semigroup[B]`

vmarquez:
I have been thinking about this for ages. I originally thought we could solve this with a 'Buildable' type class, but @tonymorris and Ed thought that wasn't a real TypeClass.  They suggested we try something like Cons (in combination with Foldable). https://github.com/julien-truffaut/Monocle/blob/c0540739c6a71c834ca13b889a18e18f5eb6618e/core/shared/src/main/scala/monocle/function/Cons.scala

Unfortunately, I don't think that will let us use Set/Map.  I have to believe there are other optics or type classes that can abstract away the idea of "putting an A into an M" lawfully.  

@edmundnoble Thought that a Free Monoid structure might help here too, but I'm not sure I understand enough.

jdegoes:
See [purescript-unfoldable](https://github.com/purescript/purescript-unfoldable/blob/master/src/Data/Unfoldable.purs).

This is, perhaps, not exactly the right abstraction, but the idea would be to define a suitable catamorphism (`Foldable`) and anamorphism (`Unfoldable`), and combine them to form a `Collection` type class that allows you to implement all the methods that require the ability to tear down and rebuild a linear structure.

Then instances could specialize functions like `groupBy`, etc., in order to improve performance over the default implementations.

julien-truffaut:
A while ago I tried to experiment with `Foldable` / `Unfoldable` relationship in cats. Here is the thread for reference: https://github.com/typelevel/cats/pull/1132

Jacoby6000:
[...]

PR #1471, opened 2017-09-27
Make IList more of a first class citizen in scalaz

### LINKED ISSUE #1462: replace uses of List with IList
while going through the typeclass hierarchy I've noticed a lot of uses of `List` where `IList` would be more appropriate.

This is a source and binary breaking change, so needs to go into 7.3.x... but it's also very mechanical and low hanging fruit if somebody wanted to spend a lazy afternoon contributing to scalaz.

Jacoby6000:
resolves #1462 

In most places, I've replaced usages of `List` with `IList`, and in general I've made sure that all methods yielding `List` in scalaz also have an `IList` analogue. 

I may have missed some places, but I got all the instances that my search tool found.

There are still some tests to write for methods I added, just wanted to make sure I'm on the right track.

Also, it seems like the scalajs optimizer is getting a stackoverflow when compiling the ephemeral stream tests.  Aside from that, everything seems good.


jbgi:
Thanks!

One thing I noticed yesterday at work is that `IList.sortBy` first construct a `List` and delegate to its api for sorting. It seems a bit wasteful. It would be nice of we got a more efficient implementation (eg. by converting `IList` directly to an `[Immutable]Array`, sorting, and then converting back to an `IList`).

Could be worth checking that other operations that make use of `IList` are not delegating to the analogue api for `List`: it should be the other way around if we want IList to be first class.

jdegoes:
@Jacoby6000 Thanks Jacob! This has been bugging me for a while.

Jacoby6000:
Seems that the search I did was insufficient and I missed quite a few things. Working on that now... I'll also see if I can figure out why the tests pass for me but fail in travis.

@jbgi I saw that in a few places.. If I catch more before this is done, and an `IList` implementation would be trivial I'll throw it in here.

Jacoby6000:
I've gone back and updated this a bit. Let me know if anything needs to change.

jdegoes:
@jbgi Any other feedback or does this look good to merge?

core/src/main/scala/scalaz/Heap.scala
```diff
@@ -87,11 +87,15 @@ sealed abstract class Heap[A] {
 
   def toUnsortedList: List[A] = toUnsortedStream.toList
 
+  def toUnsortedIList: IList[A] = IList(toUnsortedStream: _*)
+
   def toStream: Stream[A] =
```
jbgi: I think it would be nice to the same approach as in `FingerTree`: use a `reduceTo` method to implement those `to*` methods. `reduceTo` would be:
```scala
  def reduceTo[M](implicit M: Reducer[A, M]): M = {
    @tailrec
    def go(acc: M, heap: Heap[A]): M = {
      heap.uncons match {
        case Some((a, as)) => go(M.cons(a, acc), as)
        case _ => acc
      }
    }
    go(M.zero, this)
  }
```

core/src/main/scala/scalaz/ListT.scala
```diff
@@ -42,32 +42,34 @@ final case class ListT[M[_], A](run: M[List[A]]){
...
+  def tail(implicit M: Functor[M]) : ListT[M, A] = new ListT(M.map(run)(_.tailOption.get))
+
+  def tailOption(implicit M: Functor[M]) : ListT[Lambda[a => M[Option[a]]], A] = new ListT[Lambda[a => M[Option[a]]], A](M.map(run)(_.tailOption))
 
   def foldLeft[B](z: => B)(f: (=> B, => A) => B)(implicit M: Functor[M]) : M[B] = M.map(run)(_.foldLeft(z){(left, right) => f(left, right)})
 
-  def toList : M[List[A]] = run
+  def toIList : M[IList[A]] = run
+
+  def toList(implicit M: Functor[M]): M[List[A]] = foldRight(List.empty[A])(_ :: _)
```
jbgi: I think `M.map(run)(_.toList)` would likely perform better.

core/src/main/scala/scalaz/Nondeterminism.scala
```diff
@@ -62,6 +62,9 @@ trait Nondeterminism[F[_]] extends Monad[F] { self =>
 
   def chooseAny[A](head: F[A], tail: Seq[F[A]]): F[(A, Seq[F[A]])]
 
+  // TODO: Reimplement chooseAny using a more general collections method and then get rid of the toList call here.
+  def chooseAny[A](head: F[A], tail: IList[F[A]]): F[(A, Seq[F[A]])] = chooseAny(head, tail.toList)
```
jbgi: I fear a bit for overloading ambiguities. Maybe we should leave this one out until the `TODO` is resolve (could https://hackage.haskell.org/package/ListLike-4.5.1/docs/Data-ListLike.html be something to consider?)

core/src/main/scala/scalaz/Heap.scala
jbgi: but, thinking of it, `toStream` implementation should be kept as is to preserve laziness. `reduceTo` would be used only for strict containers.

jbgi: oups, `reduceTo` would produce in reverse order. So better leave it like that, sorry.

core/src/main/scala/scalaz/Nondeterminism.scala
Jacoby6000: I (poorly) approached this a little bit in another branch mentioned in #1462.  @fommil pointed me to another issue (#1452) which, after some discussion, has some better ideas about the whole thing. 

I wanted to make the `IList` overload here to make `IList` more useful.  One of the reasons I've personally avoided `IList`, is that there are methods like this which work for the standard library but not IList, so  you wind up moving away from an `IList` anyway. 

After thinking about it a bit more, that's not even really a good argument for my overload above, since it winds up returning a `Seq`. I'll remove it for now, until #1452 has been looked in to.  I don't think `Foldable` + `Reducer` would help here.  I haven't looked in to it personally, but the rest of `Nondeterminism` is implemented in terms of those typeclasses.  I suspect the original author tried and found it not possible.

core/src/main/scala/scalaz/TheseT.scala
```diff
@@ -41,7 +41,7 @@ final case class TheseT[F[_], A, B](run: F[A \&/ B]) {
 
   def exists(p: B => Boolean)(implicit F: Functor[F]): F[Boolean] = F.map(run)(_ exists p)
   def forall(p: B => Boolean)(implicit F: Functor[F]): F[Boolean] = F.map(run)(_ forall p)
-  def toList(implicit F: Functor[F]): ListT[F, B] = ListT(F.map(run)(_.toList))
+  def toList(implicit F: Functor[F]): ListT[F, B] = ListT(F.map(run)(_.toIList))
```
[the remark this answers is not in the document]
Jacoby6000: Looking at it now, I'd actually prefer to name it `toListT`, and then make `toIList: F[IList[B]]` and `toLIst: F[List[B]]` Variants.

core/src/main/scala/scalaz/Heap.scala
Jacoby6000: I vaguely remember going through the same thought process here. 😅

PR #1472, opened 2017-09-27
Fix data type

TomasMikula:
Non-boxing implementation of `Fix`.

This is the part of #1455 (`IList` via `Fix`) which is independent of `IList` and useful on its own.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/Fix.scala
+++ b/base/shared/src/main/scala/scalaz/data/Fix.scala
@@ -1,0 +1,27 @@
+package scalaz
+package data
+
+import Prelude._
+import typeclass.Liskov
+import typeclass.Liskov.<~<
+
+trait FixModule {
+  type Fix[F[_]]
+
+  /* functions */
+
+  def   fix[F[_]](f: F[Fix[F]]):   Fix[F]
+  def unfix[F[_]](f:   Fix[F] ): F[Fix[F]]
+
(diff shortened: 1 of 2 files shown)


jdegoes:
Looks superb. I'm disappointed by the `invokevirtual` on the `Fix` module, but that's way less overhead than one additional allocation for every level of recursion.

Would be interesting to see if we can in any way [cut down on the overhead](https://github.com/slamdata/matryoshka/tree/master/core/shared/src/main/scala/matryoshka/data) of `Mu` and `Nu`.

TomasMikula:
Except for non-boxing representation of `Mu`, I don't see any obvious optimization opportunities.

japgolly:
FYI I just benchmarked this and I'm seeing approx 20-30% improvement in speed. Noice!

Code and one round of BM results here: https://github.com/japgolly/microlibs-scala/commit/c2c1f8adaa9c9166ed6c10a0c3c09cfdbc0c8d58
(Only look at the anamorphism results. Cata- and hylomorphisms roughly same cos they don't construct fixtypes.)

TomasMikula:
Thanks for benchmarking and sharing your results! 🎉

japgolly:
No worries. Thanks for this PR! Scalaz 8 is going to be killer!

jdegoes:
Awesome. Thanks for the benchmarking!

PR #1474, opened 2017-09-27
Simplify and optimize apN and applyN

TomasMikula:
- In `apN`, avoid currying a function in multiple steps, creating intermediate, intermediately curried functions.
 - In `applyN`, avoid constructing and deconstructing intermediate tuples.

As a bonus, the implementation is also shorter and more regular.


jdegoes:
👍 🙏 This is definitely better. A performant implementation should override all if possible, but these are quite sensible defaults.

fommil:
I'm not sure if I agree that going via `ap` is faster... see #1903 to pick up the discussion again.

PR #1479, opened 2017-10-08
Unambiguous ops (syntax extensions) for scalaz 7.

TomasMikula:
The goal is to allow the user to structure imports so that ambiguous syntactic extensions are avoided.

Example:

```scala
import scalaz.syntax.monad._

def foo[F[_]: Monad: Traverse, A, B](fa: F[A], f: A => B): F[B] =
  fa map f // no ambiguous implicits, even though there are two implicit Functor instances
```

Typeclass encoding is left untouched, thus ambiguous typeclass instances are not avoided, just ambiguous ops. As such, this approach is less ambitious than scalaz8's, but maybe could be considered for scalaz7? On one hand, it can have great benefits for the client code, but on the other hand it increases the complexity of syntax extensions. I'm myself not yet 100% convinced that this is the better tradeoff.

The idea is that the import

```scala
import scalaz.syntax.traverse0._
```
imports syntax for `Traverse`, but doesn't include syntax for `Functor` or `Foldable`, whereas the (usual) import

```scala
import scalaz.syntax.traverse._
```

also brings in syntax for `Functor` and `Foldable`.
There is one change, though: for this import to provide `Functor` (or `Foldable`) syntax, there needs to be a `Traverse` instance in scope; `Functor` (or `Foldable`) instance will not suffice, even when only `Functor` (or `Foldable`) operations are performed. (This is why there is no ambiguity in the example at the top - for `Functor` ops imported via `scalaz.syntax.monad._`, a `Monad` instance is required, and there is only one.)

This allows us to import

```scala
import scalaz.syntax.monad._
import scalaz.syntax.traverse0._
```

and then use both `Monad` and `Traverse` syntax extensions without ambiguity.

For some examples, see the diff of `SyntaxUsage.scala`.

### Implementation comments

Where previously we had

```scala
trait ToTraverseOps extends ToFunctorOps with ToFoldableOps
```

we now have (to be refined shortly)

```scala
trait ToTraverseOps0

trait ToTraverseOps extends ToTraverseOps0 with ToFunctorOps with ToFoldableOps
```

That is, we provide `ToTraverseOps0` which doesn't inherit `Functor` and `Foldable` ops.

In order for `ToTraverseOps`'s inherited `ToFunctorOps` and `ToFoldableOps` to actually require a `Traverse` instance even for `Functor` or `Foldable` operations, we pass the required typeclass as a type argument. Thus we have

```scala
trait ToTraverseOps0[TC[F[_]] <: Traverse[F]]

trait ToTraverseOps[TC[F[_]] <: Traverse[F]] extends ToTraverseOps0[TC] with ToFunctorOps[TC] with ToFoldableOps[TC]
```

and then in `ToFunctorOps`, we require an instance of `TC` (e.g. `Traverse`) instead of `Functor`:

```scala
trait ToFunctorOps0[TC[F[_]] <: Functor[F]] {
  implicit def ToFunctorOps[F[_], A](v: F[A])(implicit F0: TC[F]) = ???
}
```

There is one more point worth mentioning.
Consider syntax for `MonadTell` extending syntax for `Monad`, using the above scheme:

```scala
trait ToMonadOps[TC[F[_]] <: Monad[F]]

trait ToMonadTellOps[TC[F[_], S] <: MonadTell[F, S]] extends ToMonadOps[???]
```

The kinds of type parameters of `ToMonadOps` and `ToMonadTellOps` do not match. Therefore, we cannot just pass the `TC` type parameter from `ToMonadTellOps` to `ToMonadOps`. The solution is to use an existential

```scala
trait ToMonadTellOps[TC[F[_], S] <: MonadTell[F, S]] extends ToMonadOps[λ[F[_] => TC[F, S] forSome { type S }]]
```

and, despite many problems with existentials in Scala, the compiler does consider `MonadTell[F, S] forSome { type S }` to be a subtype of `Monad[F]`.

### DIFF
--- a/core/src/main/scala/scalaz/syntax/AlignSyntax.scala
+++ b/core/src/main/scala/scalaz/syntax/AlignSyntax.scala
@@ -18,20 +18,22 @@ final class AlignOps[F[_],A] private[syntax](val self: F[A])(implicit val F: Ali
   ////
 }
 
-sealed trait ToAlignOps0 {
-  implicit def ToAlignOpsUnapply[FA](v: FA)(implicit F0: Unapply[Align, FA]) =
+sealed trait ToAlignOpsU[TC[F[_]] <: Align[F]] {
+  implicit def ToAlignOpsUnapply[FA](v: FA)(implicit F0: Unapply[TC, FA]) =
     new AlignOps[F0.M,F0.A](F0(v))(F0.TC)
 
 }
 
-trait ToAlignOps extends ToAlignOps0 with ToFunctorOps {
-  implicit def ToAlignOps[F[_],A](v: F[A])(implicit F0: Align[F]) =
+trait ToAlignOps0[TC[F[_]] <: Align[F]] extends ToAlignOpsU[TC] {
+  implicit def ToAlignOps[F[_],A](v: F[A])(implicit F0: TC[F]) =
     new AlignOps[F,A](v)
 
   ////
 
   ////
 }
+
+trait ToAlignOps[TC[F[_]] <: Align[F]] extends ToAlignOps0[TC] with ToFunctorOps[TC]
 
 trait AlignSyntax[F[_]] extends FunctorSyntax[F] {
   implicit def ToAlignOps[A](v: F[A]): AlignOps[F, A] = new AlignOps[F,A](v)(AlignSyntax.this.F)
--- a/core/src/main/scala/scalaz/syntax/ApplicativeBuilder.scala
+++ b/core/src/main/scala/scalaz/syntax/ApplicativeBuilder.scala
@@ -2,145 +2,102 @@ package scalaz
 package syntax
 
 /** @see [[scalaz.syntax.ApplyOps]]`#|@|` */
-private[scalaz] trait ApplicativeBuilder[M[_], A, B] {
-  val a: M[A]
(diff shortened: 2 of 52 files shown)


TomasMikula:
Might be of interest to  @adelbertc.

TomasMikula:
Failing tasks are

```
coreJVM/*:checkGenTypeClasses
effectJVM/*:checkGenTypeClasses
```

which I don't understand.

xuwei-k:
> Failing tasks are
>
> ```
> coreJVM/*:checkGenTypeClasses
> effectJVM/*:checkGenTypeClasses
> ```
>
> which I don't understand.

`${TypeClass}.scala` and `${TypeClass}Syntax.scala` are partially auto generated from [`project/GenTypeClasses.scala`](https://github.com/scalaz/scalaz/blob/645fdf81a458d6cb24937ec210e9d753b3dce1b8/project/GenTypeCla.scala).
The `checkGenTypeClasses` task validate consistency of `GenTypeClass.scala` and generated sources. We should update `GenTypeClass.scala`.


https://github.com/xuwei-k/scalaz/commit/4b214ffd575c45f11e325f3d799b0bed8934b147

TomasMikula:
Thanks, Kenji!

PR #1485, opened 2017-10-12
Generalize unfold to all Reducers.

jbgi:


core/src/main/scala/scalaz/Reducer.scala
```diff
@@ -86,7 +107,8 @@ sealed abstract class ReducerInstances {
   /** Collect `C`s into a stream, in order. */
   implicit def StreamReducer[C]: Reducer[C, Stream[C]] = {
     import std.stream._
-    unitConsReducer(Stream(_), c => c #:: _)
+    import Stream._
+    unitLazyConsReducer(cons(_, empty): Stream[C], cons(_, _))
```
TomasMikula: If this enables more laziness than before, maybe it's worthwhile to add a test case demonstrating that.

jbgi: Yeah, need to add a `ReducerSpec`.
Note that the laziness is only used for `unfoldr` not for existing methods.

TomasMikula: Oh, right.

PR #1486, opened 2017-10-14
correct type in the example CABRunLengthEncoder

ashkan-leo:
change 1C5CB to 1C5AB


example/src/main/scala/scalaz/example/ReaderWriterStateTUsage.scala
```diff
@@ -51,7 +51,7 @@ object CABRunLengthEncoder {
    "CAAAAAB."
    "C5AB."
    "C3AAAB."
-   "1C5CB."
+   "1C5CA."
```
jbgi: Shouldn't it rather be `"1C5AB."` (or `"1C5A1B."`)?

xuwei-k:
fixed https://github.com/scalaz/scalaz/commit/8ca63d6a3d4ae1d059239888133dcffa0cd9d441

PR #1494, opened 2017-10-20
 Enable -Xsource:2.13 and resolve subsequent ambiguous implicits

TomasMikula:
Taking advantage of scala/scala#6069

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/Forall.scala
+++ b/base/shared/src/main/scala/scalaz/data/Forall.scala
@@ -48,15 +48,6 @@ trait ForallModule {
   object Unapply {
     implicit def unapply[G[_]]: Unapply[∀[G]] { type F[A] = G[A] } =
       new Unapply[∀[G]] { type F[A] = G[A] }
-
-    implicit def unapply1[G[_], H[_]]: Unapply[∀[λ[α => G[H[α]]]]] { type F[A] = G[H[A]] } =
-      new Unapply[∀[λ[α => G[H[α]]]]] { type F[A] = G[H[A]] }
-
-    implicit def unapply2[P[_, _]]: Unapply[∀[λ[α => P[α, α]]]] { type F[A] = P[A, A] } =
-      new Unapply[∀[λ[α => P[α, α]]]] { type F[A] = P[A, A] }
-
-    implicit def unapply3[P[_, _], G[_], H[_]]: Unapply[∀[λ[α => P[G[α], H[α]]]]] { type F[A] = P[G[A], H[A]] } =
-      new Unapply[∀[λ[α => P[G[α], H[α]]]]] { type F[A] = P[G[A], H[A]] }
   }
 }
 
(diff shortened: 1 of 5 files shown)


base/shared/src/main/scala/scalaz/data/Forall.scala
```diff
@@ -48,26 +48,13 @@ trait ForallModule {
   object Unapply {
     implicit def unapply[G[_]]: Unapply[∀[G]] { type F[A] = G[A] } =
```
jdegoes: Does this subsume all the deleted cases?

base/shared/src/main/scala/scalaz/data/Forall.scala
```diff
@@ -48,26 +48,13 @@ trait ForallModule {
...
-
-    implicit def unapply3[P[_, _], G[_], H[_]]: Unapply[∀[λ[α => P[G[α], H[α]]]]] { type F[A] = P[G[A], H[A]] } =
-      new Unapply[∀[λ[α => P[G[α], H[α]]]]] { type F[A] = P[G[A], H[A]] }
   }
 }
 
 trait ForallSyntax {
   import ForallSyntax._
 
   implicit def toForallOps[F[_]](a: ∀[F]): Ops[F] = new Ops[F](a)
```
jdegoes: Ditto.

base/shared/src/main/scala/scalaz/data/Forall2.scala
```diff
@@ -35,26 +35,13 @@ trait Forall2Module {
   object Unapply {
     implicit def unapply[G[_, _]]: Unapply[∀∀[G]] { type F[A, B] = G[A, B] } =
       new Unapply[∀∀[G]] { type F[A, B] = G[A, B] }
```
jdegoes: Same here.

base/shared/src/main/scala/scalaz/data/Forall.scala
TomasMikula: Yes. We have [examples](https://github.com/scalaz/scalaz/blob/766c09ffc0f6ef243da94868fcd61a5c892d7a19/example/shared/src/main/scala/scalaz/example/ForallUsage.scala) that previously needed those other.

base/shared/src/main/scala/scalaz/data/Forall.scala
TomasMikula: Likewise, plus scalac now contains some very similar [test cases](https://github.com/scala/scala/blob/c2a5883891a68180b143eb462c8b0cebc8d3b021/test/files/run/hk-typevar-unification.scala) for these. 😈

base/shared/src/main/scala/scalaz/data/Forall2.scala
TomasMikula: Yup.

base/shared/src/main/scala/scalaz/data/Forall.scala
jdegoes: ❤️ ❤️ ❤️

Issue #1495, opened 2017-10-20
Show String Interpolator Instance


ChristopherDavenport:
Was wondering if there was a reason not to have a string interpolator instance for `Show`. Currently for safety writing `s"${f.shows}"` and it seems that a safe interpolator would be ideal. 

What are your thoughts? I'll happily draft the PR, but if there was a reason you were avoiding it I won't make it.

TomasMikula:
👍

jdegoes:
If it's type safe and runtime safe and makes users lives easier, I'm all for it.

ritschwumm:
yes please - and then disable the s-interpolator in wartremover

ChristopherDavenport:
Closed via #1496

jdegoes:
@christopherdavenport Thanks for the contribution!

PR #1496, opened 2017-10-20
Add Show Interpolator

ChristopherDavenport:
Having difficulty with the final materialization for sealed traits, as it doesn't find the Show.Shows, which I think it should. 

It's probably something very simple.

Otherwise works as expected.

### DIFF
--- a/core/src/main/scala/scalaz/Show.scala
+++ b/core/src/main/scala/scalaz/Show.scala
@@ -50,5 +50,6 @@ object Show {
   final case class ShowInterpolator(sc: StringContext) extends AnyVal {
     def show(args: Shows*): String = sc.s(args: _*)
   }
+  ////
 
 }
--- a/core/src/main/scala/scalaz/Show.scala
+++ b/core/src/main/scala/scalaz/Show.scala
@@ -6,7 +6,7 @@ package scalaz
  * [[scalaz.Cord]] for efficiency.
  */
 ////
-trait Show[F] extends Show.ContravariantShow[F]  { self =>
+trait Show[F]  { self =>
   ////
   def show(f: F): Cord = Cord(shows(f))
   def shows(f: F): String = show(f).toString
@@ -42,18 +42,13 @@ object Show {
     }
   }
 
-  trait ContravariantShow[-T] {
-    def shows(t: T): String
-  }
-
   final case class Shows(override val toString: String) extends AnyVal
   object Shows {
-    implicit def mat[A](x: A)(implicit S: ContravariantShow[A]): Shows = Shows(S.shows(x))
+    implicit def mat[A](x: A)(implicit S: Show[A]): Shows = Shows(S.shows(x))
   }
 
   final case class ShowInterpolator(sc: StringContext) extends AnyVal {
     def show(args: Shows*): String = sc.s(args: _*)
   }
   ////
-
 }
--- a/core/src/main/scala/scalaz/Show.scala
+++ b/core/src/main/scala/scalaz/Show.scala
@@ -42,5 +42,13 @@ object Show {
     }
   }
 
-  ////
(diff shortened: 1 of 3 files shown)


ritschwumm:
i don't think you need the implicit conversion to Shows -
why not splice in the calls to .show directly?

ritschwumm:
funny, i had imagined something quite a bit more complicated:
https://github.com/ritschwumm/show-interpolator/blob/master/src/main/scala/ShowMacros.scala?ts=4
you don't need an implicit conversion and an intermediate data type - otoh, it requires macros.

ChristopherDavenport:
Refs #1495

core/src/main/scala/scalaz/Show.scala
```diff
@@ -19,6 +19,19 @@ trait Show[F]  { self =>
 object Show {
   @inline def apply[F](implicit F: Show[F]): Show[F] = F
 
+  trait ContravariantShow[-T] {
+    def shows(t: T): String
+  }
+
+  case class Shows(override val toString: String) extends AnyVal
```
jdegoes: `final case class ...`

core/src/main/scala/scalaz/Show.scala
```diff
@@ -6,7 +6,7 @@ package scalaz
  * [[scalaz.Cord]] for efficiency.
  */
 ////
-trait Show[F]  { self =>
+trait Show[F] extends Show.ContravariantShow[F]  { self =>
```
jdegoes: Why is `ContravariantShow` needed?

jdegoes:
Looks like some [compilation errors in the test](https://travis-ci.org/scalaz/scalaz/jobs/290638592).

core/src/main/scala/scalaz/Show.scala
ChristopherDavenport: Without that it infers type `Result1T` not `CopructT` and cannot reach the Shows instance. Should only apply in the string interpolator context. As `Show` is still invariant.

ChristopherDavenport:
I understand that. I'm not seeing the same errors locally with the exact commands in the `.travis.yml`. I'll check again if I'm doing something wrong.

ChristopherDavenport:
`checkGenTypeClasses` regenerates the file, which in this class then lacks the trait extension, it also fails as it doesn't align to what it was previously, as I believe is done [HERE](https://github.com/scalaz/scalaz/blob/89d812642838d90af4c02304ba8ef054f5c95b65/project/build.scala#L160)

So I'm in a battle with the code-generator as the code as written works.

core/src/main/scala/scalaz/Show.scala
TomasMikula: I would say that's an orthogonal problem, not limited to just `Show`. Don't try to solve it in this PR.

TomasMikula:
@ritschwumm I, too, originally thought that it would require macros. After being shown otherwise (by @dwijnand), I now prefer the non-macro implementation, mainly for better readability and lower cost of maintenance.

core/src/main/scala/scalaz/Show.scala
ChristopherDavenport: So leave test pending and remove the `ContravariantShow`?

TomasMikula: Remove `ContravariantShow`. Either remove the offending test, or add a type annotation: `val r: CoproductT = Result1T`.

jdegoes:
@ChristopherDavenport Looks great, any more changes on your end, or you ready for a merge?

ChristopherDavenport:
@jdegoes  For an initial build it looks good to me. I'll hope we circle around on the contravariant issue shortly, but this looks great in addressing an initial show interpolator.

PR #1502, opened 2017-10-27
Disjunction.fromTryCatchNonFatal

NeQuissimus:
`\/.fromTryCatchNonFatal` is a life-saver when integrating with Java libraries.
This is adding it to `DisjunctionFunctions` in Scalaz 8

I am also adding example usages for `Disjunction`, I assume these are going to be turned into/serve as documentation in the future.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/DisjunctionFunctions.scala
+++ b/base/shared/src/main/scala/scalaz/data/DisjunctionFunctions.scala
@@ -1,6 +1,8 @@
 package scalaz
 package data
 
+import scala.util.control.NonFatal
+
 import Disjunction.{\/, \/-, -\/}
 
 trait DisjunctionFunctions {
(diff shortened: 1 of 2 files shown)


jbgi:
personally I'd rather we keep only `toDisjunction` from `TryFunctions`/`TryOps`, and let `\/` clean from non-RT concepts like `try-catch`.

jdegoes:
@NeQuissimus 

Side-effecting functions in Java-land would normally be wrapped in `IO.sync`, which would also capture exceptions and allow you to access exceptions as disjunctions.

Are there lots of Java functions you want to wrap whose only side-effect is throwing exceptions?

NeQuissimus:
@jdegoes There are... TBH we usually pick and choose Scalaz functionality, right now we use monix Task as our "IO", so wrapping in IO.sync would just add another layer I don't really need.
I do understand keeping things clean is important and I can live with that (or writing my own implicit helper).
I just started poking through Scalaz 8 trying to find the things most important to me :)

jdegoes:
@NeQuissimus It's possible this could show up in a `Catchable` instance. The (operational) law would have to be something like, `unexceptional(throw e) == point(-\/(e))`.

NeQuissimus:
@jdegoes yes, that would be wonderful :) I don't really mind where this goes as long as I can easily reproduce the `fromTryCatchNonFatal` behaviour. Most Java libraries we integrate with I simply don't trust, so I wrap things in disjunctions (NullPointerException, IllegalArgumentException etc. that no documentation declares...) a lot of times.

PR #1508, opened 2017-10-30
Remove NonEmptyIList

### LINKED ISSUE #1507: what is the point of NonEmptyIList?
```scala
  type NonEmptyIList[A] = OneAnd[IList, A]
```

given that `NonEmptyList` exists, is there any point in this alias? It only confuses the issue by encouraging the use of two things for the same purpose.

puffnfresh:
NonEmptyList should be used instead. Having a type alias which provides less
behaviour is a bit confusing.

Fixes #1507


tonymorris:
:+1:

Also, ideally, there'd be `Iso` for `NonEmptyList` <-> `OneAnd[List]`.

Issue #1396, opened 2017-05-30
No `MonadError` instance for `StateT`?


fosskers:
I'm trying to raise an error in some code that is running in `StateT[EitherT]`. I have:

```scala
type EitherStr2[T] = scalaz.EitherT[scalaz.Id.Id, String, T]
type Statey2[T] = scalaz.StateT[({ type λ[α] = scalaz.EitherT[scalaz.Id.Id, String, α] })#λ, Int, T]
```
and I can manage
```scala
val foo: EitherStr2[Int] = scalaz.MonadError[EitherStr2, String].raiseError("crap")
```
but not
```scala
val bar: Statey2[Int] = scalaz.MonadError[Statey2, String].raiseError("noo")
```

I'm not sure what I'm missing. Should I take `foo` and try and lift it into `StateT`? I couldn't find examples of doing something like that, so I'm not sure what to do.

Thanks.

TomasMikula:
Seems like such instance is indeed missing. Something similar to [what we have for `MonadState`](https://github.com/scalaz/scalaz/blob/6da7c79ce821f7002cc414dc6e7bb5d0c75fe96b/core/src/main/scala/scalaz/MonadState.scala#L30-L42) could be added for `MonadError`, i.e. if there is `MonadError` for `F`, then provide `MonadError` for any "bigger" monad `G`. Do you want to give it a try and open a pull request?

jvz:
I actually wrote some glue code for this, but I wasn't sure if it wasn't included in scalaz because it violated some monad laws.

```scala
  implicit def stateTMonadError[F[_], S, E](implicit F: MonadError[F, E])
  : MonadError[StateT[F, S, ?], E] = new MonadError[StateT[F, S, ?], E] {
    override def raiseError[A](e: E): StateT[F, S, A] =
      StateT(_ => F.raiseError(e))

    override def handleError[A](fa: StateT[F, S, A])(f: (E) => StateT[F, S, A]): StateT[F, S, A] =
      StateT(s => F.handleError(fa(s))(f(_)(s)))

    override def bind[A, B](fa: StateT[F, S, A])(f: (A) => StateT[F, S, B]): StateT[F, S, B] =
      fa flatMap f

    override def point[A](a: => A): StateT[F, S, A] =
      StateT(s => (s, a).point[F])
  }

  implicit def rwstMonadError[F[_], R, W, S, E](implicit F: MonadError[F, E], W: Monoid[W])
[...]

PR #1509, opened 2017-11-01
Add MonadError instances for StateT and RWST

jvz:
This adds MonadError instances given a MonadError instance for StateT
and ReaderWriterStateT.

This implements #1396.

### DIFF
--- a/core/src/main/scala/scalaz/ReaderWriterStateT.scala
+++ b/core/src/main/scala/scalaz/ReaderWriterStateT.scala
@@ -93,6 +93,12 @@ sealed abstract class IndexedReaderWriterStateTInstances extends IndexedReaderWr
       def A = F1
       def W = W0
     }
+
+  implicit def rwstMonadError[F[_], R, W, S, E](implicit F0: MonadError[F, E], W0: Monoid[W]): MonadError[ReaderWriterStateT[F, R, W, S, ?], E] =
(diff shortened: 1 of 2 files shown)


jvz:
Let me know if I need to add any property checks, laws, etc. This PR is based on the type class instances I've been using in my own project.

jdegoes:
@jvz Thank you for this work!

The changes look very safe. The main concern I'd have would be around whether or not these instances overlap with anything else in Scalaz. Have you looked into that?

jvz:
The only conflicts I know of are in regards to monad priorities due to inheritance issues (something I started avoiding doing in my own FP practice by just re-adding functions from Functor to Monad for example instead of extending it), and I believe this uses an appropriate scope to avoid ambiguous implicits.

My local use of this was similar to how I mentioned in #1396, and I haven't had any ambiguous implicit issues there (which includes several files doing `import scalaz._, Scalaz._` from before I learned more about the import hierarchy).

puffnfresh:
I would have liked to see some instance resolution compilation tests. For example:

https://github.com/scalaz/scalaz/blob/2c4b35a4379c8de7dd220620b21089bc0979848a/tests/src/test/scala/scalaz/MaybeTTest.scala#L38

jdegoes:
@puffnfresh Good idea.

PR #1511, opened 2017-11-13
 Add Zap instance for pairs of Zaps

japgolly:


jdegoes:
This looks nice. It'd be great to have an example added to the examples that shows Scala can actually locate this instance when required.

emilypi:
If this builds, I'll merge

japgolly:
Ah yeah thanks. The intent to add an example is there but realistically
I'll probably never get around to it. Even this PR looks trivial but the
same thing seemed profound when I saw it expressed as pairing rather than
zap.

emilypi:
@japgolly do you want me to just close it then? I'm good _without_ an example

japgolly:
If you're good to merge, I think it should be merged. Lots of things in
scalaz are useful without being profound :)

PR #1519, opened 2017-11-16
Scalaz 8 IO pull request!

jdegoes:
This is a WIP pull request for the Scalaz 8 IO data type and JVM-based runtime system. This is intended to be a wholesale replacement for Scalaz 7 `IO` and `Task`.

This pull request includes the following components:

 * `scalaz.effect.IO` &mdash; The core `IO` data type, which models synchronous and asynchronous effects, with a concurrency model built on fibers and primitives and semantics similar to Haskell's `Async` package.
 * `scalaz.effect.RTS` &mdash; The runtime system for the JVM, which is capable of effectfully and efficiently interpreting an `IO` program.
 * `scalaz.effect.SafeApp` &mdash; The entry point for pure functional programs, which is integrated into the runtime system.
 * `scalaz.effect.IORef` &mdash; A mutable reference similar to Haskell's `IORef` data type.
 * `scalaz.effect.MVar` &mdash; A concurrent variable similar to Haskell's `MVar` data type.
 * Miscellaneous other utility data types, such as `BracketResult`, `AsyncReturn`, and so on.

In addition, a new `benchmarks` project has been added with competitive benchmarks (Scala `Future`, cats-effect `IO`, and Monix `Task`), which are quite useful for auditing and optimizing the performance of Scalaz 8 `IO`.

The following work remains outstanding:

 * **Type Class Hierarchy** &mdash; Some work has been done formalizing algebraic and operational laws in a type class hierarchy (`MonadFork`, `Catchable`, etc.), but there is more work to do in fleshing out the design.
 * **Unit Tests** &mdash; The JVM runtime system for `IO` has a growing test suite but many more tests are required, especially for concurrent primitives. Bugs are expected, but at this stage, most should be straightforward to fix. `MVar` has no test suite and should be considered a sketch.
 * **Performance** &mdash; Performance is very good in most cases, but now that the design is settled, optimization can yield additional improvements, especially in edge cases. Some code has not been optimized for several major iterations and several scalac compiler versions, so additional low-hanging fruit may exist.
 * **Concurrency** &mdash; The current cooperative yielding model is too primitive and thread pool perfunctory, with inadequate control for those needing more powerful knobs for thread shifting. Both of these will be greatly improved in a newer design.
 * **Javascript RTS** &mdash; There currently exists no Javascript runtime system, though almost identical semantics have already been implemented in PureScript so there is an existence proof.
 * **Benchmarks** &mdash; Additional benchmarks would be useful to cover mixed load and purely asynchronous load use cases, as well as benchmarks for `MVar` when the implementation has been tested and stabilized.
 * **Code Cleanup** &mdash; The runtime system implementation is intended to be low-level, and some duplication is necessary to minimize heap allocations. However, the code can benefit from better organization and some strategic simplifications.
 
While there is more work that can and must be done, the code has reached a sufficient level of completeness and maturity that it can benefit from outside feedback and collaborators. So with the goal of encouraging this feedback and opening the door to third-party contributions, I am quite happy to present this WIP pull request for Scalaz 8 IO!

Thanks to all those whose feedback and work in other projects inspired and enabled my work here, including Vincent Marquez for the encouragement to work on this project, Alex Nedelscu for the inspiration to work on performance, Daniel Spiewak for his work on a type class hierarchy for effects, and many others for feedback around error handling.

Please dig in, break stuff, and provide whatever feedback or contributions you are inspired to provide!


rabbitonweb:
Finally it landed! Now what? :)

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -0,0 +1,547 @@
...
+ * by a given immutable `IO` value. For more information on interpreting `IO`
+ * values, see the default interpreter in `RTS` or the safe main function in
+ * `SafeApp`.
+ */
+sealed abstract class IO[A] { self =>
+  /**
+   * Maps an `IO[A]` into an `IO[B]` by applying the specified `A => B` function
+   * to the output of this action.
+   */
+  final def map[B](f: A => B): IO[B] = (self.tag : @switch) match {
```
edmundnoble: I think a note to the effect of "`map` is not always stack-safe" may be helpful.

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -0,0 +1,547 @@
...
+   * Delays this action by the specified amount of time.
+   */
+  final def delay(duration: Duration): IO[A] =
+    IO.sleep(duration) *> self
+
+  /**
+   * An integer that identifies the term in the `IO` sum type to which this
+   * instance belongs (e.g. `IO.Tags.Point`).
+   */
+  def tag: Int
```
edmundnoble: Should this be public?

effect/jvm/src/main/scala/scalaz/effect/RTS.scala
```diff
@@ -0,0 +1,1058 @@
...
+                result = \/-(value)
+              }
+
+            case IO.Tags.SyncEffect =>
+              val io = curIo.asInstanceOf[IO.SyncEffect[Any]]
+
+              try {
+                val value = io.effect()
+
+                curIo = nextInstr(io.effect(), stack)
```
ataraxer: Reevaluation of effect, should be `nextInstr(value, stack)`.

It might be a good idea to add a test which composes a somewhat complicated IO, which increments a global counter on each evaluation and tests that it ends up being equal to sum of effectively evaluated effects.

effect/shared/src/main/scala/scalaz/effect/MVar.scala
```diff
@@ -0,0 +1,68 @@
...
+ * thought of as a two-sided blocking queue of capacity 1, which blocks on takes
+ * when there is nothing to take, and blocks on puts when there is already
+ * something there.
+ *
+ * `MVar` values are asynchronous and do not actually block.
+ *
+ * {{{
+ * for {
+ *   v <- IO.emptyMVar[Int] // A new empty `MVar` that holds `Int`
+ *   _ <- v.put(3).fork     // In a separate fiber, put 4 into the `MVar`
```
japgolly: typo: s/4/3/

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -0,0 +1,547 @@
...
+}
+
+object IO extends IOInstances {
+  object Tags {
```
pedrofurla: in light of @edmundnoble 's question on https://github.com/scalaz/scalaz/pull/1519/files#r151562686 I wonder if these should be exposed too.

benchmarks/src/main/scala/scalaz/effect/IONarrowFlatMapBenchmark.scala
```diff
@@ -0,0 +1,68 @@
...
+  def scalazNarrowFlatMap(): Int = {
+    def loop(i: Int): IO[Int] =
+      if (i < size) IO(i + 1).flatMap(loop)
+      else IO(i)
+
+    unsafePerformIO(IO(0).flatMap(loop))
+  }
+
+  @Benchmark
+  def catsNarrowFlatMap(): BigInt = {
```
clhodapp: This is the only `BigInt`. I think it'll result in an implicit conversion being invoked, which makes this test a bit unfair.

benchmarks/src/main/scala/scalaz/effect/IOLeftBindBenchmark.scala
```diff
@@ -0,0 +1,76 @@
...
+    def loop(i: Int): IO[Int] =
+      if (i % depth == 0) IO(i + 1).flatMap(loop)
+      else if (i < size) loop(i + 1).flatMap(i => IO(i))
+      else IO(i)
+
+    unsafePerformIO(IO(0).flatMap(loop))
+  }
+
+  @Benchmark
+  def catsLeftBindBenchmark(): BigInt = {
```
clhodapp: Again, this is the only `BigInt`

effect/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Pure Values
+
+You can lift pure values into `IO` with `IO.apply`:
+
+```scala
+val liftedString: IO[String] = IO("Hello World")
+```
+
+The constructor uses non-strict evaluation, so the parameter will not be evaluated until when and if the `IO` action is executed at runtime.
```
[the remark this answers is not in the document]
jdegoes: Poor wording on my part. It never results in duplicated computation. If you use it for effects, the results are undefined, and could include runtime errors (e.g. `IO(throw e).map(_ => 1)`).

The name of this could be `IO.point`. But then I'd want to delete `IO.apply` because it is abused for effectful code but that abuse is not supported by `IO`.

`point` may never be used for effects, the only question is whether or not `IO.apply` should be used for `point` (as in 7), or whether it should not be used at all to force users to be explicit.

effect/shared/src/main/scala/scalaz/effect/IO.scala
jdegoes: Added note. It's stack safe to > 10k repetitions on the JVM (probably much more).

effect/shared/src/main/scala/scalaz/effect/IO.scala
jdegoes: I could go either way on this. There's no unsafety in this API, but there's also no utility in the API unless you're going to write your own interpreter.

effect/shared/src/main/scala/scalaz/effect/MVar.scala
jdegoes: Thanks, fixed!

effect/shared/src/main/scala/scalaz/effect/IO.scala
jdegoes: Yes, I think if we hide `tag` we should hide all of these, too.

benchmarks/src/main/scala/scalaz/effect/IONarrowFlatMapBenchmark.scala
jdegoes: Oops. Copy paste error. It's outside the main loop though so won't really affect the benchmark.

benchmarks/src/main/scala/scalaz/effect/IOLeftBindBenchmark.scala
jdegoes: Fixed. Thank you! Did you see any other ones like this?

effect/README.md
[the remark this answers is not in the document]
edmundnoble: I think it's best to just remove `IO.apply` and have it named maybe `IO.point` or `IO.lazyPure` or something.

benchmarks/src/main/scala/scalaz/effect/IOLeftBindBenchmark.scala
clhodapp: Cool. I only saw the two where it was inconsistent.

puffnfresh:
How hard would this be to backport to Scalaz 7? Can it exist as a library?

effect/jvm/src/main/scala/scalaz/effect/MVarImpl.scala
```diff
@@ -0,0 +1,284 @@
...
+            Surplus(value, putters.tail)
+          }
+        case Deficit(takers, readers) =>
+          finish = () => AsyncReturn.maybeLater(removeTaker(taker, _))
+
+          Deficit(takers :+ taker, readers)
+      }
+
+      if (state.compareAndSet(oldState, newState)) loop = false
+    }
```
TomasMikula: Scala does have `do while` loop. 😉 
Here and in several other places below `var loop` can be replaced with a `do while` loop.

TomasMikula:
Looking great! Also 👍  for the documentation!

Btw., how do you reconcile having a _data structure_ for I/O with your opinion that [Data Structures Are Antithetical to Functional Programming](http://degoes.net/articles/kill-data)? 😉

effect/jvm/src/main/scala/scalaz/effect/console.scala
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2017 John A. De Goes. All rights reserved.
+package scalaz.effect
+
+package object console {
```
jbgi: should this be `package object`  or rather a normal `object`?

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -0,0 +1,549 @@
...
+   *
+   * {{{
+   * openFile("data.json").catchSome {
+   *   case FileNotFoundException(_) => openFile("backup.json")
+   * }
+   * }}}
+   */
+  final def catchSome(pf: PartialFunction[Throwable, IO[A]]): IO[A] = {
+    def tryRescue(t: Throwable): IO[A] = t match {
+      case t if pf.isDefinedAt(t) => pf(t)
```
jbgi: could be a simple `if`/`else` instead of a guard.

effect/jvm/src/main/scala/scalaz/effect/RTS.scala
```diff
@@ -0,0 +1,1058 @@
...
+    }
+}
+
+private object RTS {
+  // Utility function to avoid catching truly fatal exceptions. Do not allocate
+  // memory here since this would defeat the point of checking for OOME.
+  def nonFatal(t: Throwable): Boolean =
+    if (t.isInstanceOf[InternalError]) false
+    else if (t.isInstanceOf[OutOfMemoryError]) false
+    else true
```
jbgi: could be `!t.isInstanceOf[InternalError] && !t.isInstanceOf[OutOfMemoryError]`

ritschwumm: or even !(t.isInstanceOf[InternalError] || t.isInstanceOf[OutOfMemoryError]), which is a little bit closer to how i'd talk in english

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -0,0 +1,549 @@
...
+   * Supervises the specified action, which ensures that any actions directly
+   * forked by the action are killed with the specified error upon the action's
+   * own termination.
+   */
+  final def supervise[A](io: IO[A], error: Throwable): IO[A] = Supervise(io, error)
+
+  /**
+   * Flattens a nested action.
+   */
+  final def flatten[A](io: IO[IO[A]]): IO[A] = io.flatMap(a => a)
```
notxcain: Why there is no `flatten` method on `IO[A]`? Given `ev: A <:< IO[B]` it's just a `flatMap(ev)`. Same question about `absolve`.

effect/shared/src/main/scala/scalaz/effect/IORef.scala
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * A high-performance, mutable reference for the `IO` monad. This is the `IO`
+ * equivalent of a volatile `var`, and is useful for mutable references inside
+ * a single fiber, not communication between fibers. For communication between
+ * fibers, see `MVar`.
+ *
+ * {{{
+ * for {
+ *   ref <- IORef(2)
+ *   v   <- ref.mod(_ + 3)
```
notxcain: `mod` should be `modify`

jdegoes:
> Btw., how do you reconcile having a data structure for I/O with your opinion that Data Structures Are Antithetical to Functional Programming? 😉

The grand conclusion of many of my articles is to do the exact opposite of the best practice. Because the languages we use are so terrible. `IO` and Scala are no exception. 😉

effect/shared/src/main/scala/scalaz/effect/IO.scala
jdegoes: Because this should be `Liskov` but we do not have such a type class yet in Scalaz 8. 😄 I will definitely add both when we get `Liskov`.

effect/shared/src/main/scala/scalaz/effect/IORef.scala
jdegoes: Thanks, fixed!

effect/jvm/src/main/scala/scalaz/effect/console.scala
jdegoes: What are the differences between them? I was thinking users would probably do:

```
import scalaz.effect.console._
```

effect/shared/src/main/scala/scalaz/effect/IO.scala
jdegoes: Fixed. 👍

effect/jvm/src/main/scala/scalaz/effect/RTS.scala
jdegoes: Fixed. 👍

effect/shared/src/main/scala/scalaz/effect/IO.scala
TomasMikula: https://github.com/scalaz/scalaz/blob/series/8.0.x/base/shared/src/main/scala/scalaz/typeclass/Liskov.scala

effect/jvm/src/main/scala/scalaz/effect/console.scala
TomasMikula: At the very least it is a convention that `package object console` goes into `console/package.scala`.

effect/shared/src/main/scala/scalaz/effect/IO.scala
notxcain: Ah, I see. Ok then :)

effect/jvm/src/main/scala/scalaz/effect/console.scala
jdegoes: Fixed. 👍

jdegoes:
Last minute changes I made to the code before the pull request actually de-optimized it (in most cases). It's nearly back to being faster than it's ever been, just a bit more work to do...

All the feedback has been incorporated. I'll cleanup a few things then probably just merge this as-is with new tickets for the outstanding work (unless folks have any more suggestions).

jdegoes:
@puffnfresh Backport would be pretty easy. Standalone library is easy too, but I will have Scalaz 8 builds running in a week or so, so people can pull in 8 just for IO with ease (then define, e.g. Scalaz 7  or Cats Effect type class instances for it).

jbgi:
@jdegoes see also some minor fixes in https://github.com/jdegoes/scalaz/pull/1 (not sure relevant anymore after last commits)

effect/README.md
```diff
@@ -0,0 +1,294 @@
...
+```scala
+val io: IO[String] = IO.fail(new Error("Oh noes!"))
+```
+
+Like all `IO` values, these are immutable values and do not actually throw any exceptions; they merely describe failure as a first-class value.
+
+You can surface failures with `attempt`, which takes an `IO[A]` and produces an `IO[Throwable \/ A]`:
+
+```scala
+val result = openFile("data.json").attempt.map {
```
scr: Where is this awesome-sauce? I've been curious about how one could have truly non-blocking IO, which called back to the rts rather than having to waste a thread with something like `Files.readAllBytes(...)`.  Java 8 has some sort of stuff in java.nio, Selector, SelectableChannel, e,g. that seems to offer some non-blocking facilities that look quite painful to manage as a client and perfect for rolling into something like this so clients can just program naturally without worrying about the nitty gritty of selecting, registering keys, reading more bytes when ready, etc.  When I first saw this readme a few days ago I was hopeful that truly non-blocking yet sensibly concurrent/parallel file operations existed, but I guess they're just a concept/WIP?  If not please point me to where I can look at how they're hooked up!

jdegoes: It does not exist yet, but will most definitely exist soon. 😄 

The first step is `scalaz-nio`.

jdegoes:
@jbgi  All right, I should have replicated your suggested improvements (thank you!). Take a look and let me know if you see anything else!

effect/README.md
scr: Ok, so I played with how one might read a file (hopefully some file manipulation can be rolled into the library), but I can't quite get it the final mile (i.e. so that each chunk read is done from the RTS so that cancellation or fiber-timeslicing can occur) - sorry, but I just can't my head around how to convert impure code i.e. where making the same call could have different results (bytes read) into pure IO/effect code and how to keep making that call like a do while in the IO RTS…  Any advice/insight appreciated!

https://gist.github.com/scr/9d9fd533f02f7dcf4c4029ab55901412

scr: I see there's a `forever` and `retry` but both of those work on failures - it would seem that a tri-state is needed, or `Throwable \/ Option[T]` with something like a `.doWhile` or `.doUntil` operator that calls the code until it receives either the throwable or an empty/None result (or nonEmpty/Some result if doUntil.

scr: or some sort of `IO.trampoline` thing?

scr: Ok… got it working with existing primitives (thanks @jdegoes for nudge to try IO.flatten); updated gist to show how IO.flatten of the IO.async[IO[A]] lets you chain File work indefinitely to read the entire file in and give chances for cooperation with this Fiber stuff!  So cool!

aloiscochard:
Congratulations @jdegoes for the work being done here.

It is a great feeling for me, after having spent a few months of my life on `matterhorn` and `scalaz8` before totally giving up about Scala OSS... I thought that work would be lost and be just a wasted effort at the end (even though it gave me motivation to create `codec-jvm` in Haskell that was later used by Eta).

It took times... so long that I totally lost interest in Scala FP to be honest, but it is good to see what is happening here, very good feeling.

So this is probably my last message on this repository, you probably noticed I wasn't reviewing much... my interests have shift quite a bit now.

I wish you all my best wishes for further development, I'll keep a curious eye on it for sure!

👏

TomasMikula:
Your contributions will be greatly missed, @aloiscochard. Laying the foundations of scalaz 8 was a spectacular last contribution!

jdegoes:
@aloiscochard It's very sad you are leaving, after the excellent work you have done throughout the years, and especially your work building the new foundations for Scalaz 8. But it's also quite understandable, as better opportunities exist outside the JVM, and better options are emerging on the JVM. I really can't blame you and wish I was headed your way too. 😉 

Thanks for everything you've done over the years. You'll be missed. Hope to run into your work in other OSS projects in better languages.

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -0,0 +1,554 @@
...
+
+  /**
+   * Lazily produces an `IO` value whose construction may have actional costs
+   * that should be be deferred until evaluation.
+   *
+   * Do not use this method to effectfully construct `IO` values. The results
+   * will be undefined and most likely involve the physical explosion of your
+   * computer in a heap of rubble.
+   */
+  final def suspend[A](io: => IO[A]): IO[A] = Suspend(() => io)
```
scr: I'm a little confused by the expected behavior - should lazy mean both deferred and memoized like it is for scala's lazy vals or just deferred?  This currently behaviors just as deferred - if you have, something e 1like the following, wouldn't you expect the output to be 0, then 1?  Instead, subsequent invocations, re-construct the IO, so it's not memoized; only deferred (i.e. not the same lazy as scala, which has both properties).

```
object MyApp extends SafeApp {
  lazy val intGetter = IO.suspend {
    val atomicInt = new AtomicInteger()
    IO.sync {
      atomicInt.getAndIncrement
    }
  }
  for {
   intOne <- intGetter
    _ <- putStrLn(intOne)
   intTwo <- intGetter
    _ <- putStrLn(intTwo)
  } yield ()
}
```

effect/README.md
```diff
@@ -0,0 +1,294 @@
...
+    fiber1 <- fib(n - 2).fork
+    fiber2 <- fib(n - 1).fork
+    v2     <- fiber2.join
+    v1     <- fiber1.join
+  } yield v1 + v2
+```
+
+Interrupting a fiber returns an action that resumes when the fiber has completed or has been interrupted and all its finalizers have been run. These semantics allow construction of programs that do not leak resources.
+
+A more powerful variant of `fork`, called `fork0`, allows specification of a handler that will be passed any uncaught errors from the forked fiber, including all uncaught errors that occur in finalizers. If this handler is not specified, then the handler of the parent fiber will be used, recursively, up to the root handler, which can be specified in `RTS`.
```
hepin1989: could the `fork0` have a better name?

effect/shared/src/main/scala/scalaz/effect/IO.scala
jdegoes: Unfortunately this code is invalid. You are performing an effect inside the block passed to `suspend`, so the semantics of the code are undefined.

`suspend` is designed only to defer the (pure) construction of an `IO`. This is needed in some cases to ensure lazy evaluation in the presence of runtime fusion.

Other literature (incorrectly) encourages the use of `suspend` for performing effects (such as memoization).

effect/README.md
jdegoes: I'd like to use `fork'` but it's not allowed in Scala. 😄 

I'm open to suggestions. Generally I use `0` variants to define the most powerful version of a function (e.g. `bracket0`), from which all other variants are defined.

effect/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Retry
+
+There are a number of useful combinators for repeating actions until failure or success:
+
+ * `IO.forever` &mdash; Repeats the action until the first failure.
+ * `IO.retry` &mdash; Repeats the action until the first success.
+ * `IO.retryN(n)` &mdash; Repeats the action until the first success, for up to the specified number of times (`n`).
+ * `IO.retryFor(d)` &mdash; Repeats the action until the first success, for up to the specified amount of time (`d`).
+
```
hepin1989: add `exponential` delay support

effect/README.md
TomasMikula: Or just overload `fork`?

mielientiev:
Hi @jdegoes
Maybe I miss something, but I wonder is it possible to "convert" scala future to scalaz 8 IO?

(I mean if I already have a legacy method that returns a future and I want to wrap it to IO)

mielientiev:
reply to my previous message, I believe it's very easy to achieve like:
```
def funcWithFuture: Future[Int] = ???
val result: IO[Int] = IO.async { cb =>
    funcWithFuture.onComplete {
        case Success(a) =>cb(\/-(a))
        case Failure(t) => cb(-\/(t))
    }
}
```

jdegoes:
@mielientiev Yes, it's very easy (as the code you have), but this functionality will have to be placed in a third-party library. I can help write that after I get this PR in.

mielientiev:
@jdegoes Also how can we do vice versa? (`IO` -> `Future`), I found only sync unsafePerform, but I didn't find any way to run it in an async way? Can you please point me? ;)

jdegoes:
@mielientiev I can make that function public in my next update (it's buried in RTS).

mielientiev:
@jdegoes it will be great! Thanks

mielientiev:
@jdegoes by the way, just curious why `AsyncEffect` and accordingly `IO.async` uses `(Throwable \/ A => Unit)` and not a `Try[A] => Unit` ?

eleaar:
@jdegoes @mielientiev The transformation from `IO` -> `Future` could be easily doable with just a few changes to the existing code, by separating the actual fiber runtime from the interpreter itself.

I've spiked an example here: https://github.com/jdegoes/scalaz/pull/3. 
What do you think of it?

benchmarks/src/main/scala/scalaz/effect/IODeepAttemptBenchmark.scala
```diff
@@ -28,16 +28,14 @@ class IODeepAttemptBenchmark {
 
   @Benchmark
   def futureDeepAttempt(): BigInt = {
-    import scala.util.{Success, Failure}
+    import scala.util.Success
     import scala.concurrent.Future
     import scala.concurrent.duration.Duration.Inf
 
     def descend(n: Int): Future[BigInt] =
-      if (n == depth) Future.failed(new Error("Oh noes!"))
```
viktorklang: It could easily be argued that throwing Errors and expecting to catch them is madness, that is unless of course, you prefer writing unreasonable applications:

«An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch.» - https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html

jdegoes: All the other implementations support throwing and catching errors. `Future` is the odd duck out.

I can special case this for `Future`, as I have done for numerous other things (wrapping of `Error`, for example).

jdegoes:
@mielientiev  There is no `Try`, or rather, `Try[A]` is a synonym for `Throwable \/ A`. I will delete the type alias since it's confused people. I just did it to save typing.

@eleaar I like it and I have had some time to digest. I need to figure out one more thing before I implement this (or something like it).

effect/README.md
jdegoes: We should have a policy on this. Personally, I think it's OK to use overloading _only_ for methods (never for type class functions or top-level functions / syntax).

effect/jvm/src/main/scala/scalaz/effect/RTS.scala
```diff
@@ -471,13 +471,14 @@ private object RTS {
                         curIo = nextInstr(value, stack)
 
                         if (curIo == null) {
+                          println("GOT HERE!!!!!!!!!!!!")
```
ritschwumm: heh?

jdegoes: Hehe. Ignore, will delete in next commit. Tried to force that code path through tests (unsuccessfully).

Baccata:
I've started trying it to get acquainted with the API. It looks great, however it looks like fibers prevent the process from finishing. For instance, a simple program like the following does print as expected, but then jvm just hangs there afterwards : 

``` scala
import scalaz.effect.{ IO, SafeApp }
import scala.concurrent.duration._

object Main extends SafeApp {
  override def run(args: List[String]): IO[Unit] =
    IO.sync(println("Hello world")).timeout(1.seconds)
}
```

Is this is the expected behavior ?

eleaar:
I've had the same problem when using this branch as a submodule in a larger test project. 
The reason is that the underlying ExecutorServices are not being explicitly shut down in the SafeApp. I've ended up doing a:

```scala
trait SafeApp2 extends RTS {

  def run(args: List[String]): IO[Unit]

  final def main(args0: Array[String]): Unit = {
   try { 
    unsafePerformIO(run(args0.toList))
   } finally {
    threadPool.shutdown()
    scheduledExecutor.shutdown()
   }
  }
}
```


Oddly enough, I didn't experience this behaviour when running a SafeApp from within a fork of this branch.

jdegoes:
@eleaar This is a known issue. It creates threads which hold up exit of the JVM. I'll fix it, for sure.

jdegoes:
@Baccata ^^^

mielientiev:
@jdegoes when does this PR would be merged and released?

jdegoes:
@mielientiev Next week.

fwbrasil:
@jdegoes it seems that this new IO implementation suffers from the same problem as cat-effect regarding strict evaluation. See typelevel/cats-effect#92. Here's a simplified repro:

```scala
def sum(l: List[IO[Int]]): IO[Int] =
  l match {
    case head :: tail => 
      head
        .map(i => loop(tail).map(_ + i))
        .flatMap(identity)
  }

// java.lang.StackOverflowError
sum(List.fill(100000)(IO.now(1)))
```

jdegoes:
@fwbrasil Thanks for reporting, and good to see you here!

This limitation is documented in the source code and README and probably won't be fixed. `map` is guaranteed stack-safe for up to 10k repeated evaluations, but not beyond.

Similarly, `io.map(f1).map(f2).map(f3)...map(fn)` is by law equivalent to `io.map(f1 >>> f2 >>> f3 >>> ... fn)`, but obviously the latter will stack overflow even earlier than the former and cannot be "fixed".

I'd definitely consider fixing this but I'd like to see a non-contrived example of code people have actually written that triggers the SOE. If anyone reading this has any such example, please share!

fwbrasil:
@jdegoes I find Alexandru's code an appealing example of a real-world scenario. I wouldn't wait for people to stumble across this issue to fix it but it's your call :)

BTW, I've been working on that Arrow implementation we spoke about, but for twitter-util. I hope to have something ready in January

jdegoes:
@fwbrasil This class of examples can only be constructed with composition of `point` and `map`, i.e. effect-free code. I am not sure it's a useful example in real-life. However, I will open a ticket and implement a fix if I can preserve most of the performance gains of the current approach (it should be possible).

Also really look forward to seeing the Arrow implementation. ❤️ Keep me posted!

Issue #1487, opened 2017-10-14
Use Liskov in all places instead of subtyping


fommil:
`Bifunctor` (syntax) widen looks like

```
final def widen[C >: A, D >: B]: F[C, D] = F.bimap(self)(identity[C], identity[D])
```

vs `Functor`

```
final def widen[B](implicit ev: A <~< B): F[B] = F.widen(self)
```

seems inconsistent... and more generally should `<:` and `>:` be replaced with Liskov everywhere?

puffnfresh:
Yes, pretty sure. @S11001001 would definitely know.

jdegoes:
I think it makes sense in any case which is not capturing `Liskov` from a subtyping relationship. So that'd include things like `Bifunctor`. We await @S11001001 for the official verdict. 😄

TomasMikula:
I would say yes, as a matter of having the least requirements necessary to do the job. Compiler's knowledge of `A <: B` can easily be turned into `A <~< B`. The reverse is not possible. So `A <: B` is a stronger requirement than `A <~< B`. Let's go with the weaker one.

jdegoes:
@TomasMikula That's a strong argument. 👍

S11001001:
If GADTs were better implemented, it wouldn't matter. But they aren't, so switch to `<~<` as you like.

fommil:
other than @S11001001's two blog posts and SI-2509, does anybody have an example of when Liskov works better than standand subtyping bounds?

I had an example to motivate Liskov over subtypes in my book, but it turned out to be wrong. Basically, why use this instead of `<:` when `<:` works?

TomasMikula:
> Basically, why use this instead of `<:` when `<:` works?

When designing an API that requires `<:`, how do you tell whether `<:` will work for the user? You can't.

There are cases where we can provide `Liskov` via unsafe `force`, justified by some external reasoning (such as the reasoning in [When can Liskov be lifted?](https://typelevel.org/blog/2014/03/09/liskov_lifting.html)), but cannot convince the compiler of `<:`. In that case, if we need to use an API that requires `<:`, we are out of luck, because we only have `Liskov` at hand.

fommil:
I somehow missed *that* blog post, thanks!

TomasMikula:
Other examples (that is, other than the `Functor` data types mentioned in that post) of externally justified Liskov include (in scalaz 8):
[...]

Issue #1469, opened 2017-09-26
\/ should not be covariant


fommil:
It seems out of place that `\/` (and `Validation`) are covariant in their type parameters when so many other data structures are invariant. Is there a reason for this, or is it just inconsistency that we should fix in a breaking change?

I am writing the chapter on data structures pretty soon, I'll gather other variance inconsistencies that I find in here.

TomasMikula:
I think it's just an oversight that needs to be fixed.

puffnfresh:
Yes, oversight!

jdegoes:
Oversight and already fixed in 8.0.

Jacoby6000:
This might be off-topic for this dicussion, but is disjunction necessary now that scala 2.12's `Either` has a right-bias?  Does disjunction do something that `Either` doesn't?  I think it should probably stay in 7.3 for the sake of people still on 2.11 and simpler migrations between minor releases, but 8.0 could go without it (assuming I'm not missing something).

jbgi:
The plan for 8.0 (and also 7.3) is [to make an Disjunction invariant](https://github.com/scalaz/scalaz/issues/1469). But potentially we could use @TomasMikula [trick](
https://github.com/scalaz/scalaz/issues/1288#issuecomment-290599793) to have zero cost conversion to/from std lib `Either`.

See also @S11001001's [comment on the subject](https://github.com/scalaz/scalaz/issues/1288#issuecomment-290599793)

fommil:
@jacoby6000 the stdlib types still have unsafe methods.

tonymorris:
The stdlib `Either` is still missing some useful libraries and adds unsafe libraries.

The original stdlib `Either` proposal, when written 10 years ago and before forced modification, looked very similar to `\/` today. I'll wait.

S11001001:
Our disjunction *should* be zero-cost wrapper around `Either`, just as @TomasMikula is doing in #1450. But scala/bug#10283 hurts us a lot; we need a minimum of Scala 2.12.

@Jacoby6000 While we may have an **opaque** type in 8 suchly, we will *not* simply suggest using `Either` instead, or provide a transparent alias, for reasons already stated.

[...]

PR #1520, opened 2017-11-18
Replace subtying with liskov.

Jacoby6000:
There are a few places in here where I left the subtyping behind. These are mostly places where we'd need a higher-order Liskov, but there are also some others, like `Leibniz`, `Unapply` and `SeqLikeLensFamily` in `Lens.scala`. Additionally, all `FooSyntax` files require a subtyping relationship, and I did _not_ shove Liskov in to those. I'm not sure there would be much benefit anyway.

I'll admit to being relatively new to `Liskov`, so some of my invocations of it may not be the best.

There's quite a bit of unnecessary duplication in here, but since I'm not aware of best practices with liskov I'll leave it as-is for now, since there may be better ways to do some things anyway.

### DIFF
--- a/core/src/main/scala/scalaz/Bifunctor.scala
+++ b/core/src/main/scala/scalaz/Bifunctor.scala
@@ -1,7 +1,5 @@
 package scalaz
 
-import Liskov.>~>
-
 ////
 /**
  * A type giving rise to two unrelated [[scalaz.Functor]]s.
@@ -9,6 +7,8 @@ import Liskov.>~>
 ////
 trait Bifunctor[F[_, _]]  { self =>
   ////
+
+  import Liskov.>~>
 
   /** `map` over both type parameters. */
   def bimap[A, B, C, D](fab: F[A, B])(f: A => C, g: B => D): F[C, D]
--- a/core/src/main/scala/scalaz/Equal.scala
+++ b/core/src/main/scala/scalaz/Equal.scala
@@ -1,6 +1,5 @@
 package scalaz
 
-import Liskov.<~<
 ////
 /**
  * A type safe alternative to universal equality (`scala.Any#==`).
@@ -41,6 +40,9 @@ object Equal {
   @inline def apply[F](implicit F: Equal[F]): Equal[F] = F
 
   ////
+
+  import Liskov.<~<
+
   /** Creates an Equal instance based on universal equality, `a1 == a2` */
   def equalA[A]: Equal[A] = new Equal[A] {
     def equal(a1: A, a2: A): Boolean = a1 == a2
--- a/core/src/main/scala/scalaz/Foldable.scala
+++ b/core/src/main/scala/scalaz/Foldable.scala
@@ -1,7 +1,5 @@
 package scalaz
 
-
-import Liskov.<~<
 ////
 /**
  * A type parameter implying the ability to extract zero or more
@@ -10,6 +8,8 @@ import Liskov.<~<
 ////
 trait Foldable[F[_]]  { self =>
   ////
+
+  import Liskov.<~<
   import collection.generic.CanBuildFrom
 
   /** Map each element of the structure to a [[scalaz.Monoid]], and combine the results. */


tests/src/test/scala/scalaz/ImmutableArrayTest.scala
```diff
@@ -5,10 +5,13 @@ import std.anyVal._
 
 object ImmutableArrayTest extends SpecLite {
 
+  /*
+   TODO: Talk about this.
```
Jacoby6000: I'm not sure what happened here, but it seems like using `B >~> A` instead of `B >: A` caused some type inference issues.  This test no longer compiles because 

```
[error] /Users/jacobbarber/projects/scalaz/tests/src/test/scala/scalaz/ImmutableArrayTest.scala:9: could not find implicit value for parameter ev: String >~> Int
[error]     val xs = ImmutableArray.fromArray(Array(1)) ++ ImmutableArray.fromArray(Array("a"))
```

This should be attempting to Lub in to `AnyRef`.

tests/src/test/scala/scalaz/CofreeTest.scala
```diff
@@ -69,7 +69,7 @@ object CofreeTest extends SpecLite {
           fa.head,
           fa.tail.map(s =>
             Foldable[CofreeLazyOption].foldRight(s, Stream.empty[A])(_ #:: _)
-          ).getOrElse(Stream.empty)
+          ).getOrElse(Stream.empty[A])
```
Jacoby6000: Example of type inference getting worse in some places.

core/src/main/scala/scalaz/Either.scala
```diff
@@ -263,52 +263,52 @@ sealed abstract class \/[+A, +B] extends Product with Serializable {
...
-        case \/-(b2) => Order[BB].apply(b1, b2)
+        case \/-(b2) => Order[BB].apply(evB(b1), b2)
         case -\/(_) => Ordering.GT
       }
     }
 
   /** Show for a disjunction value. */
-  def show[AA >: A, BB >: B](implicit SA: Show[AA], SB: Show[BB]): Cord =
-    this match {
+  def show[AA, BB](implicit evA: A <~< AA, evB: B <~< BB, SA: Show[AA], SB: Show[BB]): Cord =
```
Jacoby6000: Discussion in #1487 says we should get rid of the subtype relationship here and just use `Show[A]` and `Show[B]`.  However, the scala compiler whines about variance if we do it now.  This will have to wait until Issue #1469 is resolved... Unless there's some scalac variance magic I don't know about.

core/src/main/scala/scalaz/StateT.scala
```diff
@@ -73,26 +74,26 @@ sealed abstract class IndexedStateT[F[_], -S1, S2, A] { self =>
     )
 
   import Liskov._
-  def unlift[M[_], FF[_], S <: S1](implicit M: Comonad[M], F: Monad[λ[α => M[FF[α]]]], ev: this.type <~< IndexedStateT[λ[α => M[FF[α]]], S, S2, A]): IndexedStateT[FF, S, S2, A] = IndexedStateT.createState(
+  def unlift[M[_], FF[_], S](implicit M: Comonad[M], F: Monad[λ[α => M[FF[α]]]], ev: this.type <~< IndexedStateT[λ[α => M[FF[α]]], S, S2, A]): IndexedStateT[FF, S, S2, A] = IndexedStateT.createState(
```
Jacoby6000: The S1 subtype relationship is no longer here (its gone on `unliftId` as well.)  I think this is fune, because of the subtype relationship on IndexedStateT. Just thought I'd mention it in case this may cause an issue I'm unaware of.

core/src/main/scala/scalaz/ImmutableArray.scala
```diff
@@ -158,15 +159,15 @@ object ImmutableArray extends ImmutableArrayInstances {
     def apply(idx: Int) = arr(idx)
 
     def length = arr.length
-    def toArray[B >: A : ClassTag] = arr.clone.asInstanceOf[Array[B]]
-    def copyToArray[B >: A](xs: Array[B], start: Int, len: Int): Unit = { arr.copyToArray(xs, start, len) }
+    def toArray[B: ClassTag](implicit ev: B >~> A): Array[B] = arr.clone.asInstanceOf[Array[B]]
+    def copyToArray[B](xs: Array[B], start: Int, len: Int)(implicit ev: B >~> A): Unit = { arr.asInstanceOf[Array[B]].copyToArray(xs, start, len) }
```
alexknvl: Why is this `asInstanceOf` necessary? If it is indeed necessary, perhaps add a short comment explaining why.

core/src/main/scala/scalaz/Either.scala
TomasMikula: What exactly is the problem with variance here?

Jacoby6000: Covariant type in invariant position.

```
[error] /Users/jacobbarber/projects/scalaz/core/src/main/scala/scalaz/Either.scala:310:21: covariant type A occurs in invariant position in type scalaz.Show[A] of value SA
[error]   def show(implicit SA: Lambda[a => Show[a]][A], SB: Show[B]): Cord =
[error]                     ^
```

core/src/main/scala/scalaz/ImmutableArray.scala
Jacoby6000: The reason is that provided an Array[A], copyToArray wants an Array[A] as input.  Any thing else and the compiler claims that it really wants an Array[Any], which is quite obviously bogus. 

```scala
[error] /Users/jacobbarber/projects/scalaz/core/src/main/scala/scalaz/ImmutableArray.scala:164:108: type mismatch;
[error]  found   : Array[B]
[error]  required: Array[Any]
[error] Note: B <: Any, but class Array is invariant in type T.
[error] You may wish to investigate a wildcard type such as `_ <: Any`. (SLS 3.2.10)
[error]     def copyToArray[B](xs: Array[B], start: Int, len: Int)(implicit ev: B >~> A): Unit = { arr.copyToArray(xs, start, len) }
[error]                                                                                                            ^
```

If I make `arr` (defined above) private to the package rather than private to `this`, then I can use the implicit evidence to convert `this` in to an `ImmutableArray1[B]`, and then perform the copy from there without a cast.  I'm not sure if it would be okay for `arr` to leak to other instances though.

Jacoby6000:
After #1603 gets merged, I'll sync this branch and I suspect quite a bit will change. Probably a lot simplified.

core/src/main/scala/scalaz/LazyOption.scala
```diff
@@ -96,8 +98,11 @@ sealed abstract class LazyOption[+A] extends Product with Serializable {
       y <- b
     } yield (x, y)
 
-  def unzip[X, Y](implicit ev: A <:< (X, Y)): (LazyOption[X], LazyOption[Y]) =
-    fold(xy => (lazySome(xy._1), lazySome(xy._2)), (lazyNone, lazyNone))
+  def unzip[X, Y](implicit ev: A <~< (X, Y)): (LazyOption[X], LazyOption[Y]) =
+    this.fold(
+      { a => val (xy1, xy2) = ev(a); (lazySome(xy1), lazySome(xy2)) },
+      (lazyNone, lazyNone)
+    )
```
emilypi: Ditto here with the pattern matches

fommil:
ouch, big conflicts

fommil:
(we need to get better at merging stuff before it goes stale...)

fommil:
:sadpanda:

PR #1521, opened 2017-11-21
Maybe: Add just syntax

NeQuissimus:
- Removed `POptionAsMaybe` from `Prelude`, a copy of it exists as `MaybeSyntax.OptionAsMaybe` already
- Include `MaybeSyntax` in `MaybePrelude` to restore the removed functionality
- Add `MaybeOps` that provides a generic `just`
- Add above `just` call to `Maybe` examples

### DIFF
--- a/base/shared/src/main/scala/scalaz/Prelude.scala
+++ b/base/shared/src/main/scala/scalaz/Prelude.scala
@@ -1,7 +1,6 @@
 package scalaz
 
 import typeclass._
-import data._
 
 import scala.language.implicitConversions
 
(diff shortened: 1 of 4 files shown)


NeQuissimus:
Ping @TomasMikula as the author of `Maybe`

TomasMikula:
👍  for this change. In general though, I'm a bit wary of polluting `Prelude` too much.

NeQuissimus:
@TomasMikula What is the general idea for the prelude? Having `Prelude` and maybe a `Syntax` object that can be imported? I am thinking it could be similar to the current `import scalaz._` and `import Scalaz._`?

TomasMikula:
I don't think we have guidelines for Prelude. My worry is that we could accidentally put something in `MaybeSyntax` that we didn't intend to include in Prelude. But then, where to put it?

NeQuissimus:
@TomasMikula I am with you, totally agreed... As I said, it may just make sense to have an extra `Syntax._` import that users can grab for the convenience method. Then `Prelude` can be limited to the essentials.

TomasMikula:
👍

NeQuissimus:
I am closing this in favor of #1530

NeQuissimus:
Let's try this again once we have #1528 and #1533 figured out. I messed something up pretty good in here anyways :D

PR #1524, opened 2017-11-22
Leibniz & Liskov : Part 1

alexknvl:
I have decided to redo the https://github.com/scalaz/scalaz/pull/1314 PR and split it into a number of smaller PRs.

* Moved `Liskov` into `data` and renamed it as `As`.
* Renamed `Leibniz` to `Is`.
* Added bounded versions of `Is` and `As` named respectively `Leibniz` and `Liskov`.
* Added a number of useful combinators to both `Is` and `As`.


project/ScalazBuild.scala
```diff
@@ -27,7 +27,7 @@ object Scalaz {
     "-Ywarn-nullary-override",
     "-Ywarn-nullary-unit",
     "-Ywarn-numeric-widen",
-    "-Ywarn-unused:_,-imports",
+    "-Ywarn-unused:_,-imports,-locals",
```
NeQuissimus: Why is this an issue?

base/shared/src/main/scala/scalaz/data/Forall.scala
```diff
@@ -58,13 +57,13 @@ trait ForallSyntax {
 }
 
 object ForallSyntax {
-  final class Ops[F[_]](val a: ∀[F]) extends AnyVal {
+  final class Ops[F[_]](val a: Prelude.∀[F]) extends AnyVal {
```
NeQuissimus: Can you import this?

project/ScalazBuild.scala
alexknvl: https://github.com/scala/bug/issues/10623

base/shared/src/main/scala/scalaz/data/Forall.scala
alexknvl: Oops, I changed this accidentally.

alexknvl: Fixed and rebased.

project/ScalazBuild.scala
NeQuissimus: thx, good to know :)

jbgi: I think using [`@silent` annotation](https://github.com/ghik/silencer) for that specific statement (with a `FIXME` that refer to the bug) would be better that disabling the warnings for the whole project.

alexknvl: Fixed.

base/shared/src/main/scala/scalaz/data/Liskov.scala
```diff
@@ -0,0 +1,177 @@
...
+
+    def substCt[LF, HF >: LF, F[_] >: LF <: HF]
+    (fb: F[B])(implicit F: IsContravariant[F]): F[A] =
+      liftCtF[LF, HF, F].coerce(fb)
+  }
+
+  /**
+    * Subtyping is antisymmetric in theory (and in Dotty). Notice that this is
+    * not true in Scala until [[https://issues.scala-lang.org/browse/SI-7278
+    * SI-7278]] is fixed, so this function is marked unsafe.
```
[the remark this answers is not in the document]
alexknvl: Ah, that's a remnant from https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/Liskov.scala#L136 (`unsafe: Unsafe`). I'd like to introduce something like that into Scalaz later. I'll fix the comment.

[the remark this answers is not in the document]
TomasMikula: I don't have a strong opinion about that.

Btw, is "bracket" an  established name for this? If not, wouldn't `antisymmetry`be better?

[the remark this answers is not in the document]
alexknvl: Let me see if I can come up with a test case that shows unsoundness (also need to add `bracket` to `As`, I think it is safe there). There is already an unavoidable unsoundness related to type projections and intersections - https://gist.github.com/alexknvl/56028633ef73faeec0f5d3e7e3a14746...

I took that name from Scalaz7 - https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Liskov.scala#L207.

jdegoes:
@alexknvl Really exciting work! I think @TomasMikula is the perfect one to review this (thank you Tomas for doing the review!).

base/shared/src/main/scala/scalaz/data/Liskov.scala
```diff
@@ -0,0 +1,177 @@
...
+/**
+  * Liskov substitutability: A better `<:<`.
+  *
+  * `Liskov[A, B]` witnesses that `A` can be used in any negative context
+  * that expects a `B`. (e.g. if you could pass an `A` into any function
+  * that expects a `B`.)
+  *
+  * @see [[<~<]] `A <~< B` is a type synonym to `Liskov[A, B]`
+  */
+sealed abstract class Liskov[-L, +H >: L, -A >: L <: H, +B >: L <: H] private[Liskov]() { ab =>
```
emilypi: It's already sealed, so there's no point to a private constructor `sealed abstract class Liskov[-L, +H >: L, -A >: L <: H, +B >: L <: H] private[Liskov]() `

base/shared/src/main/scala/scalaz/data/Liskov.scala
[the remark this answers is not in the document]
alexknvl: https://gist.github.com/alexknvl/369002d1338e410586a5c0a517699f29
It looks like all unsafety of `bracket` is caused by the interaction of other language features that we have no control over and that can be exploited without using `bracket` directly:

 * Inner classes + type projections is unsound.
 * Type projections + existentials is unsound.
 * Type projections + linearization of `with` is unsound.
 * Type projections + linearization of `with` + `(A with B) ~ (B with A)` is unsound.
 * Bounded virtual types + intersections (apparently even in Dotty) + `null` is unsound.
 * Bounded virtual types + intersections + non-totality.

I propose use the subset of the language that does not include type projections, `null` and evaluate `lazy val p` before using `p.T` to be morally correct (a la http://www.cse.chalmers.se/~nad/publications/danielsson-et-al-popl2006.pdf), and declare `bracket` safe within this subset of the language.

base/shared/src/main/scala/scalaz/data/Is.scala
```diff
@@ -0,0 +1,218 @@
...
+    *     ab lift2[Either] ij
+    * }}}
+    *
+    * @see [[Is.lift]]
+    * @see [[Is.lift2]]
+    * @see [[Is.lift3]]
+    */
+  def lift2[F[_, _]]: PartiallyAppliedLift2[F] =
+    new PartiallyAppliedLift2[F]
+  final class PartiallyAppliedLift2[F[_, _]] {
```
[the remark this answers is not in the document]
hrhino: You can't do that exactly, because it's a non-static inner class; you could (should) have a `class Is.PartiallyAppliedLift2[A, B] private (val ab: A Is B) extends AnyVal`...

base/shared/src/main/scala/scalaz/data/Liskov.scala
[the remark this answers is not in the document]
alexknvl: Another possible solution is to ban intersections entirely (they are present in every single case above except the inner class one) unless used in `extends`.

base/shared/src/main/scala/scalaz/data/Liskov.scala
alexknvl: Fixed.

alexknvl:
@TomasMikula 
> would it be of any benefit to provide `Is[A, B] === Leibniz[Nothing, Any, A, B]`

I don't think it is particularly useful since both types have at most one instance. It's unlikely you'll ever want to do `List[Is[A, B]] => List[Leibniz[Nothing, Any, A, B]]` :smile:

base/shared/src/main/scala/scalaz/data/Liskov.scala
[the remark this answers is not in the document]
alexknvl: I added `bracket` to `As` and removed the unsafety warning from the docs since it is a problem with the language, not with `bracket` per say.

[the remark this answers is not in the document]
TomasMikula: Thanks for digging in, that's a nice summary!

PR #1525, opened 2017-11-22
Simplify the first 5 tupled methods of ApplicationBuilder(s)

phderome:
use ap.tuple2 to ap.tuple5 for ApplicationBuilder to ApplicationBuilder5 tupled methods and thus use implicit ap directly

It seems a bit simpler and consistent with the availability of the ap parameter (observed in discussion with @fommil)


phderome:
Feel free to reject if you are uneasy about the change as I cannot argue strongly for it.

fommil:
Aaah, interesting, that's why you did it this way. We all learnt something today :-D

PR #1529, opened 2017-11-23
Add a These data type.

### LINKED ISSUE #1527: Make a shiny new These.
For [shiny new scalaz 8](https://github.com/scalaz/scalaz/issues/1526).

Making an issue to point out that I'm doing it, as suggested in gitter.

hrhino:
`These[A, B] == Either3[A, B, (A, B)]`

Instances are those given in [`Data.These`](https://hackage.haskell.org/package/these-0.7.4/docs/Data-These.html) (where the typeclass exists).

This isn't complete (it needs also `TheseT`), and it's missing some instances (like `Bifunctor`) as they don't yet exist. I'm putting it up anyhow for feedback on style and what methods/functions deserve to exist.

Closes #1527.


base/shared/src/main/scala/scalaz/data/These.scala
```diff
@@ -0,0 +1,125 @@
...
+  /**
+    * The canonical catamorphism from `These` values.
+    */
+  final def fold[C](fl: L => C)(fr: R => C)(fboth: (L, R) => C): C = this match {
+    case This(left)        => fl(left)
+    case That(right)       => fr(right)
+    case Both(left, right) => fboth(left, right)
+  }
+
+  final def thisSide: Option[L] = this match {
```
jdegoes: On these, and elsewhere, can we use `Maybe` instead of `Option`?

base/shared/src/main/scala/scalaz/data/These.scala
```diff
@@ -0,0 +1,125 @@
...
+  type \&/[A, B] = These[A, B]
+
+  /* ADT cases */
+  final case class This[L, R](thisValue: L)               extends These[L, R]
+  final case class That[L, R](thatValue: R)               extends These[L, R]
+  final case class Both[L, R](thisValue: L, thatValue: R) extends These[L, R]
+
+  /* "smart" constructors */
+  object This {
+    @inline def apply[L, R](thisValue: L): These[L, R] = new This[L, R](thisValue)
```
jdegoes: I'd add `final` to all these smart constructors too, as a matter of style.

base/shared/src/main/scala/scalaz/data/TheseInstances.scala
```diff
@@ -0,0 +1,41 @@
...
+import scalaz.typeclass._
+
+trait TheseInstances {
+  import These._
+
+  // implicit def bifunctor: Bifunctor[These] = ...
+
+  implicit def monad[L: Semigroup]: Monad[These[L, ?]] =
+    new MonadClass.Template[These[L, ?]] with BindClass.Ap[These[L, ?]] {
+      override def map[A, B](ma: These[L, A])(f: A => B) = ma.rmap(f)
```
jdegoes: `final` on all these, too.

example/shared/src/main/scala/scalaz/example/TheseUsage.scala
```diff
@@ -0,0 +1,49 @@
+package scalaz
+package example
+
+import data._, typeclass._
+
+object TheseUsage extends App with Instances {
```
jdegoes: 👍 on the examples. Maybe nice to add one for `swap` or any other common methods.

example/shared/src/main/scala/scalaz/example/TheseUsage.scala
```diff
@@ -0,0 +1,49 @@
...
+  assert(foo.append(one)   == Both(1, "foo"))
+  assert(foo.append(foo)   == That("foofoo"))
+  assert(foo.append(both)  == Both(1, "foofoo"))
+  assert(both.append(one)  == Both(2, "foo"))
+  assert(both.append(foo)  == Both(1, "foofoo"))
+  assert(both.append(both) == Both(2, "foofoo"))
+
+}
+
+trait Instances {
```
jdegoes: Having the name `Instances`, it's likely to clash with other examples in the same package. Maybe `TheseInstances`?

hrhino:
Gotcha, @jdegoes. Thanks.

I have a `TheseT` locally, but given the recent gitter discussion re: monad transformer stacks in Scala, I didn't commit it yet. If people think we should have it I'll push it up, otherwise I think this stands on its own.

example/shared/src/main/scala/scalaz/example/TheseUsage.scala
```diff
@@ -0,0 +1,65 @@
+package scalaz
```
NeQuissimus: Could you refactor this to similar to #1537 ?

hrhino: Sure thing! I think the `tut` plugin wasn't in yet when I did this. Thanks for the reminder.

jdegoes:
@hrhino Should be good to go as soon as the `tut` refactoring is complete! Thank you. 😄

example/shared/src/main/scala/scalaz/example/TheseUsage.scala
NeQuissimus: No worries, things overlapped a little, I had the same thing with the usage of Disjunction

jdegoes:
@hrhino Thank you!

tonymorris:
A common operation using `These` is zipping where any misalignment is retained, such that an inverse could be written:

    def xzip[A, B](a: List[A], b: List[B]): List[These[A, B]]

Also, `TheseT` can be written from which `These` can be an instance, but as a church encoding. This would likely have consequences for performance.

I am also suspicious of the `flatMap` method. Is it lawful?

hrhino:
That `xzip` looks like good stuff. Regarding design, do you think it belongs on `object These`, `Foldable`, or somewhere else?

There was a chat in gitter about monad transformer stacks being of dubious practicality in Scala. I can add a `TheseT` that's a newtype wrapper over `F[These[L, T]]`, too, if that's something people want irrespective of the wisdom of transformer stacks.

The `flatMap` is lawful:
- Conceptually, it's a monad which collects `L` values as non-fatal errors, as opposed to `\/` which collects fatal errors. `f >>= g` sequences the value inside `f` with `g`, and collects the `L` values in both.
- Sloppily: https://gist.github.com/hrhino/a81f2b7199b7cfce9570256f309c2304. There's no law checking in 8.0.x yet but I'll gladly formalize this when the law testing particulars are decided.

PR #1532, opened 2017-11-24
Show `Show` some love.

hrhino:
It's a very useful typeclass if you have an `Any#toString`-shaped hole in your cornea. Someone seems to have made a file for it and then never done anything with it. This resurrects it as a typical typeclass, with syntax, instances, usage examples, &c.

### DIFF
--- a/base/shared/src/main/scala/scalaz/typeclass/Show.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/Show.scala
@@ -1,5 +1,21 @@
 package scalaz
+package typeclass
 
-abstract class Show[A] {
+/** A typeclass describing types which can be meaningfully represented as a `String`.
+  */
+trait Show[A] {
+  /** Produce a `String` representation of `a`.
+    */
+  /* todo: when we have a `Cord`, change this to return that instead,
+   * todo: and add a `def shows(a: A): String = show(a).toString`
+   */
   def show(a: A): String
+}
+
+object Show extends ShowInstances with ShowSyntax {
+  @inline def apply[A](implicit A: Show[A]): Show[A] = A
+
+  /** A factory for `Show` instances which delegates to Scala's `.toString` method.
+    */
+  def fromToString[A]: Show[A] = _.toString()
 }
--- a/base/shared/src/main/scala/scalaz/typeclass/ShowInstances.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/ShowInstances.scala
@@ -1,0 +1,14 @@
+package scalaz
+package typeclass
+
+trait ShowInstances {
+
+  implicit final def string: Show[String] = s => s
+
+
+  implicit final def contravariant: Contravariant[Show] =
+    new Contravariant[Show] {
+      def contramap[A, B](r: Show[A])(f: B => A): Show[B] =
+        b => r.show(f(b))
+    }
+}
--- a/base/shared/src/main/scala/scalaz/Prelude.scala
+++ b/base/shared/src/main/scala/scalaz/Prelude.scala
@@ -31,6 +31,7 @@ trait Prelude  extends data.DisjunctionFunctions
   type Cobind[F[_]] = typeclass.Cobind[F]
   type IsCovariant[F[_]] = typeclass.IsCovariant[F]
   type IsContravariant[F[_]] = typeclass.IsContravariant[F]
+  type Show[A] = typeclass.Show[A]
 
   def Applicative[F[_]](implicit F: Applicative[F]): Applicative[F] = F
   def Apply[F[_]](implicit F: Apply[F]): Apply[F] = F
@@ -48,6 +49,7 @@ trait Prelude  extends data.DisjunctionFunctions
   def Cobind[F[_]](implicit F: Cobind[F]): Cobind[F] = F
   def IsCovariant[F[_]](implicit F: IsCovariant[F]): IsCovariant[F] = F
   def IsContravariant[F[_]](implicit F: IsContravariant[F]): IsContravariant[F] = F
+  def Show[A](implicit A: Show[A]): Show[A] = A
 
   // ApplicativeSyntax
   implicit def PapplicativeOpsA[A](a: A): ApplicativeSyntax.OpsA[A] = new ApplicativeSyntax.OpsA(a)
@@ -93,6 +95,9 @@ trait Prelude  extends data.DisjunctionFunctions
 
   implicit def ComonadOps[F[_], A](fa: F[A])(implicit F: Comonad[F]): ComonadSyntax.Ops[F, A] =
     new ComonadSyntax.Ops(fa)
+
+  implicit def ShowOps[A](a: A)(implicit A: Show[A]): ShowSyntax.Ops[A] =
+    new ShowSyntax.Ops[A](a)(A)
 
 
   // Base Data
--- a/base/shared/src/main/scala/scalaz/data/AMaybe.scala
+++ b/base/shared/src/main/scala/scalaz/data/AMaybe.scala
@@ -8,7 +8,7 @@ import Prelude.===
   */
 sealed abstract class AMaybe[F[_, _], A, B]
 
-case class AJust[F[_, _], A, B](value: F[A, B]) extends AMaybe[F, A, B]
+final case class AJust[F[_, _], A, B](value: F[A, B]) extends AMaybe[F, A, B]
 
 /**
   * The empty case contains evidence of type equality
@@ -20,7 +20,7 @@ sealed abstract case class AEmpty[F[_, _], A, B]() extends AMaybe[F, A, B] {
   def leibniz: A Is B = subst[A === ?](Is.refl)
 }
 
-object AMaybe {
+object AMaybe extends AMaybeInstances {
   def empty[F[_, _], A]: AMaybe[F, A, A] = None.asInstanceOf[AMaybe[F, A, A]]
 
   private val None = none[Nothing, Nothing]
--- a/base/shared/src/main/scala/scalaz/data/AMaybeInstances.scala
+++ b/base/shared/src/main/scala/scalaz/data/AMaybeInstances.scala
@@ -1,0 +1,13 @@
+package scalaz
+package data
+
+import typeclass._
+
+trait AMaybeInstances {
+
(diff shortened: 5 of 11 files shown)


base/shared/src/main/scala/scalaz/typeclass/ShowInstances.scala
```diff
@@ -0,0 +1,15 @@
+package scalaz
+package typeclass
+
+trait ShowInstances {
+
+  implicit final def string: Show[String] =
+    Show.instance[String](identity)
+
+
+  implicit final def contravariant: Contravariant[Show] =
```
hrhino: I feel somewhat strange about having both `Show` instances for stdlib types and typeclass instances for `Show` in the same file, but I couldn't find precedent for how to do this. I personally think that `Show` is an excellent contravariant functor, so I put it in.

In general, I'm not sure how many stdlib instances should go here. I suspect `Int` and friends, but it raises the question of whether `1.toByte.show` should just be `"1"`, with no indication that it's a `Byte`. I personally like `Show` instances, where it's not contradicted by another goal, to make it as clear as possible what type is being shown.

fommil: it's not just a contravariant, it's also a [`LabelledContravariantDerivez`](https://gitlab.com/fommil/scalaz-deriving/blob/v0.9.0/scalaz-deriving/src/main/scala/scalaz/Derivez.scala#L95-110) :wink:

base/shared/src/main/scala/scalaz/typeclass/Show.scala
```diff
@@ -1,5 +1,22 @@
 package scalaz
+package typeclass
 
-abstract class Show[A] {
+/** A typeclass describing types which can be meaningfully represented as a `String`.
+  */
+trait Show[A] {
   def show(a: A): String
```
fommil: I really prefer the `Cord` approach of scalaz7. Consider the case of building up the Show for a nested ADT. And indeed if we leave the `Cord` method abstract it means we can use SAM types when defining instances.

base/shared/src/main/scala/scalaz/typeclass/Show.scala
```diff
@@ -1,5 +1,22 @@
...
 }
+
+object Show extends ShowInstances with ShowSyntax {
+  def apply[A](implicit A: Show[A]): Show[A] = A
+
+  /** Create a `Show` instance for `A` from a function `A => String`.
+    */
+  def instance[A](show0: A => String): Show[A] = new Show[A] {
```
fommil: this is not needed... SAM types FTW

base/shared/src/main/scala/scalaz/typeclass/Show.scala
```diff
@@ -1,5 +1,22 @@
...
 
-abstract class Show[A] {
+/** A typeclass describing types which can be meaningfully represented as a `String`.
+  */
+trait Show[A] {
   def show(a: A): String
 }
+
+object Show extends ShowInstances with ShowSyntax {
+  def apply[A](implicit A: Show[A]): Show[A] = A
```
fommil: `@inline` ?

base/shared/src/main/scala/scalaz/typeclass/Show.scala
hrhino: I do as well, but `Cord` isn't in scalaz8 yet, right? I'll put a comment saying that once it's in this should be changed.

base/shared/src/main/scala/scalaz/typeclass/Show.scala
hrhino: :+1: to supporting only 2.12+!

fommil: this works in 2.11 :smile: 

```
scalacOptions += "-Xexperimental"
```

base/shared/src/main/scala/scalaz/typeclass/Show.scala
fommil: I feel like we really need a macro or compiler plugin like `@simulacrum.typeclass` to remove the boilerplate of all of this.

hrhino: Yeah. `simulacrum` doesn't do the zero-cost syntax; it could be made to support it or we could adapt it for our own purposes. That's not really my call, though; I'm just following what seems to be the current style.

TomasMikula:
We should also add a string interpolator for `Show`, as in #1496.

hrhino:
I did some of that for this PR, but I deleted it because I want to add `Exists` first. Since we have `Forall` (thanks!) it seems fitting, and would make the interpolator work in a bit less ad-hoc way than writing a version specialized to `Show`.

puffnfresh:
Should Show exist?

fommil:
We really need a better string interpolator or its all too easy to call toString. With custom predef this is possible.

TomasMikula:
@puffnfresh Can you expand?

puffnfresh:
Show always seemed not very useful as a type-classes. What does it mean? What are the laws?

puffnfresh:
I would absolutely support scalaz 8 not having Show

hrhino:
@fommil yeah, it's going to happen

@puffnfresh indeed, I'm interested in why. It's got laws (or at least I'd argue for some) and it's quite useful.

Laws off the top of my head:
- If `A` has an `Eq`, `a == b` implies `show(a) == show(b)`
- If `A` has a `Read`, ` read(show(a)) == Just(a)`

And while there's a lot of types which serialize better as structured data (such as JSON), I think the prevalence of a (much worse) `.toString` in most other languages argues at least that it can be useful. Even Haskell quite happily has one. 

I do understand that (in comparison to other common typeclasses) `Show` is less endowed with laws and has a much weaker claim to being backed by category theory or mathematics in general, and these do argue against its inclusion in a pure FP library. However, any serialization format will have those characteristics, and while I'd be disappointed to see a JSON library in scalaz, I think that the prevalence of strings as a data interchange format and their utility in communication with users and developers puts it in a somewhat different category. I won't be personally insulted if it's removed and banished to a separate library, but I think that many people would not use scalaz for very much without pulling it in.

TomasMikula:
It is useful, and it is convenient for methods to take it as an implicit parameter. Whether we call it a typeclass or not is less important.

TomasMikula:
Happy to put it in package `outlaws`.

puffnfresh:
I don't think we've given Eq a substitution law, so that first law would have meaning. What do we gain from not being able to write instances for types which have an Eq without substitution?

The data type is (=> String) if it's not a type-class. What useful methods do we get which we don't from (=>) already?

hrhino:
I'm confused. Wouldn't a substitution law for `Eq` in general obviate the first law I mentioned? I think it may be nice to have substitution specifically for `Show`. Maybe not. I'm mostly spitballing here.

puffnfresh:
Yes, I'm asking what it gives, that's all.

tonymorris:
I think `Show` is useful to work against legacy libraries that use `String` in the usual inappropriate places.

Also, I think a JSON library in Scalaz would be great. It is what argonaut should have been when we wrote it.

puffnfresh:
Also we should have Eq with substitution law as a separate class, I think.

hrhino:
@puffnfresh +1 to that. 

@tonymorris I suppose that depends what the scope of scalaz is, and that's not my call.

tonymorris:
@hrhino I am of the opinion that the best way to manage a scala project is to compile as much as possible, all at once, to prevent many problems, including the one we have today with library versions and incompatibilities all over the place.

A JSON library would utilise a library such as Scalaz, with lenses written properly, for example. This would significantly advantage the users of the library.

This is all a different discussion though.

hrhino:
I could get behind that, actually. Thanks.

jdegoes:
One option is to mandate that any `Show` can be copy/pasted into the REPL to construct a data structure that is `===` to the original. With macros, this can be tested.

Another potential option is to _just_ have a type class that combines `Read` and `Show`. Basically string serialization codec. 

Yet another option is some sort of `Generic` type class that we can `@derive` automatically. Then we can do a `print` and `prettyPrint` and lots of useful _functions_ (not type classes) based on this generic representation.

fommil:
best to have contravariant `Show` and covariant `Read` as separate type class interfaces, which could be combined with an implicit into one. `scalaz-deriving` doesn't have a nice API for invariant typeclasses that combine encoders/decoders.

edmundnoble:
@fommil Combining type class instances automatically means that the user doesn't have to make sure the combined instance follows the laws, which seems bad to me. Having a type class which subclasses both `Show` and `Read` though would be fine in my estimation.

tonymorris:
I am of the opinion that `Read` has zero use-cases and its omission incurs no penalty.

puffnfresh:
I agree with `Read`. It shouldn't exist.

I don't know when `Show` is useful for legacy reasons, but I'll defer to @tonymorris and others.

tonymorris:
`Show` is useful for when dealing with a stringy API. Or in development, "show me the value for this data structure." Ideally we'd navigate the data structure using more data structures, but we aren't there yet.

fommil:
I find `Show` to be useful for displaying debugging information to a developer. It would have far more utility if `s"..."` interpolation only accepted things with a `Show` and then we could avoid the nonsense default `toString` on functions and so on. I think it should be possible to have a macro that creates the `Show` for lambdas: just display their code! (or at least a reference to the source and line) I'm not so sure about general closures where other variables would require more analysis to see their origin.

ritschwumm:
there is https://github.com/scalaz/scalaz/issues/1495 - someone needs to make a wartremover wart for ``s""```.

fommil:
with scalafix's `Disable` rule, a dedicated wart may not be needed.

aloiscochard:
In Rust there is a sort of equivalent typeclass named `Debug`, I like that it explicit this is used only to print debug output (as opposed to be general about "Showing" a value)..

There is some usability aspect as well, for example it won't compile if you just try to print a value... you a have to be explicit that you want to "debug" that value.

Milyardo:
I've always been under the impression that `Show` was for human consumption.
An applicative adjunction for reading and writing typeclasses would still be useful, I just don't think it should use Show.

PR #1535, opened 2017-11-25
Reduce unnecessary work in Cord#toString, Cord#:-, Cord#-:

edmundnoble:
Please tell me if I'm missing something, because it looks like `toString` is terrible. It recreates the entire tree, it creates a huge `Trampoline` instance to suspend side effects, and the cardinal sin, it's more code than the simpler and more efficient solution.

Also, `x.toString` means boxing `x` before calling `toString`. `String.valueOf` could be a small optimization, but less boxing is never a bad thing.


edmundnoble:
Ugh, github is crap. Reopening on 7.2.x.

PR #1536, opened 2017-11-25
Redux: Reduce unnecessary work in Cord#toString, Cord#:-, Cord#-:

edmundnoble:
Pasted:

Please tell me if I'm missing something, because it looks like toString is terrible. It recreates the entire tree, it creates a huge Trampoline instance to suspend side effects, and the cardinal sin, it's more code than the simpler and more efficient solution.

Also, x.toString means boxing x before calling toString. String.valueOf could be a small optimization, but less boxing is never a bad thing.

### DIFF
--- a/core/src/main/scala/scalaz/Cord.scala
+++ b/core/src/main/scala/scalaz/Cord.scala
@@ -69,13 +69,13 @@ final case class Cord(self: FingerTree[Int, String]) {
    * Prepends a `Char` to the beginning of this `Cord`.
    * Time complexity: O(1)
    */
-  def -:(x: => Char): Cord = cord(x.toString +: self)
+  def -:(x: => Char): Cord = cord(String.valueOf(x) +: self)
 
   /**
    * Appends a `Char` to the end of this `Cord`.
    * Time complexity: O(1)
    */
-  def :-(x: => Char): Cord = cord(self :+ x.toString)
+  def :-(x: => Char): Cord = cord(self :+ String.valueOf(x))
 
   /**
    * Removes the first character of this `Cord`.
@@ -111,11 +111,8 @@ final case class Cord(self: FingerTree[Int, String]) {
   def toStream: Stream[Char] = toVector.toStream
   def toVector: Vector[Char] = self.foldMap(_.toVector)
   override def toString: String = {
-    import syntax.foldable._
-    import Free._
     val sb = new StringBuilder(self.measure)
-    val t = self.traverse_[Trampoline](x => Trampoline.delay(sb ++= x))
-    t.run
+    self.foreach(sb ++= _)
     sb.toString
   }
 


jdegoes:
Yeah, it looks like it was terrible. Nice. 👍 

A slightly faster way to do it would be to `while` loop on the iterator.

It will be interesting seeing if `Cord` is any faster than a `String`. It seems unlikely (at least for some cases) given the inefficiency of representing strings as lists of characters...

fosskers:
`while`ing with an `Iterator` is [about 20% faster](https://github.com/fosskers/scala-benchmarks#folding) than at least a `foldLeft`. Not sure on a single `.map`. `@tailrec` with `IList` is always fastest.

fosskers:
Small law-like property that might be helpful to consider: `.shows` on a `String` should be instantaneous. Currently it's not.

edmundnoble:
@jdegoes Which iterator are you talking about, exactly? The one on `Cord`? @fosskers we aren't using `IList` here, from what I can see.

`Cord` by the way is not a linked structure of characters, but a linked structure of `String`s. The main benefit I believe is the asymptotics of things like `++`, `+:/:+` and `-:/:-`, in addition to the constant-time `length`.

edmundnoble:
Also @fosskers I'm taking a look at the `Show` instance for `String`. It should be overloading `shows`.

fosskers:
What I meant by the `IList` comment was that were some iterating to be done here (I didn't look closely), doing so is always fastest with `IList` and tail recursion. Only `Array` + primitives beats it.

fosskers:
CI failure is spurrious:
```
Err:1 http://ppa.launchpad.net/webupd8team/java/ubuntu trusty/main i386 oracle-java8-installer all 8u151-1~webupd8~0

  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out

Err:1 http://ppa.launchpad.net/webupd8team/java/ubuntu trusty/main amd64 oracle-java8-installer all 8u151-1~webupd8~0

  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out

Err:1 http://ppa.launchpad.net/webupd8team/java/ubuntu trusty/main i386 oracle-java8-installer all 8u151-1~webupd8~0

  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out

Err:1 http://ppa.launchpad.net/webupd8team/java/ubuntu trusty/main amd64 oracle-java8-installer all 8u151-1~webupd8~0

  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out

Err:1 http://ppa.launchpad.net/webupd8team/java/ubuntu trusty/main i386 oracle-java8-installer all 8u151-1~webupd8~0

  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out

Err:1 http://ppa.launchpad.net/webupd8team/java/ubuntu trusty/main i386 oracle-java8-installer all 8u151-1~webupd8~0

  Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out

W: --force-yes is deprecated, use one of the options starting with --allow instead.

E: Failed to fetch http://ppa.launchpad.net/webupd8team/java/ubuntu/pool/main/o/oracle-java8-installer/oracle-java8-installer_8u151-1~webupd8~0_all.deb  Could not connect to ppa.launchpad.net:80 
```

edmundnoble:
Thanks @fosskers, all good now.

TomasMikula:
👍 , but do you plan to forward-port it to 7.3.x, toooo?

jdegoes:
@edmundnoble The `foreach` that's used to build the string in `toString`.

edmundnoble:
@TomasMikula Sure.
@jdegoes Oh I see. I don't believe there really is an iterator for `FingerTree` though. Just a `fold`.

fosskers:
https://github.com/scalaz/scalaz/blob/v7.2.17/core/src/main/scala/scalaz/FingerTree.scala#L826

edmundnoble:
@fosskers That iterator is built using the `fold` function, so I doubt it would be any faster than using `fold` on its own.

fosskers:
Yeah eh.

fosskers:
I don't believe the `Show` instance for `String` was updated, as @edmundnoble and I discussed higher up.

jdegoes:
@fosskers 👍 Can be a separate add-on PR.

PR #1530, opened 2017-11-23
[8.0.x] Convert example module to use tut

NeQuissimus:
Related to #1526 

I took `Forall` as a first example, the code there was pretty good already.
`Maybe` and `Maybe2` should probably be rewritten/expanded upon.

I chose to make `example` a regular module without cross building since we really only need to generate documentation once.

I also updated SBT because I was encountering the bug where you can't exit SBT after a while, except by killing the process (I can't find the issue right now but it was fixed in 1.0.2 IIRC). Anyways, updating SBT should not hurt us...

I will add the content from #1521 once it has been merged.

This is what `sbt example/tut` generates:

---

# Forall

## Instance creation

Create instance using `of` or `mk`

```scala
scala> import scalaz.Prelude._
import scalaz.Prelude._

scala> val nil: ∀[List] = ∀.of[List](Nil)
nil: scalaz.Prelude.∀[List] = List()

scala> val emptyMap: ∀∀[Map] = ∀∀.of[Map](Map())
emptyMap: scalaz.Prelude.∀∀[Map] = Map()

scala> val nil1: ∀[List] = ∀.mk[∀[List]].from(Nil)
nil1: scalaz.Prelude.∀[List] = List()

scala> val emptyMap1: ∀∀[Map] = ∀∀.mk[∀∀[Map]].from(Map())
emptyMap1: scalaz.Prelude.∀∀[Map] = Map()
```

## Universally quantified Semigroup

```scala
scala> import scalaz.Prelude._
import scalaz.Prelude._

scala> import scalaz.typeclass.Semigroup
import scalaz.typeclass.Semigroup

scala> def listSemigroup[A]: Semigroup[List[A]] = new Semigroup[List[A]] {
     |   def append(x: List[A], y: => List[A]) = x ++ y
     | }
listSemigroup: [A]=> scalaz.typeclass.Semigroup[List[A]]

scala> type Plus[F[_]] = ∀[λ[A => Semigroup[F[A]]]]
defined type alias Plus

scala> val listPlus: Plus[List] = ∀.mk[Plus[List]].from(listSemigroup)
listPlus: Plus[List] = $anon$1@4fd9f447

scala> listPlus[Int].append(List(1, 2), List(3, 4)) == List(1, 2, 3, 4)
res0: Boolean = true
```

## Natural transformation

```scala
scala> import scalaz.Prelude._
import scalaz.Prelude._

scala> type ~>[F[_], G[_]] = ∀[λ[A => F[A] => G[A]]]
[...]

PR #1537, opened 2017-11-27
Convert the rest of the examples project

NeQuissimus:
These were committed while #1530 was open, so I am converting them into tut examples here...


example/src/main/tut/typeclass/Show.md
```diff
@@ -0,0 +1,37 @@
...
+**Typical imports**
+
+```tut:silent
+import scalaz.Prelude._
+```
+
+# Built-in instances
+
+```tut
+val stringShow: Show[String] = implicitly
```
hrhino: `Show` has a companion `apply` that looks better IMO

```scala
val stringShow: Show[String] = Show[String]
```

example/src/main/tut/typeclass/Show.md
```diff
@@ -0,0 +1,37 @@
...
+case class Foo(a: Int)
+implicit val fooShow: Show[Foo] = (f: Foo) => s"Foo[${f.a}]"
+```
+
+# Usage
+
+```tut
+val bad = Foo(1).toString
+val good = Foo(1).show
+```
```
hrhino: An example showing that `Show` instances for (say) `Maybe[A]` expects and uses a `Show[A]` rather than just blindly calling `.toString` like `s" ... "` and case class `toString` would be nice.

Even just like in the original `ShowUsage`:

```scala
assert(Maybe.just(Foo(1)).show == "Just(Foo! 1)")
```

NeQuissimus: Oops, I meant to migrate that over!

example/src/main/tut/typeclass/Show.md
NeQuissimus: This was just meant to show that instances for standard types are already (implicitly) available

jdegoes:
@NeQuissimus Nice work!

Issue #1538, opened 2017-11-28
[8.0.x] Maybe2 needs work


NeQuissimus:
I tried using Maybe2.
I understand the benefit of it not creating a Tuple2 instance.
However, since it does not have any type class instances other than `Show` and `IsCovariant` I am not sure how useful it is.
I assume that for the moment, it is a bit of a stub?

- `import scalaz.data.Maybe2._` is needed to create instances, the prelude is not sufficient
- I can't figure out how to access the `Show` instance :)
- without `Monoid`, `Monad` and friends, `Maybe2` is a little "bare"

I may try enhancing `Maybe2` myself, this issue is for tracking purposes mostly.

TomasMikula:
Yeah, `Maybe2` is pretty barebones at the moment. It was introduced to support efficient implementation of `IList` via `Fix`. No bells and whistles were added.

PR #1539, opened 2017-11-28
[8.0.x] example: Add Maybe and Maybe2

NeQuissimus:
Adding more examples. 

`Maybe2` is rather limited given #1538

### DIFF
--- a/example/src/main/tut/data/Maybe.md
+++ b/example/src/main/tut/data/Maybe.md
@@ -6,7 +6,7 @@ title:  "Maybe"
 # Maybe
 
 `Maybe` can either signal the presence or absence of a value and is a safer alternative to `scala.Option`.
-`Maybe` does not provide unsafe method such as `Option.get` and `Option.head`, both of which may throw exceptions.
+`Maybe` does not provide unsafe methods such as `Option.get` and `Option.head`, both of which may throw exceptions.
 
 **Typical imports**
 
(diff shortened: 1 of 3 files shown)


example/src/main/tut/data/Maybe.md
```diff
@@ -0,0 +1,30 @@
+---
+layout: docs
+title:  "Maybe"
+---
+
+# Maybe
+
+`Maybe` can either signal the presence or absence of a value and is a safer alternative to `scala.Option`
```
TomasMikula: Let's try to qualify statements like "is a safer alternative to [...]". For example, add

```
in the sense that it does not have unsafe methods like `Option#get`.
```

example/src/main/tut/data/Maybe.md
```diff
@@ -0,0 +1,31 @@
+---
+layout: docs
+title:  "Maybe"
+---
+
+# Maybe
+
+`Maybe` can either signal the presence or absence of a value and is a safer alternative to `scala.Option`.
+`Maybe` does not provide unsafe method such as `Option.get` and `Option.head`, both of which may throw exceptions.
```
hrhino: > ... does not provide unsafe method[s] ...

example/src/main/tut/data/Maybe.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+## Use the Monad instance
+
+```tut
+val x: Maybe[Long] = just(50L)
+x.flatMap { y =>
+  just(2 * y)
+}
+```
```
hrhino: Could show off `map` and `pure`, too. Also, as it's an opaque type alias, the existence of `Just` and `Empty` extractors may be a useful thing to point out.

NeQuissimus: #1547

jdegoes:
@NeQuissimus Merge conflicts.

jdegoes:
🙏

PR #1541, opened 2017-11-28
[8.0.x] example: Add Functor

NeQuissimus:


TomasMikula:
I think we should at least mention the laws required to qualify as a functor.

jdegoes:
Would be nice to show an example of using syntax for functor... assuming we have that in 8.

NeQuissimus:
I added the laws and a usage example

jdegoes:
@NeQuissimus Looks good, thanks!

PR #1543, opened 2017-11-28
Change the shape of Semigroup#append to what Ops._f1 expects.

### LINKED ISSUE #1542: [8.0.x] Semigroup: broken macro (?)
Alright, so I was trying to write an example for `Semigroup` and I am getting a little frustrated here :D (mostly because I know nothing about macros)

I implemented a simple semigroup for String
```scala
implicit val StringSemigroup = new Semigroup[String] {
  def append(a1: String, a2: => String) = a1 + a2
}
```

Then I grab the usual suspects

```scala
import scalaz.typeclass.Semigroup
import scalaz.typeclass.Semigroup._
```

The second import gives me access to this:

```scala
object SemigroupSyntax {
  class OpsA[A: Semigroup](@silent a: A) {
    def append(f: => A): A = macro meta.Ops._f1[A, A]
  }
}
```

Now watch what happens when I try to append two strings:

```scala
@ "a".append("b") 
cmd11.sc:1: not enough arguments for method append: (a1: String, a2: => String)String.
Unspecified value parameter a2.
val res11 = "a".append("b")
                      ^
Compilation Failed
```

This should work, no? But for the heck of it, let's entertain the second parameter:

```scala
@ "a".append("b", "c") 
cmd11.sc:1: too many arguments (2) for macro method append: (f: => String)String
val res11 = "a".append("b", "c")
                            ^
Compilation Failed
```

So, one parameter is not enough, two is too many?
What's happening here? :D

hrhino:
I have mixed feelings on this shape. It's consistent with other typeclasses but inconsistent with scalaz 7 and cats. The alternative is to add a shape to `Ops` to handle this kind of typeclass method. 

Fixes #1542.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/ConstInstances.scala
+++ b/base/shared/src/main/scala/scalaz/data/ConstInstances.scala
@@ -28,7 +28,7 @@ trait ConstInstances {
   implicit def apply[R](implicit R: Semigroup[R]): Apply[Const[R, ?]] = new Apply[Const[R, ?]] {
     def functor: Functor[Const[R, ?]] = Const.functor[R]
     def ap[A, B](fa: Const[R, A])(f: Const[R, A => B]): Const[R, B] =
-      Const(R.append(fa.getConst, f.getConst))
+      Const(R.append(fa.getConst)(f.getConst))
   }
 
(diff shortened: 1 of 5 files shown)


TomasMikula:
I would add a shape to `Ops` in any case.

hrhino:
brb with this PR

PR #1544, opened 2017-11-28
[8.0.x] example: Add Semigroup

NeQuissimus:
Depends on #1546

### DIFF
--- a/example/src/main/tut/typeclass/Semigroup.md
+++ b/example/src/main/tut/typeclass/Semigroup.md
@@ -1,0 +1,33 @@
+---
+layout: docs
+title:  "Semigroup"
+---
+
+# Semigroup
+
+*A semigroup is an algebraic structure consisting of a set and an associative binary operation.*
+
+A semigroup instance needs to satisfy the following law:
+
+- Associativity: `append(x, append(y, z)) === append(append(x, y), z)`
+
+**Typical imports**
+```tut:silent
+import scalaz.typeclass.Semigroup
+import scalaz.typeclass.Semigroup._
+```
+
+## Instance declaration
+
+```tut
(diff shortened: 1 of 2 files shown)


example/src/main/tut/typeclass/Semigroup.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: docs
+title:  "Semigroup"
+---
+
+# Semigroup
+
+**Category Theory**
+
+*A semigroup is an associative magma. It contains a set and an associative binary operation.*
```
hrhino: Mentioning magmas seems a little strange without a definition. A link to wiki/ncatlab may be welcome.

I'm not sure if it "contains" them; maybe
> ... magma. That is, it is a set along with an associative binary operation.

example/src/main/tut/typeclass/Semigroup.md
```diff
@@ -0,0 +1,35 @@
...
+
+A semigroup instance is required to satisfy associativity such that the following holds true:
```
hrhino: "be associative" or "satisfy the associativity law"

example/src/main/tut/typeclass/Semigroup.md
TomasMikula: Unless we have magma in scalaz, I would avoid mentioning it altogether.

example/src/main/tut/typeclass/Semigroup.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: docs
+title:  "Semigroup"
+---
+
+# Semigroup
+
+**Category Theory**
```
TomasMikula: Semigroup is not really from category theory.

jdegoes:
🙏

PR #1546, opened 2017-11-29
Modernize meta.Ops, adapt it to new useful shapes, and use it more.

### LINKED ISSUE #1542: [8.0.x] Semigroup: broken macro (?)

hrhino:
Since 8.0.x won't support scala 2.10, there's no reason to bother using raw trees and `c.Expr`s all over the place. With this, it's possible to drop all the type parameters and write the impls in a somewhat less esoteric style (for when more shapes are inevitably needed).

The ops macros need to be blackbox, otherwise `f_0` won't fix the result type based on its type argument, and infer `Nothing`. It's not trivial to return a properly-typed tree if the return type is an arbitrary function of a cloud of types floating over the macro application, so it's easier to just accept the blackboxity and go ahead with unrestricted type inference.

Fixes #1542, so a ping to @NeQuissimus.

### DIFF
--- a/meta/shared/src/main/scala/scalaz/meta/Ops.scala
+++ b/meta/shared/src/main/scala/scalaz/meta/Ops.scala
@@ -1,35 +1,65 @@
 package scalaz
 package meta
 
-import scala.reflect.macros.whitebox.Context
+import scala.reflect.macros.blackbox
 
 // Originally inspired by http://typelevel.org/blog/2013/10/13/spires-ops-macros.html
 
-class Ops(val c: Context) {
+/** Zero-cost syntax ops macros.
+  *
+  * Inside of a syntax class, these macros can be used as the definition of
+  * extension methods to entirely eliminate the allocation of the syntax class
+  * and the implicit conversion.
+  *
+  * After defining
+  * {{{
+  * class Ops[A: Typeclass](a: A) {
+  *   def method[B](b: B): Result = macro Ops.f_1
+  * }
+  * }}}
+  *
+  * then calling `a.method(b)` (if `a`'s type has a `Typeclass` instance) will
+  * expand to `tcA.method(a)(b)` (where `tcA : Typeclass[A]` is the implicit
+  * evidence).
+  *
+  * Due to the way that scala.reflect macros work, there is no shape-polymorphic
+  * macro implementation. The methods are named according to the shape of the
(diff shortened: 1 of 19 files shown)


TomasMikula:
Looks great! 👍

meta/shared/src/main/scala/scalaz/meta/Ops.scala
```diff
@@ -1,35 +1,66 @@
...
+  /* def method(f: A)(g: B): R */
+  def fa_1_1(f: Tree)(g: Tree): Tree = q"$ev.$name($lhs, $f)($g)"
+
+  /* def method(f: A, g: B): R */
+  def f_2(f: Tree, g: Tree): Tree = q"$ev.$name($lhs)($f, $g)"
+
+  /** Destructured macro application.
+    * - `ev`: the typeclass evidence
+    * - `name`: the name of the invoked method
+    * - `targs`: the inferred type arguments of the method call
```
hrhino: Yikes, missed this last night. Will remove.

NeQuissimus:
YES!

jdegoes:
@hrhino Nice! Ready to merge whenever you are. Thanks!

PR #1548, opened 2017-11-30
Add macro shape for OpsA-style syntax.

hrhino:
This allows for such syntax as `f.pure[F]` and (eventually) `err.raiseError[M]`.

After this I'll leave the macros alone for a while.


jdegoes:
@hrhino Nice!

PR #1549, opened 2017-11-30
Improved `Show` instance for String

fosskers:
Since there is no `Read` typeclass in Scalaz (nor should there be), `Show` becomes nearly lawless. That's okay if it's only used for debugging and never intended for serious serialization.

That said, here's an attempt at one law, which is reflected in the code change and alters the output of `.shows`:

> Law of String Identity: `String.shows == id`

### DIFF
--- a/core/src/main/scala/scalaz/std/String.scala
+++ b/core/src/main/scala/scalaz/std/String.scala
@@ -6,7 +6,7 @@ trait StringInstances {
     type SA[α] = String
     def append(f1: String, f2: => String) = f1 + f2
     def zero: String = ""
-    override def shows(f: String): String = f
+    override def shows(f: String): String = '"' + f + '"'
     def order(x: String, y: String) = Ordering.fromInt(x.compareTo(y))
     override def equal(x: String, y: String) = x == y
     override def equalIsNatural: Boolean = true
--- a/core/src/main/scala/scalaz/std/String.scala
+++ b/core/src/main/scala/scalaz/std/String.scala
@@ -6,6 +6,7 @@ trait StringInstances {
     type SA[α] = String
     def append(f1: String, f2: => String) = f1 + f2
     def zero: String = ""
+    override def show(f: String): Cord = shows(f)
     override def shows(f: String): String = '"' + f + '"'
     def order(x: String, y: String) = Ordering.fromInt(x.compareTo(y))
     override def equal(x: String, y: String) = x == y
--- a/core/src/main/scala/scalaz/std/String.scala
+++ b/core/src/main/scala/scalaz/std/String.scala
@@ -6,7 +6,7 @@ trait StringInstances {
     type SA[α] = String
     def append(f1: String, f2: => String) = f1 + f2
     def zero: String = ""
-    override def show(f: String): Cord = shows(f)
+    override def show(f: String): Cord = new Cord(FingerTree.three("\"", f, "\"")(Cord.sizer).toTree)
     override def shows(f: String): String = '"' + f + '"'
     def order(x: String, y: String) = Ordering.fromInt(x.compareTo(y))
     override def equal(x: String, y: String) = x == y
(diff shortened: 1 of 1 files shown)


core/src/main/scala/scalaz/std/String.scala
```diff
@@ -6,7 +6,7 @@ trait StringInstances {
     type SA[α] = String
     def append(f1: String, f2: => String) = f1 + f2
     def zero: String = ""
-    override def show(f: String) = '"' + f + '"'
+    override def shows(f: String): String = f
```
fosskers: Notice the change from `show` to `shows`.

edmundnoble: I don't think it's a good idea to change behavior like this in 7.2.x. I do agree that this is how it should've been from the start, but that's beside the point.

fosskers:
Lemme do a `publishLocal` and see. Right now we're looking at:
```
| Benchmark                               | ScalaZ 7.2.16 | ScalaZ 7.2.17 | Cats 1.0.0-RC1 | Vanilla Scala | Haskell 8.0.2 |
|-----------------------------------------+---------------+---------------+----------------+---------------+---------------|
| ~Show~ - ~String~                       | 216.6         |               | 3.2            | 2.8           | 199.4         |
```

fosskers:
```
[info] ShowBench.showStrCats      avgt    5       3.308 ±     0.194  ns/op
[info] ShowBench.showStrScalaz    avgt    5       3.749 ±     0.104  ns/op
[info] ShowBench.showStrVanilla   avgt    5       2.877 ±     0.159  ns/op
```

fosskers:
The only other discussion point is whether removing the surrounding quotes is okay. Haskell only adds them because its `Read` instance for `String` requires it.

fommil:
definitely need the quotes around `String`. @edmundnoble I'm sure there's a way of making the change such that it's compatible with 7.2? Why not implement both methods?

fosskers:
```
[error]  * method show(java.lang.String)scalaz.Cord in object scalaz.std.StringInstances#stringInstance's type is different in current version, where it is (java.lang.Object)scalaz.Cord instead of (java.lang.String)scalaz.Cord
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("scalaz.std.StringInstances#stringInstance.show")
```
Zing, I'll add back an explicit `def show`.

edmundnoble:
@fommil I'm pretty sure we're good now.

fosskers:
Gross eh? That's another thing that threw me off.

fommil:
I'd still have thought it'd be better to implement both methods independently. Fosskers you don't need to publish local to do perftests, just patch the monkey for great good.

fommil:
Some numbers...


```
# scalaz 7.2.17
ShowBench.showStrScalaz  avgt   15  2167.086 ± 59.773  ns/op

# override def show(f: String): Cord = Cord.stringToCord(shows(f))
# override def shows(f: String): String = "\"" + f + "\""
[info] ShowBench.showStrScalaz  avgt   15  97.004 ± 1.336  ns/op

# override def show(f: String): Cord = new Cord(FingerTree.three("\"", f, "\"")(Cord.sizer).toTree)
# override def shows(f: String): String = "\"" + f + "\""
[info] ShowBench.showStrScalaz  avgt   15  96.869 ± 1.246  ns/op
```

this is the standard .show test in the https://github.com/fosskers/scalaz-and-cats

we can safely conclude that either approach is equivalent in terms of the cost to construct the Cord, but the latter has the best memory overhead so I think it is best. Consider the case where the strings are *huge*. I'll update the test to do that and also consider nested in a List.

fommil:
I'm pretty amazed about the final result... doing `.toString` on a single `Cord` appears to do a *lot* more work than a no-op.

fommil:
Needs a sync.

I forgot to check perf of this version with mixed char/string concatenation. Is it ok? My concern is that it's going via the anyToString implicit.

fommil:
Oh, I'm also not sure if the fingertree is doing the right then, I never actually tested/printed it out. I assume it has an order preserving constructor.

fommil:
yup, it's used correctly

```
scala> val cord = new Cord(FingerTree.three("\"", "hello", "\"")(Cord.sizer).toTree)
cord: scalaz.Cord = "hello"
```

Issue #1526, opened 2017-11-22
Scalaz 8 Roadmap


jdegoes:
# Checklist

 - [ ] **Core**
    - [ ] **Type**
      - [x] **Leibniz**
      - [x] **Liskov**
      - [ ] **AList**
      - [ ] ...
    - [ ] **Data**
      - [ ] **List**
      - [x] **Maybe**
      - [x] **Disjunction**
      - [ ] **Map**
      - [ ] ...
    - [ ] **Functor Hierarchy**
      - [ ] ...
    - [ ] **MTL**
      - [ ] ...
    - [ ] **Algebra**
      - [ ] ...
    - [ ] **Other CT**
      - [ ] ...
 - [ ] **Syntax**
    - [ ] **Zero-cost, by-name syntax macros**
    - [ ] ...
 - [ ] **Effect**
    - [x] **IO**
    - [x] **MVar**
    - [x] **RTS**
    - [ ] **MonadIO**
    - [ ] **MonadFork**
    - [ ] **MonadBracket**
    - [ ] **STM**
    - [ ] **TVar**
    - [ ] ...
 - [ ] **Example**
    - [ ] ...
 - [ ] **Test**
    - [ ] ...
 - [ ] **Benchmark**
    - [x] scalaz.effect.IO
    - [ ] scalaz.effect.MVar
 - [ ] **Design Guide**

# Overview

### Data

**Needed**: Someone to spearhead this work.

### Functor Hierarchy

The Scalaz 8 functor hierarchy needs to be modularized to minimize the number of new abstractions needed to capture distinctions between different types of functors.

While the scope and nature of these changes is TBD, one useful suggestion by @ekmett is to factor out `Productive` and `Coproductive` classes:

```scala
trait Productive[F[_]] {
  def product[A, B](fa: F[A], fb: F[B]): F[(A, B)]
  def unit: F[Unit]
}
```

```scala
trait Coproductive[F[_]] {
  def coproduct[A, B](fa: F[A], fb: F[B]): F[A \/ B]
  def counit: F[Void]
}
```

These abstractions permit more sharing between, for example, an invariant applicative and a covariant applicative, reducing the number of type class variations that would otherwise have to exist.

Similarly, modeling `Bind` as `F[F[A]] => F[A]` (rather than `(F[A], A => F[B]) => F[B])`) allows the question of arrow type to be deferred and isolated to `map` in the corresponding type of functor.

**Needed**: Someone to spearhead this work.

### MTL 

[...]

PR #1551, opened 2017-12-01
adding Magma and Group

caente:
First stab at expanding the Algebra hierarchy.
Adding:
- Magma
- Group
Made `Semigroup` to be composed of Magma, which implied that now `Semigroup` does not have a "native" `append`. This seems awkward, since now some parts of the code must do:
```
Semigroup[A].magma.append
```
On the other hand, the above could be solved using syntax sugar. It wasn't very clear how Semigroup syntax should work, so I didn't use it on the first commit.

### DIFF
--- a/base/shared/src/main/scala/scalaz/typeclass/Group.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/Group.scala
@@ -1,0 +1,16 @@
+package scalaz
+package typeclass
+
+/* A Group is a Monoid with negation and inverse, such that
+ * a |+| inverse(a)  = monoid.empty
+ * inverse(a) |+| a  = monoid.empty
+ *  
+ * */
+trait Group[A] {
+  def monoid: Monoid[A]
+  def inverse(a: A): A
+}
+
+object Group extends GroupInstances {
+  def apply[A](implicit A: Group[A]): Group[A] = A
+}
--- a/base/shared/src/main/scala/scalaz/typeclass/GroupInstances.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/GroupInstances.scala
@@ -1,0 +1,12 @@
+package scalaz
+package typeclass
+
+trait GroupInstances {
+
+  implicit val int: Group[Int] = new GroupClass[Int] {
+    def append(a1: Int, a2: => Int) = a1 + a2
+    def empty = 0
+    def inverse(a:Int) = a * -1
+  }
+
+}
--- a/base/shared/src/main/scala/scalaz/typeclass/Magma.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/Magma.scala
@@ -1,0 +1,12 @@
+package scalaz
+package typeclass
+
+/* A Magma is a Set with a binary operation that returns a member of the same Set
+ * */
+trait Magma[A] {
+  def append(a1: A, a2: => A): A
+}
+
+object Magma extends MagmaInstances {
+  def apply[A](implicit A: Magma[A]): Magma[A] = A
+}
--- a/base/shared/src/main/scala/scalaz/typeclass/Semigroup.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/Semigroup.scala
@@ -1,8 +1,11 @@
 package scalaz
 package typeclass
 
+/**
+ * A Semigroup is a Magma with associativity
+ */
 trait Semigroup[A] {
-  def append(a1: A, a2: => A): A
+  def magma:Magma[A]
 }
 
 object Semigroup extends SemigroupSyntax {
--- a/base/shared/src/main/scala/scalaz/typeclass/SemigroupClass.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/SemigroupClass.scala
@@ -1,6 +1,6 @@
 package scalaz
 package typeclass
 
-trait SemigroupClass[A] extends Semigroup[A]{
+trait SemigroupClass[A] extends Semigroup[A] with MagmaClass[A]{
   final def semigroup: Semigroup[A] = this
 }
--- a/base/shared/src/main/scala/scalaz/data/ConstInstances.scala
+++ b/base/shared/src/main/scala/scalaz/data/ConstInstances.scala
@@ -28,7 +28,7 @@ trait ConstInstances {
   implicit def apply[R](implicit R: Semigroup[R]): Apply[Const[R, ?]] = new Apply[Const[R, ?]] {
     def functor: Functor[Const[R, ?]] = Const.functor[R]
     def ap[A, B](fa: Const[R, A])(f: Const[R, A => B]): Const[R, B] =
-      Const(R.append(fa.getConst, f.getConst))
+      Const(R.magma.append(fa.getConst, f.getConst))
   }
 
   implicit def applicative[R](implicit R: Monoid[R]): Applicative[Const[R, ?]] = new Applicative[Const[R, ?]] {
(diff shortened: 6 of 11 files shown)


caente:
The failing tests are because the `Semigroup.meta.append` change, but again, I'm not sure what's going on at `SemigroupSyntax`

vmarquez:
Does Group have any laws?  I don't know if we need it... (non rhetorical, curious to see what people think).  cc @TomasMikula @jdegoes ?

base/shared/src/main/scala/scalaz/typeclass/Group.scala
```diff
@@ -0,0 +1,16 @@
+package scalaz
+package typeclass
+
+/* A Group is a Monoid with negation and inverse, such that
+ * a |+| inverse(a)  = monoid.empty
+ * inverse(a) |+| a  = monoid.empty
```
hrhino: Aren't these `Group` laws? @vmarquez

TomasMikula:
You mean does magma have any laws? It doesn't. I'm not terribly interested
in having it.

base/shared/src/main/scala/scalaz/typeclass/GroupInstances.scala
```diff
@@ -0,0 +1,12 @@
+package scalaz
+package typeclass
+
+trait GroupInstances {
+
+  implicit val int: Group[Int] = new GroupClass[Int] {
```
hrhino: This implies that `Int`'s (semi)group/monoid instance is (0, +), not (1, *). 

Is that intended?

base/shared/src/main/scala/scalaz/typeclass/Magma.scala
```diff
@@ -0,0 +1,12 @@
+package scalaz
+package typeclass
+
+/* A Magma is a Set with a binary operation that returns a member of the same Set
+ * */
+trait Magma[A] {
+  def append(a1: A, a2: => A): A
+}
```
hrhino: Surely this is lawless. What good does it do us to have a `Magma` that's not a `Semigroup`?

base/shared/src/main/scala/scalaz/typeclass/SemigroupClass.scala
```diff
@@ -1,6 +1,6 @@
 package scalaz
 package typeclass
 
-trait SemigroupClass[A] extends Semigroup[A]{
+trait SemigroupClass[A] extends Semigroup[A] with MagmaClass[A]{
```
hrhino: These subclassings need to be reflected in `BaseHierarchy`.

vmarquez:
Yeah, i'm of the opinion that if an algebraic structure doesn't have laws, let's ditch it.

base/shared/src/main/scala/scalaz/typeclass/Group.scala
caente: these are the laws, as per @jdegoes  comment https://github.com/scalaz/scalaz/issues/1526#issuecomment-347657494 we will define laws later on

caente:
I completely agree that Magma seems unnecessary, and it adds awkwardness to the code...

base/shared/src/main/scala/scalaz/typeclass/GroupInstances.scala
caente: good question, I have no answer for it, except that perhaps it is not time yet to add instances

base/shared/src/main/scala/scalaz/typeclass/Group.scala
hrhino: Yeah. I wrote this after not realizing that "Does Group have any laws?" was intended to mention "Magma" instead.

base/shared/src/main/scala/scalaz/typeclass/GroupInstances.scala
TomasMikula: We sure could use some tagging that @S11001001 worked on in Scalaz 7.

base/shared/src/main/scala/scalaz/typeclass/SemigroupClass.scala
caente: even if we consider removing Magma, I added it there already
https://github.com/scalaz/scalaz/pull/1551/commits/316260023ea769db3598e9b9883f10e850b69f36

caente:
the only "law" that Magma must enforce is that the result of the binary operation belongs to the same set... which I would think is already given by the type system, will remove it.

vmarquez:
@caente is this a WIP?  If so, can you label it/change the name?  When you're ready for a review/merge, can you squash the commits too? thanks :)

caente:
@vmarquez I was hopping to have smaller PRs, but this is the first time doing this kind of things, and I'm not used to the slow pace... name changed, and definitely will squash commits.

caente:
I'm actually waiting for some feedback regarding Rings @vmarquez @jdegoes @hrhino @TomasMikula

base/shared/src/main/scala/scalaz/typeclass/AbelianGroupInstances.scala
```diff
@@ -0,0 +1,12 @@
+package scalaz
+package typeclass
+
+trait AbelianGroupInstances {
+
+  implicit val int: AbelianGroup[Int] = new AbelianGroupClass[Int] {
```
hrhino: So if you have this here it's probably going to clash with `GroupInstances#int` if you ask for `Group[Int]`. You should (assuming we intend to provide a `Group[Int]` at all) keep this and remove the other.

caente:
also @jdegoes mentioned Commutative as a typeclass, but it seems to me that it should "just" be a law? I'd still want to understand why are we deferring laws implementation? Is it lack of consensus?

base/shared/src/main/scala/scalaz/BaseHierarchy.scala
```diff
@@ -12,6 +12,10 @@ object BaseHierarchy {
     implicit def monadApply[M[_]](implicit M: Monad[M]): Apply[M] = M.applicative.apply
     implicit def monadFunctor[M[_]](implicit M: Monad[M]): Functor[M] = M.applicative.apply.functor
     implicit def monoidSemigroup[A](implicit A: Monoid[A]): Semigroup[A] = A.semigroup
+    implicit def groupMonoid[A](implicit A:Group[A]):Monoid[A] = A.monoid
+    implicit def abelianGroupGroup[A](implicit A:AbelianGroup[A]):Group[A] = A.group
+    implicit def ringAbelianGroup[A](implicit A:Ring[A]):AbelianGroup[A] = A.abelianGroup
+    implicit def ringMonoid[A](implicit A:Ring[A]):Monoid[A] = A.monoid
```
hrhino: This worries me; a `Ring` contains two distinct `Monoid`s, and this picks the "multiplicative" one. However, `groupMonoid compose abelianGroupGroup compose ringAbelianGroup` _also_ produces a `Monoid` from a `Ring`.  This will cause the somewhat counter-intuitive behavior where asking for `Monoid[A]` gives you the multiplicative one, but calling a method that asks for `Group[A]` and then internally uses `Monoid[A]` will find the additive one.

I think we should have a rule such as

```haskell
newtype Mult a = Mult a
instance Ring a => Monoid (Mult a) where
  ...
```

using the (as yet forthcoming) tag types.

base/shared/src/main/scala/scalaz/typeclass/Semigroup.scala
```diff
@@ -1,6 +1,11 @@
 package scalaz
 package typeclass
 
+/**
+ * A Semigroup is a binary opertion such that:
+ * 1 - The result is in the same Set/Type
```
hrhino: This seems a little extraneous, as a quick glance at the type of `append` tells you exactly this. I'd opt for terseness and say

> A semigroup is a type equipped with an associative binary operation

Also, `opertion` is misspelled...

hrhino:
I think the lack of laws has been due to a lack of bandwidth. I was under the impression that scalaprops had been decided on as the law checking framework. I'd like to go poke at that some this week, if I find time.

jdegoes:
@caente We don't have machinery to encode laws just yet. We need a data structure such as:

```scala
final case class IsEqual[A](lhs: A, rhs: A)
```

Then all the "laws" can return data.

It's OK to have a type class that merely adds another law to an existing structure (e.g. `Commutative`).

I will review in more depth later.

base/shared/src/main/scala/scalaz/typeclass/AbelianGroupInstances.scala
caente: yeah, I agree with your above comment, these instances belong to examples, will move them

edmundnoble:
I'm going to try to revive this tomorrow.

PR #1552, opened 2017-12-01
Functor.fmap

fosskers:
This PR adds a `.fmap` method to `Functor` syntax as an alias of `.map`. 

This `.fmap` (unlike the default `.map`) is injectable, exposing lawful Functorness to implementing classes (i.e. `Map` has a built-in `.map` that breaks Functor laws, but its injected `.fmap` would be lawful). This also benefits authors of new Functor instances.


TomasMikula:
Just like with `map`, this will only work if the class doesn't have a method named `fmap`.

Also, if you know you have a `Map`, you can use `mapValues`. On the other hand, if at runtime you have a `Map` but that's not known at compile time, then you can use `map` from `Functor` syntax and there's no conflict.

(I'm not against this, I'm just not convinced by these arguments for it.)

fosskers:
`Map.mapValues` is deeply evil. It looks like a pure transformation over an immutable structure, but it isn't.

fosskers:
```
scala> import scala.util.Random
import scala.util.Random

scala> val m: Map[Int, String] = Map(1 -> "hi", 2 -> "there", 3 -> "you")
m: Map[Int,String] = Map(1 -> hi, 2 -> there, 3 -> you)

scala> m.mapValues(_ => Random.nextInt())
res1: scala.collection.immutable.Map[Int,Int] = Map(1 -> 473085757, 2 -> -2079451500, 3 -> -831712548)

scala> res1
res2: scala.collection.immutable.Map[Int,Int] = Map(1 -> 2131490836, 2 -> -196439511, 3 -> -708440041)

scala> res1
res3: scala.collection.immutable.Map[Int,Int] = Map(1 -> 887266213, 2 -> -388995073, 3 -> -270043264)
```
Let the horror consume you.

TomasMikula:
OMG! 😱 

But it was _you_ who used a non-referentially transparent function. 😉 

Anyway, for any name we choose for syntax extension, there is a potential collision with something a data type will already have.

fosskers:
> But it was you who used a non-referentially transparent function. 

True, true, but nobody could claim they expected this to happen. This bit us *hard* in production about a half-year back 😭 

> there is a potential collision with something a data type will already have.

Also true, although you must admit that the likelihood of `.map` preexisting is high, while `.fmap` is low.

TomasMikula:
Yeah, I certainly didn't expect that.

I guess it doesn't hurt anyone to add this.

fosskers:
I fix it up tomorrow morning.

fosskers:
How's that?

PR #1554, opened 2017-12-03
[8.0.x] example: Add Applicative

NeQuissimus:
- Remove the `Category Theory` headings. We can figure out links to literature or so later
- Add `map` to the Applicative syntax
- Add an Applicative doc

I am also opening another ticket because when you look at the implementation of `ApplicativeSyntax.Ops`, it is rather strange to hop through all the "embedded" instances.

### DIFF
--- a/example/src/main/tut/typeclass/Applicative.md
+++ b/example/src/main/tut/typeclass/Applicative.md
@@ -10,7 +10,7 @@ title:  "Applicative"
 **Typical imports**
 
 ```tut:silent
-import scalaz.Prelude._
+import scalaz.Scalaz._
 ```
 
 ## Instance declaration
@@ -27,11 +27,17 @@ implicit val listap: Applicative[List] = new ApplicativeClass[List] {
 
 ## Usage
 
-```tut
+```tut:reset
+import scalaz.Scalaz._
 
 val l: List[Int] = 123.pure[List]
 
-implicit val f: Functor[List] = listap.apply.functor
-
 l.map(_ + 25)
+
+val l2 = List(17, 19, 23)
+
+val fs: List[Int => Int] = List(_ * 2, _ * 3, _ * 4, _ * 5)
+
+l2.ap(fs)
 ```
+
--- a/example/src/main/tut/typeclass/Applicative.md
+++ b/example/src/main/tut/typeclass/Applicative.md
@@ -1,0 +1,37 @@
+---
+layout: docs
+title:  "Applicative"
+---
+
+# Applicative
+
+*Whereas a [functor](./Functor.html) allows application of a pure function to a value in a context, an Applicative also allows application of a function in a context to a value in a context.*
+
+**Typical imports**
+
+```tut:silent
+import scalaz.Prelude._
+```
+
+## Instance declaration
+
+```tut
+import scalaz.typeclass.ApplicativeClass
+
+implicit val listap: Applicative[List] = new ApplicativeClass[List] {
+  def pure[A](a: A): List[A] = List(a)
+  def ap[A, B](fa: List[A])(f: List[A => B]): List[B] = fa.zip(f).map(t => t._2(t._1))
+  def map[A, B](ma: List[A])(f: A => B): List[B] = ma.map(f)
+}
+```
+
+## Usage
+
+```tut
+
+val l: List[Int] = 123.pure[List]
+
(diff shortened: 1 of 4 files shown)


base/shared/src/main/scala/scalaz/typeclass/ApplicativeSyntax.scala
```diff
@@ -8,10 +8,15 @@ import language.experimental.macros
...
+  implicit def applicativeOps[A, F[_]: Applicative](fa: F[A]): ApplicativeSyntax.Ops[A, F] = new ApplicativeSyntax.Ops(fa)
 }
 
 object ApplicativeSyntax {
   class OpsA[A](@silent a: A) {
     def pure[F[_]: Applicative]: F[A] = macro meta.IdOps.id_1
   }
+
+  class Ops[A, F[_]](@silent fa: F[A])(implicit F: Applicative[F]) {
+    def map[B](f: A => B): F[B] = F.apply.functor.map(fa)(f)
```
hrhino: Shouldn't we be getting this from `FunctorOps` instead?

example/src/main/tut/typeclass/Functor.md
```diff
@@ -5,10 +5,6 @@ title:  "Functor"
 
 # Functor
 
-The `Functor` type class represents a type that can be mapped over.
```
hrhino: I think this is a more useful line to keep than the one about "(mathematical) functor"s.

base/shared/src/main/scala/scalaz/typeclass/ApplicativeSyntax.scala
NeQuissimus: Well, for that you would need to import all the syntax ops one-by-one...

example/src/main/tut/typeclass/Functor.md
NeQuissimus: I can agree with that :)

base/shared/src/main/scala/scalaz/typeclass/ApplicativeSyntax.scala
NeQuissimus: Oh, because they would override each other?!

hrhino: Neither wins; they ambiguate each other.

The fact that `Applicative` is a sub-typeclass of `Functor` isn't known to the compiler, and it sees two applicable implicit conversions from `F[A]` to something with a `map` method, and gives up.

NeQuissimus: That is, if I have both an implicit `Functor` and an implicit `Applicative` in scope, correct?
I guess the issue is a much bigger one then, since we somehow need to tell the compiler that is can extract a functor from an applicative (or rather applicative -> apply -> functor).
I will take the syntax back out. Clearly, more thought needs to be invested here.

hrhino: We tell the compiler that via `BaseHierarchy` right now, whereby it knows that given an `Applicative` it can have a `Functor`.

NeQuissimus:
Rebased, resolved conflicts

example/src/main/tut/typeclass/Applicative.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## Usage
+
+```tut:reset
+import scalaz.Scalaz._
+
+val l: List[Int] = 123.pure[List]
+
+l.map(_ + 25)
```
jdegoes: One more example here using `ap` would be nice.

example/src/main/tut/typeclass/Monoid.md
```diff
@@ -5,8 +5,6 @@ title:  "Monoid"
 
 # Monoid
 
-**Category Theory**
```
jdegoes: **Abstract Algebra**

NeQuissimus: I removed all those references, maybe we can add the academic notes back later.

example/src/main/tut/typeclass/Applicative.md
NeQuissimus: I shall add one, that would make sense given this is **Applicative** after all xD

PR #1558, opened 2017-12-04
[8.0.x] example: Add Monad

NeQuissimus:
Depends on #1554

### DIFF
--- a/example/src/main/tut/typeclass/Monad.md
+++ b/example/src/main/tut/typeclass/Monad.md
@@ -1,0 +1,46 @@
+---
+layout: docs
+title:  "Monad"
+---
+
+# Monad
+
+*A monad is the combination of an Applicative with a Bind*
+
+A monad instance needs to satisfy the monad laws in addition to those defined by [Applicative](./Applicative.html) and [Bind](./Bind.html):
+
+- Left identity: `pure(a).flatMap(f) === f(a)`
+- Right identity: `x.flatMap(pure) === x`
+- Associativity: `x.flatMap(f).flatMap(g) === m.flatMap(x => f(x).flatMap(g))`
+
+**Typical imports**
+
+```tut:silent
+import scalaz.Scalaz._
+```
+
+## Instance declaration
+
+```tut
+import scalaz.typeclass.MonadClass
+
+val listMonad: Monad[List] = new MonadClass[List] {
+  def pure[A](a: A): List[A] = List(a)
+  def ap[A, B](fa: List[A])(f: List[A => B]): List[B] = fa.zip(f).map(t => t._2(t._1))
+  def flatMap[A, B](ma: List[A])(f: A => List[B]): List[B] = ma.flatMap(f)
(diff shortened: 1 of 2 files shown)


example/src/main/tut/typeclass/Monad.md
```diff
@@ -0,0 +1,49 @@
+---
+layout: docs
+title:  "Monad"
+---
+
+# Monad
+
+*A monad is the combination of an [Applicative](./Applicative.html) with a [Bind](./Bind.html)*
+
+A monad instance needs to satifsy the monad laws:
```
hrhino: > ... in addition to the laws for `Applicative` and `Bind`

... but does `Bind` have laws?

also, "satisfy".

example/src/main/tut/typeclass/Monad.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+## Instance declaration
+
+```tut
+import scalaz.typeclass.MonadClass
+
+val listMonad: Monad[List] = new MonadClass[List] {
+  def pure[A](a: A): List[A] = List(a)
+  def ap[A, B](fa: List[A])(f: List[A => B]): List[B] = fa.zip(f).map(t => t._2(t._1))
```
hrhino: I feel a little bad that we don't have a default `ap` lying around, but eh.

example/src/main/tut/typeclass/Monad.md
```diff
@@ -0,0 +1,49 @@
...
+
+```tut
+val l = List(1, 2, 3).flatMap(x => List(x + 5))
+val k = List(2, 3, 4)
+
+for {
+  x <- l
+  y <- k
+} yield x * y
+```
```
hrhino: 👍 for the `for` syntax example!

hrhino: although... isn't this just using List#flatMap? You should wrap this in a def example[F[_]: Monad] or similar.

NeQuissimus: Yeah, the reason I am using `List` is that I can later just replace all the types with `IList` and the examples should still work.

hrhino: Well, but... since scala knows that the receiver of `flatMap` is a `List`, it'll pick the instance method `flatMap` and not the typeclass syntax one (so this isn't even demoing anything scalaz!)

example/src/main/tut/typeclass/Monad.md
NeQuissimus: I am unsure whether `Bind` has laws.

example/src/main/tut/typeclass/Monad.md
NeQuissimus: I know :)
But it is way easier to use `List` for the moment and replace it with `IList` later than use one of our data types right now. Most of them are still missing good chunks of syntax or type class instances.

NeQuissimus:
Rebased onto upstream changes

Issue #1449, opened 2017-09-03
I heard you like Enum, so I made an Enum for your Enum


fommil:
it would be pretty nice to have a `scalaz.Enum` for https://docs.oracle.com/javase/8/docs/api/java/lang/Enum.html

It would allow things like

```
java.time.DayOfWeek.MONDAY |-> 2
```

look like there is only an `Equal` instance.

jdegoes:
What do you think of [the PureScript version](https://github.com/purescript/purescript-enums) as inspiration? Java 8 enums would have an efficient implementation and could also provide automatic Show.

fommil:
Nice! I guess I'm even thinking ahead to the possibility of having `sealed trait` of `object`s converted into `enum`s at the bytecode level.

PR #1559, opened 2017-12-05
Series/7.2.x enum typeclass evidence for java enum

merlinorg:
May mima shine upon me. #1449


xuwei-k:
not binary compatible

PR #1562, opened 2017-12-07
 Add "safely" to BindRec

puffnfresh:


tonymorris:
:+1:

core/src/main/scala/scalaz/BindRec.scala
```diff
@@ -47,9 +49,40 @@ trait BindRec[F[_]] extends Bind[F] { self =>
...
+            def apply[A](fa: F[A]): Free[F, A] =
+              Free.liftF(fa)
+          }
+      }
+
+    iso[M].from(f(iso[M]))
+  }
+
+  def traverse_[F[_], M[_], A](fa: F[A])(f: A => M[Unit])
+               (implicit F: Foldable[F], M: BindRec[M], MA: Applicative[M]): M[Unit] =
```
TomasMikula: I think this method should be on `Foldable` (called perhaps `traverseRec_`). The default implementation might use translation to and from `Free` as is done here, but for many concrete foldable structures `traverseRec_` might be just `traverse_`.

core/src/main/scala/scalaz/BindRec.scala
```diff
@@ -47,9 +49,40 @@ trait BindRec[F[_]] extends Bind[F] { self =>
   val bindRecSyntax = new scalaz.syntax.BindRecSyntax[F] { def F = BindRec.this }
 }
 
+trait MonadRecSafely[M[_], A] {
+  def apply[S[_]: BindRec: Applicative](lift: M <~> S): S[A]
```
TomasMikula: Is `BindRec` constraint intentional here (as opposed to just `Bind`)? Isn't this supposed to represent an operation implemented against the `Monad` interface, which is potentially unsafe if the monad is not lazy? On that note, if it represents a potentially unsafe operation, `MonadRecSafely` doesn't seem like a suitable name.

core/src/main/scala/scalaz/BindRec.scala
```diff
@@ -47,9 +49,40 @@ trait BindRec[F[_]] extends Bind[F] { self =>
   val bindRecSyntax = new scalaz.syntax.BindRecSyntax[F] { def F = BindRec.this }
 }
 
+trait MonadRecSafely[M[_], A] {
```
xuwei-k: https://travis-ci.org/scalaz/scalaz/jobs/313239337#L811

[checkGenTypeClasses](https://github.com/scalaz/scalaz/blob/68ee32f5bbc25126935d98e62cbfb81144fe7f5b/project/build.scala#L180) task failed. please separate to a new file 🙏

core/src/main/scala/scalaz/BindRec.scala
```diff
@@ -47,9 +49,40 @@ trait BindRec[F[_]] extends Bind[F] { self =>
...
+  def apply[S[_]: BindRec: Applicative](lift: M <~> S): S[A]
+}
+
 object BindRec {
   @inline def apply[F[_]](implicit F: BindRec[F]): BindRec[F] = F
 
+  def safely[M[_], A](f: MonadRecSafely[M, A])(implicit M: BindRec[M], MA: Applicative[M]): M[A] = {
```
TomasMikula: Wouldn't (some variation on) this be better placed on `Free`.

core/src/main/scala/scalaz/BindRec.scala
puffnfresh: It's intentional. It is representing an unsafe operation, but only ones which can be transformed into safe ones.

core/src/main/scala/scalaz/Foldable.scala
```diff
@@ -1,5 +1,7 @@
 package scalaz
 
+import Isomorphism.<~>
+
```
xuwei-k: https://travis-ci.org/scalaz/scalaz/jobs/313424793#L790

🙏 

```diff
@@ -1,8 +1,8 @@
 package scalaz
 
+////
 import Isomorphism.<~>
 
-////
 /**
  * A type parameter implying the ability to extract zero or more
```

TomasMikula:
I'm having a hard time imagining what we can do safely with

```scala
trait MonadRecSafely[M[_], A] {
  def apply[S[_]: BindRec: Applicative](lift: M <~> S): S[A]
}
```

that we couldn't do safely with the simpler

```scala
// notice that unlike MonadRecSafely, Monadic doesn't commit early to a specific target monad
trait Monadic[A] {
  def apply[M[_]: Monad]: M[A]
}
```

where we adjust the signature of `safely` to be

```scala
def safely[M[_], A](f: Monadic[A])(implicit M: BindRec[M], MA: Applicative[M]): M[A]
```

If there is such use case, let's add it to tests or examples.

---

Tests that `traverseRec_` is short-circuiting on an infinite stream (and/or a long list) would be nice.

puffnfresh:
@TomasMikula it's annoying, because Scala reasons, but I still prefer the code reuse.

TomasMikula:
Scala reasons aside, instead of `Free.liftF` we would write (still manually) `lift.to` and instead of `runRecM` we would write `safely`. We don't save anything and I'm not convinced (yet) that `MonadRecSafely` is a useful abstraction. Subjectively, I don't think that hiding the use of `Free` serves to clarify things.

core/src/main/scala/scalaz/MonadRecSafely.scala
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+  * Takes functions operating on Bind and converts them into a BindRec version.
+  *
+  * Port of Phil Freeman's work in PureScript:
+  *
+  * https://github.com/paf31/purescript-safely/blob/ef7377d7f7659250f2f4a5da02c5d229e4eeed53/src/Control/Safely.purs#L19
+  */
+trait MonadRecSafely[M[_], A] {
+  def apply[S[_]: BindRec: Applicative](lift: M <~> S): S[A]
```
TomasMikula: Seems like one-directional `lift: M ~> S` would be sufficient.

puffnfresh:
> Subjectively, I don't think that hiding the use of Free serves to clarify things.

It does, because parametricity.

core/src/main/scala/scalaz/MonadRecSafely.scala
emilypi: I agree with @TomasMikula - we need at least a `M ~> S` and `S ~> M` here. I agree this doesn't necessarily imply isomorphism, which is too strong of a condition I think. Just an adjunction will do.

PR #1566, opened 2017-12-09
Monad Plus and Traverse instances for Dequeue

Jacoby6000:
There might be a reason these didn't exist before, but I didn't see any old issues about it.  If this gets approved, I'll backport to 7.2.x.

### DIFF
--- a/core/src/main/scala/scalaz/Dequeue.scala
+++ b/core/src/main/scala/scalaz/Dequeue.scala
@@ -237,8 +237,8 @@ sealed abstract class DequeueInstances {
     def append(a: Dequeue[A], b: => Dequeue[A]): Dequeue[A] = a ++ b
   }
 
-  implicit val dequeueInstances: Foldable[Dequeue] with IsEmpty[Dequeue] with MonadPlus[Dequeue] =
-    new Foldable[Dequeue] with IsEmpty[Dequeue] with MonadPlus[Dequeue] {
+  implicit val dequeueInstances: Traverse[Dequeue] with IsEmpty[Dequeue] with MonadPlus[Dequeue] =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/Dequeue.scala
```diff
@@ -237,16 +237,22 @@ sealed abstract class DequeueInstances {
...
+      override def foldLeft[A,B](fa: Dequeue[A], b: B)(f: (B,A)=>B): B = fa.foldLeft(b)(f)
+      override def foldMap[A,B](fa: Dequeue[A])(f: A => B)(implicit F: Monoid[B]): B = fa.foldLeft(F.zero)((b,a) => F.append(b, f(a)))
+      override def empty[A]: Dequeue[A] = Dequeue.empty
+      override def plus[A](a: Dequeue[A], b: => Dequeue[A]): Dequeue[A] = a ++ b
+      override def isEmpty[A](fa: Dequeue[A]) = fa.isEmpty
+      override def length[A](fa: Dequeue[A]) = fa.size
+      override def map[A,B](fa: Dequeue[A])(f: A => B): Dequeue[B] = fa map f
+      override def point[A](a: => A): Dequeue[A] = Dequeue(a)
+      override def bind[A, B](fa: Dequeue[A])(f: A => Dequeue[B]): Dequeue[B] = foldMap(fa)(f)
+      override def traverseImpl[F[_], A, B](fa: Dequeue[A])(f: A => F[B])(implicit F: scalaz.Applicative[F]): F[Dequeue[B]] =
```
TomasMikula: Why the `scalaz.` prefix in `scalaz.Applicative`?

Jacoby6000: Ah, IDE vomit.

PR #1570, opened 2017-12-10
Remove implicit conversion from ST to IO

puffnfresh:
And make IORef.ioRef public

### DIFF
--- a/effect/src/main/scala/scalaz/effect/ST.scala
+++ b/effect/src/main/scala/scalaz/effect/ST.scala
@@ -134,8 +134,7 @@ object ST extends STInstances {
     private[effect] def run = f()
   }
 
-  // Implicit conversions between IO and ST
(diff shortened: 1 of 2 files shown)


effect/src/main/scala/scalaz/effect/ST.scala
```diff
@@ -134,8 +134,7 @@ object ST extends STInstances {
     private[effect] def run = f()
   }
 
-  // Implicit conversions between IO and ST
-  implicit def STToIO[A](st: ST[IvoryTower, A]): IO[A] =
+  def STToIO[A](st: ST[IvoryTower, A]): IO[A] =
```
jedws: not specifically related to this PR, but would this be better defined as a `NaturalTransformation[ST[IvoryTower, _], IO]`?

puffnfresh: No preference

PR #1571, opened 2017-12-10
Remove implicit from natToFunction

puffnfresh:


fommil:
I'm pretty sure scalafix could be set up to find these things.

We found a particularly nasty `String => Cord => String` bouncy in 7.2 recently that I think has been forward ported already.

PR #1573, opened 2017-12-14
sbt 1.0

xuwei-k:


Jacoby6000:
The 2.13 cross builds are a nice touch!

PR #1583, opened 2017-12-17
optimisation in ISet#delete

xuwei-k:


fommil:
IMap could benefit from the same optimisations.

PR #1585, opened 2017-12-17
use "sealed abstract case class" for empty objects

xuwei-k:


core/src/main/scala/scalaz/IList.scala
```diff
@@ -475,9 +475,10 @@ sealed abstract class IList[A] extends Product with Serializable {
 
 }
 
-final case object INil extends IList[Nothing] {
-  def apply[A](): IList[A] = this.asInstanceOf[IList[A]]
-  def unapply[A](e: IList[A]): Boolean = this eq e
+sealed abstract case class INil[A] private() extends IList[A]
```
fommil: argh, this means we can't use a wart to detect uses of non-`final` case class. This will incur additional "allow the wart here" markers.

fommil:
I marginally prefer the `case object` encoding, because it is less hacky and follows the convention of enforcing `final case class`.

However, given the loss of exhaustivity checking, I'm ok with this as an alternative encoding.

fommil:
upstream https://github.com/scala/bug/issues/10660

if it is ever fixed (lol) then we can revert to the slightly cleaner (note I don't say "clean") encoding in a binary breaking change.

core/src/main/scala/scalaz/IList.scala
S11001001: That wart should probably change then.

fommil:
@xuwei-k you missed Map :stuck_out_tongue:

xuwei-k:
>  you missed Map 😛

https://github.com/scalaz/scalaz/pull/1588

Issue #1590, opened 2017-12-20
stringops parse methods should not encourage exceptions


fommil:
the return type of the parse methods in the string ops/syntax uses IllegalArgumentException and NumberFormat etc, which is leakage of bad jvm details. It would be better if we used String as the error type and included more information about the failure, e.g. I've been using this in some decoder code

```
  @inline private def str[A](f: String => A)(implicit A: ClassTag[A]): XDecoder[A] =
    string.andThen { s =>
      try Right(f(s))
      catch {
        case NonFatal(t) =>
          Left(s"expected $A in '$txt', failed: ${t.getMessage}")
      }
    }
```

NeQuissimus:
I've used this from time to time: https://gist.github.com/NeQuissimus/d778e7b621dd2a23cc77d869594bdb64

Is that what you mean?

fommil:
@NeQuissimus yes! Much better than what we have. Although two comments about your gist

- I'm not a fan of type aliases to `String`, it'd rather just see `String` in the sigs
- Your error messages do not include the original string. It's usually nice to see something like `"'Falze' is neither 'true' nor 'false'"`

ritschwumm:
string as an error type sounds a bit lacking - sometime you want to know precisely _why_ parsing e.g. a number went wrong.

ritschwumm:
string as an error type sounds a bit lacking - sometime you want to know precisely _why_ parsing e.g. a number went wrong. or at least at which position in the input string.

fommil:
@ritschwumm but that's exactly what the proposed string message gives you. The java exceptions are useless in this regard. We'll recover what we can from the exception message but wrap it in a contextual message.

jbgi:
> I'm not a fan of type aliases to String, it'd rather just see String in the sigs

I'd rather we use a newtype, it's free after all (at least from 7.3 onward).

ritschwumm:
@fommil for me as a human, it does. but i might prefer dealing with the errors in code. maybe some ADT might make more sense here?

fommil:
@ritschwumm an ADT may be too much, but still better than an exception. How about a case class that has the original string, the exception error, and an optional position? (if there is any way to extract that from the exception).

fommil:
[...]

PR #1592, opened 2017-12-21
Add safe parsing without exceptions

NeQuissimus:
PR for #1590

/cc @fommil

### DIFF
--- a/core/src/main/scala/scalaz/std/String.scala
+++ b/core/src/main/scala/scalaz/std/String.scala
@@ -1,6 +1,11 @@
 package scalaz
 package std
 
+import _root_.java.math.{ BigDecimal, BigInteger }
+
+import scala.reflect.ClassTag
+import scala.util.control.NonFatal
+
 trait StringInstances {
   implicit object stringInstance extends Monoid[String] with Show[String] with Equal[String] with Order[String] with IsEmpty[λ[α => String]] {
     type SA[α] = String
@@ -18,6 +23,8 @@ trait StringInstances {
 }
 
 trait StringFunctions {
+  private[this] implicit val instance = new StringInstances {}.stringInstance
+
   /**
    * Returns the same String value if the given value is 1 otherwise pluralises this String by appending an "s" unless
    * this String ends with "y" and not one of ["ay", "ey", "iy", "oy", "uy"] in which case the 'y' character is chopped and "ies"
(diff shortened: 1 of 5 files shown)


fommil:
I think something even simpler would be better... just throwing away the error message and returning `Maybe[A]`. And replacing the existing syntax. For simple primitives I don't think we really need to see a detailed breakdown about why parsing failed and it can be added retrospectively.

NeQuissimus:
There should be a `toOption` and `toMaybe` on `Either`.
I would prefer keeping the error messages. I have used it to decide between truly bad input or just a required numeric widening (Int -> Long for example).

fommil:
perhaps these can more easily replace the existing syntax in that case, which is currently using `Validation[Throwable, A]`. This is an improvement over that.

NeQuissimus:
Oh, are we able to change the API?

fommil:
@NeQuissimus in 7.3, yup :smile:

core/src/main/scala/scalaz/std/String.scala
```diff
@@ -38,33 +38,32 @@ trait StringFunctions {
...
+  private def intChars(s: String): Boolean = s.forall(c => c.isDigit || c == '-')
+  private def floatChars(s: String): Boolean = s.forall(c => c.isDigit || c == '-' || c == '.')
 
-  def parseBoolean(s:String): Validation[IllegalArgumentException, Boolean] =
-    Validation.fromTryCatchThrowable[Boolean, IllegalArgumentException](s.toBoolean)
+  private def asNumber[T](f: String => T, b: String => Boolean, r: Tuple2[T, T], s: String, t: String): Validation[String, T] =
+    try {
+      Success(f(s))
+    } catch {
+      case _: NumberFormatException if intChars(s) => Failure(s"${s} is outside of range for ${t} (${r._1} - ${r._2})")
```
fommil: won't this be true if `s` is empty?

core/src/main/scala/scalaz/std/String.scala
```diff
@@ -38,33 +38,32 @@ trait StringFunctions {
   def charsNel(s:String, e: => NonEmptyList[Char]) : NonEmptyList[Char] = charsNel(s) getOrElse e
 
   // Parsing functions.
+  private def intChars(s: String): Boolean = s.forall(c => c.isDigit || c == '-')
+  private def floatChars(s: String): Boolean = s.forall(c => c.isDigit || c == '-' || c == '.')
 
-  def parseBoolean(s:String): Validation[IllegalArgumentException, Boolean] =
-    Validation.fromTryCatchThrowable[Boolean, IllegalArgumentException](s.toBoolean)
+  private def asNumber[T](f: String => T, b: String => Boolean, r: Tuple2[T, T], s: String, t: String): Validation[String, T] =
```
fommil: you can use `ClassTag` instead of `t: String` and get the primitive name expanded for free.

core/src/main/scala/scalaz/syntax/std/StringOps.scala
```diff
@@ -24,23 +24,23 @@ final class StringOps(private val self: String) extends AnyVal {
...
+  def parseFloat: Validation[String, Float] = s.parseFloat(self)
 
-  def parseDouble: Validation[NumberFormatException, Double] = s.parseDouble(self)
+  def parseDouble: Validation[String, Double] = s.parseDouble(self)
 
-  def parseBigInt: Validation[NumberFormatException, BigInt] = s.parseBigInt(self)
+  def parseBigInt: Validation[String, BigInt] = s.parseBigInt(self)
 
-  def parseBigDecimal: Validation[NumberFormatException, BigDecimal] = s.parseBigDecimal(self)
+  def parseBigDecimal: Validation[String, BigDecimal] = s.parseBigDecimal(self)
```
fommil: I'm still in favour or removing the `scala.BigDecimal` constructor and replacing with the java variant (noting the scala bug in comments).

fommil:
you forgot to delete the syntax for the "bad" parsers

```
[error] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/syntax/std/StringOps.scala:41:51: value parseBigInt is not a member of object scalaz.std.string
[error]   def parseBigInt: Validation[String, BigInt] = s.parseBigInt(self)
[error]                                                   ^
[error] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/syntax/std/StringOps.scala:43:59: value parseBigDecimal is not a member of object scalaz.std.string
[error]   def parseBigDecimal: Validation[String, BigDecimal] = s.parseBigDecimal(self)
[error]                                                           ^
```

Adding some tests would be :100:

core/src/main/scala/scalaz/std/String.scala
```diff
@@ -38,33 +38,32 @@ trait StringFunctions {
...
+    try {
+      Success(f(s))
+    } catch {
+      case _: NumberFormatException if intChars(s) => Failure(s"${s} is outside of range for ${t} (${r._1} - ${r._2})")
+      case _: NumberFormatException => Failure(s"${s} does not represent a valid ${t}")
+      case e: Throwable => Failure(e.getMessage)
```
fommil: apparently `NonFatal` is better... (personally I think if you're making decisions based on whether you catch or don't catch OOM, you're fucked anyway)

core/src/main/scala/scalaz/std/String.scala
NeQuissimus: Ah, yes, I was trying `TypeTag` and that wasn't making me happy...

fommil: Generally there is need for a new `Describable` typeclass with a macro that generates the refined type name. But `ClassTag` is the closest thing we have from stdlib.

NeQuissimus:
- Switched to the `java.math` variants
- Added tests, which required me to add a `Show[BigDecimal]` and `Equal[BigDecimal]` for `java.math.BigDecimal` as well as simplistic `Arbitrary[BigDecimal]` and `Arbitrary[BigInteger]`

Issue #1513, opened 2017-11-13
FromFoldable and FromFoldable1


fommil:
Maybe I'm not fully grokking `Unfoldable` (which should probably have a ticket) but I found myself needing these trivial little things

```scala
trait FromFoldable[F[_]] {
  def fromFoldable[G[_]: Foldable, A](g: G[A]): F[A]
}
trait FromFoldable1[F[_]] {
  def fromFoldable1[G[_]: Foldable1, A](g: G[A]): F[A]
}
```

with mostly trivial implementations that all our data types could have and could have a `CanBuildFrom` implicit for legacy reasons. This allows the removal of a lot of methods across the board that are effectively `Foldable.toAHardCodedDataType`

I guess what we lose is the ability for `Foldable[IList]` to have an optimal `toIList` method, and such, which is maybe an argument to retain some hardcoded methods on Foldable.

Possibly related to #1452 and #1448

jbgi:
@fommil did you have a look at  `Reducer` (and maybe `Generator`)? I think they could help.

In particular I think we could add the following to `Foldable`:
```scala
  def reduceTo[A, M](fa: F[A])(implicit M: Reducer[A, M]): M = ???
  def to[A, F[_]](fa: F[A])(implicit M: Reducer[A, F[A]]): F[A] = reduceTo[F[A]]
```
But maybe there is already a better way that I did not see.

Jacoby6000:
@jbgi interestingly, it looks like `ReducerOps` has those exact methods hidden inside a sealed abstract class, `UnfoldTo`.
https://github.com/scalaz/scalaz/blob/1a01e0017a9bbc1d07b13a3825e2c7367d0bf624/core/src/main/scala/scalaz/syntax/ReducerOps.scala#L47

fommil:
I love deleting hacky code that I wrote :smile: I'll take a look and report back tonight.

We should definitely get these added to `Foldable`. For 7.3 I'm all in favour of removing the stdlib collection methods on Foldable and requiring people to "reduce" to a `LiterallyCantEvenBuildFrom`

fommil:
I think I need a `Reducer1` (for non empty things)

jbgi:
@fommil  yes, could be worth solving https://github.com/scalaz/scalaz/issues/1038 for 7.3

fommil:
https://hackage.haskell.org/package/reducers-3.12.2/docs/Data-Semigroup-Reducer.html

fommil:
[...]

PR #1602, opened 2017-12-27
 Migrate Reducer in sync with reducer haskell lib

jbgi:
(https://hackage.haskell.org/package/reducers)
Essentially lower the constraint on Reducer from Monoid to Semigroup.
Previous API was based on the now deprecated `monoids` package.

**Question 1**: should `Reducer` `snoc`/`cons` be lazy on their second argument, to match `semigroup` laziness behavior?
**Question 2**: should `Generator` be a type class, as in haskell? (ie. should we add implicit instances?)

**Question 3**: should we remove the unnecessary `Monoid` constraint on `mapTo`? IUC, this would mean that `mapTo` become the only primitive operation of `Generator` and that `mapFrom` is removed entirely.

this is  preliminary work to have a clean solution to  #1513 (`FromFoldable1`)
  
Above questions can be discussed here, but are probably better addressed (if needed) on separate PRs.

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Future.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Future.scala
@@ -302,9 +302,9 @@ object Future {
 
     // implementation runs all threads, dumping to a shared queue
     // last thread to finish invokes the callback with the results
-    override def reduceUnordered[A, M](fs: Seq[Future[A]])(implicit R: Reducer[A, M]): Future[M] =
+    override def reduceUnordered[A, M](fs: Seq[Future[A]])(implicit R: Reducer[A, M], M: Monoid[M]): Future[M] =
       fs match {
-      case Seq() => Future.now(R.zero)
+      case Seq() => Future.now(M.zero)
       case Seq(f) => f.map(R.unit)
       case other => Async { cb =>
         val results = new ConcurrentLinkedQueue[M]
@@ -321,7 +321,7 @@ object Future {
 
             // only last completed f will hit the 0 here.
             if (c.decrementAndGet() == 0)
-              cb(results.asScala.foldLeft(R.zero)((a, b) => R.append(a, b)))
+              cb(results.asScala.foldLeft(M.zero)((a, b) => R.append(a, b)))
             else Trampoline.done(())
           }
         }
@@ -397,6 +397,6 @@ object Future {
   def gatherUnordered[A](fs: Seq[Future[A]]): Future[IList[A]] =
     futureInstance.gatherUnordered(fs)
 
-  def reduceUnordered[A, M](fs: Seq[Future[A]])(implicit R: Reducer[A, M]): Future[M] =
+  def reduceUnordered[A, M: Reducer[A, ?]: Monoid](fs: Seq[Future[A]]): Future[M] =
     futureInstance.reduceUnordered(fs)
 }
--- a/concurrent/src/main/scala/scalaz/concurrent/Task.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Task.scala
@@ -247,15 +247,15 @@ object Task {
           a.map((_, residuals.map(new Task(_))))
         })
       val AE = Apply[Throwable \/ ?]
-      override def reduceUnordered[A, M](fs: Seq[Task[A]])(implicit R: Reducer[A, M]): Task[M] = {
-        import R.monoid
-        val RR: Reducer[Throwable \/ A, Throwable \/ M] =
+      override def reduceUnordered[A, M](fs: Seq[Task[A]])(implicit R: Reducer[A, M], M: Monoid[M]): Task[M] = {
+        implicit val MM: Monoid[Throwable \/ M] = Monoid.liftMonoid[Throwable \/ ?, M]
+        implicit val RR: Reducer[Throwable \/ A, Throwable \/ M] =
           Reducer[Throwable \/ A, Throwable \/ M](
             _.map(R.unit),
-            AE.apply2(_, _)(R.cons(_, _)),
-            AE.apply2(_, _)(R.snoc(_, _))
-          )(Monoid.liftMonoid[Throwable \/ ?, M])
-        new Task(F.reduceUnordered(fs.map(_.get))(RR))
+            AE.apply2(_, _)(R.cons),
+            AE.apply2(_, _)(R.snoc)
+          )
+        new Task(F.reduceUnordered[Throwable \/ A, Throwable \/ M](fs.map(_.get)))
       }
       def fail[A](e: Throwable): Task[A] = new Task(Future.now(-\/(e)))
       def attempt[A](a: Task[A]): Task[Throwable \/ A] = a.attempt
@@ -346,13 +346,13 @@ object Task {
     else
       reduceUnordered[A, IList[A]](tasks, exceptionCancels)
 
-  def reduceUnordered[A, M](tasks: Seq[Task[A]], exceptionCancels: Boolean = false)(implicit R: Reducer[A, M]): Task[M] =
+  def reduceUnordered[A, M](tasks: Seq[Task[A]], exceptionCancels: Boolean = false)(implicit R: Reducer[A, M], M: Monoid[M]): Task[M] =
     if (!exceptionCancels) taskInstance.reduceUnordered(tasks)
     else tasks match {
       // Unfortunately we cannot reuse the future's combinator
       // due to early terminating requirement on task
       // when task fails.  This also makes implementation a bit trickier
-      case Seq() => Task.now(R.zero)
+      case Seq() => Task.now(M.zero)
       case Seq(t) => t.map(R.unit)
       case _ => new Task(Future.Async { cb =>
         val interrupt = new AtomicBoolean(false)
@@ -371,7 +371,7 @@ object Task {
 
               // only last completed f will hit the 0 here.
               if (togo.decrementAndGet() == 0)
-                cb(\/-(results.asScala.foldLeft(R.zero)((a, b) => R.append(a, b))))
+                cb(\/-(results.asScala.foldLeft(M.zero)((a, b) => R.append(a, b))))
               else
                 Trampoline.done(())
             case e@(-\/(failure)) =>
--- a/core/src/main/scala/scalaz/FingerTree.scala
+++ b/core/src/main/scala/scalaz/FingerTree.scala
@@ -3,6 +3,8 @@ package scalaz
 import collection.Iterator
 import syntax.Ops
 import std.option._
+import std.list.listMonoid
+import std.stream.streamMonoid
 
 import FingerTree._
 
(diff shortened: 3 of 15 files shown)


core/src/main/scala/scalaz/Generator.scala
```diff
@@ -1,29 +1,32 @@
...
-  def reduce[E, M](r: Reducer[E, M], c: C[E]): M = to(r, r.zero, c)
+trait Generator[C, Elem] {
 
-  def to[E, M](r: Reducer[E, M], m: M, c: C[E]): M = r.append(m, reduce(r, c))
+  def reduce[E, M](e: Elem => E, c: C)(implicit r: Reducer[E, M], m: Monoid[M]): M = to(e, m.zero, c)
 
-  def from[E, M](r: Reducer[E, M], c: C[E], m: M): M = r.append(reduce(r, c), m)
+  def to[E, M](e: Elem => E, m: M, c: C)(implicit r: Reducer[E, M], M: Monoid[M]): M = r.append(m, reduce(e, c))
+
+  def from[E, M](e: Elem => E, c: C, m: M)(implicit r: Reducer[E, M], M: Monoid[M]): M = r.append(reduce(e, c), m)
```
TomasMikula: I know these are not introduced in this PR but what's the purpose of `to` and `from`? Why are they two distinct methods? The signature is the same, modulo parameter order, and the default implementation suspicious (again pre-existent, not introduced in this PR). Also, since `m: M` is taken as an argument, it seems strange to also require `Monoid[M]`.

core/src/main/scala/scalaz/Generator.scala
```diff
@@ -1,29 +1,32 @@
 package scalaz
 
 /**
- * A `Generator[C]` is a container of elements, and which knows how to efficiently apply a [[scalaz.Reducer]]
+ * A `Generator[C, Elem]` is a container of elements, and which knows how to efficiently apply a [[scalaz.Reducer]]
```
TomasMikula: Seems like `C` is the container (not `Generator[C, Elem]`)

core/src/main/scala/scalaz/Generator.scala
jbgi: Good point question. I'm guessing this is to allows optimal implementation for some instance where `append` is costly. The only instance I found that actually overwrite this method is for `Free`: https://github.com/ekmett/ekmett.github.com/blob/master/haskell/monoids/Data/Generator/Free.hs#L109

The unnecessary `Monoid` constraint on `to` is so that it can be implemented in term of `reduce` for the case where the later method is used as minimal definition.

We could drop the constraint and have `to` as the only minimal definition. (but this would also likely means we also drop the `from` method altogether, unless we keep the constraint but that would be weird...)

@ekmett, do you have any recommendation on the matter? (is `Generator.mapFrom` useful in practice?)

TomasMikula: Regarding

> which knows how to efficiently apply a [[scalaz.Reducer]]
> to extract an answer by combining elements. A `Reducer` may supply efficient left-to-right and
> right-to-left reduction strategies that a `Generator` may avail itself of.

the `Generator` only knows how to efficiently traverse a container. The `Reducer` may have an efficient reduction in one direction, but not the other. The `Generator` would choose the optimal strategy for the container, but that might have pathological performance for the reducer. It might therefore be better to have one-directional reducers.

core/src/main/scala/scalaz/Generator.scala
jbgi: The main purpose of the Generator/Reducer classes is to be a lightweight general purpose map-reduce framework. So the Generator must have the ability to reduce in parallel, aggregating from both sides. Caller is responsible to chose Generator/reducer pair that work well with each other.

core/src/main/scala/scalaz/Generator.scala
jbgi: The main purpose of the Generator/Reducer classes is to be a lightweight general purpose map-reduce framework. So the Generator must have the ability to reduce in parallel, aggregating from both sides. Caller is responsible to chose Generator/reducer pair that work well with each other.
(edit: moved comment)

TomasMikula: If the caller is responsible for matching of reducers to generators, then shouldn't both be explicit (as opposed to implicit)?

TomasMikula:
Are there some efficiency requirements on `Reducer` instances? For example, is it fine to provide a `Reducer` into `List`, even though `snoc` is inefficient?

TomasMikula:
Maybe unify the naming `cons`/`snoc` and `mapTo`/`mapFrom`, since they seem to be related. For example, rename `mapTo`/`mapFrom` to `snocMap`/`consMap`?

core/src/main/scala/scalaz/Generator.scala
jbgi: Well, currently Generator is not a type class, but Reducer is one, and probably need to stay so, to ensure coherence with Semigroup.

jbgi:
@TomasMikula I would say so, in the same manner that the list semigroup is not efficient if lhs is big list.

jbgi:
> Maybe unify the naming cons / snoc and mapTo / mapFrom , since they seem to be related. 

There's maybe a link as for how the 'zero' element is aggregated (end or beginning) but I fear it would suggest that implementation rely only on one of the two Reducer methods, which is generally false.

concurrent/src/main/scala/scalaz/concurrent/Future.scala
```diff
@@ -302,9 +302,9 @@ object Future {
 
     // implementation runs all threads, dumping to a shared queue
     // last thread to finish invokes the callback with the results
-    override def reduceUnordered[A, M](fs: Seq[Future[A]])(implicit R: Reducer[A, M]): Future[M] =
+    override def reduceUnordered[A, M](fs: Seq[Future[A]])(implicit R: Reducer[A, M], M: Monoid[M]): Future[M] =
       fs match {
-      case Seq() => Future.now(R.zero)
+      case Seq() => Future.now(M.zero)
```
tonymorris: I think not using `Seq` is a good idea. Consider either a concrete type such as `List`, or use a type-variable with `Foldable`.

concurrent/src/main/scala/scalaz/concurrent/Task.scala
```diff
@@ -247,15 +247,15 @@ object Task {
           a.map((_, residuals.map(new Task(_))))
         })
       val AE = Apply[Throwable \/ ?]
-      override def reduceUnordered[A, M](fs: Seq[Task[A]])(implicit R: Reducer[A, M]): Task[M] = {
-        import R.monoid
-        val RR: Reducer[Throwable \/ A, Throwable \/ M] =
+      override def reduceUnordered[A, M](fs: Seq[Task[A]])(implicit R: Reducer[A, M], M: Monoid[M]): Task[M] = {
+        implicit val MM: Monoid[Throwable \/ M] = Monoid.liftMonoid[Throwable \/ ?, M]
```
tonymorris: Inserting a `Monoid` in the middle of code seems wonky. A `Monoid` is a type-class right?

core/src/main/scala/scalaz/Reducer.scala
```diff
@@ -29,21 +29,18 @@ sealed abstract class Reducer[C, M] {
...
-    implicit val m = Reducer.this.monoid
-    implicit val n = r.monoid
+    implicit val m = Reducer.this.semigroup
+    implicit val n = r.semigroup
     new Reducer[C, (M, N)] {
 
       import std.tuple._
 
-      val monoid = Monoid[(M, N)]
+      val semigroup = Semigroup[(M, N)]
```
tonymorris: ?

concurrent/src/main/scala/scalaz/concurrent/Task.scala
jbgi: yes, same thing for the reducer above.

core/src/main/scala/scalaz/Reducer.scala
jbgi: what is the issue here?

core/src/main/scala/scalaz/FingerTree.scala
```diff
@@ -784,11 +779,11 @@ object FingerTree extends FingerTreeInstances {
 
     def measure: V
 
-    def toList: List[A] = map(x => x)(Reducer.ListReducer[A]).measure
+    def toList: List[A] = map[A, List[A]](x => x).measure
```
edmundnoble: 👍

core/src/main/scala/scalaz/Generator.scala
```diff
@@ -1,29 +1,32 @@
...
+ * `Generator` a class of container of elements that knows how to efficiently apply a [[scalaz.Reducer]]
  * to extract an answer by combining elements. A `Reducer` may supply efficient left-to-right and
  * right-to-left reduction strategies that a `Generator` may avail itself of.
  */
-trait Generator[C[_]] {
-  def reduce[E, M](r: Reducer[E, M], c: C[E]): M = to(r, r.zero, c)
+trait Generator[C, Elem] {
 
-  def to[E, M](r: Reducer[E, M], m: M, c: C[E]): M = r.append(m, reduce(r, c))
+  def mapReduce[E, M](e: Elem => E, c: C)(implicit r: Reducer[E, M], m: Monoid[M]): M = mapTo(e, m.zero, c)
```
edmundnoble: 👍

core/src/main/scala/scalaz/Reducer.scala
edmundnoble: ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

core/src/main/scala/scalaz/Reducer.scala
```diff
@@ -109,6 +106,11 @@ sealed abstract class ReducerInstances {
     unitConsReducer(_ :: INil(), _ :: _)
   }
 
+  /** Collect `C`s into an NonEmptyList, in order. */
+  implicit def NonEmptyListReducer[C]: Reducer[C, NonEmptyList[C]] = {
+    unitConsReducer(NonEmptyList.nel(_, INil()),  _ <:: _)
```
edmundnoble: 👍

concurrent/src/main/scala/scalaz/concurrent/Future.scala
jbgi: Created #1627 to address the issue separately. (as this PR did not introduced any usage of `Seq`)

concurrent/src/main/scala/scalaz/concurrent/Task.scala
jbgi: Created #1628 to address the issue separately and at large (as this PR did not introduced the local instances).

jbgi:
@TomasMikula would you be ok to pursue an eventual redesign of Generator primitives in a separate PR/issue? The current PR is rather conservative and `Generator` and could be merged as a first step.

TomasMikula:
Yes. My inquiries were not meant to hold this PR back.

PR #1603, opened 2017-12-28
Make several data types invariant

### LINKED ISSUE #1469: \/ should not be covariant

TomasMikula:
* `\/`
 * `Either3`
 * `LazyEither`
 * `MonoidCoproduct` (`:+:`)
 * `These` (`\&/`)
 * `Validation`
 * `Future`
 * `Task`

Resolves #1469.

Note that I added coercion methods like

```scala
-\/[A, B](a).coerceRight[C] : (A \/ C)
```

so that `A -\/ B` can be coerced to `A \/ C` at zero allocation cost.

I could remove these methods, at the cost of introducing some new allocations.

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Future.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Future.scala
@@ -51,7 +51,7 @@ import scala.concurrent.duration._
  * `Future[Throwable \/ A]` with a number of additional
  * convenience functions.
  */
-sealed abstract class Future[+A] {
+sealed abstract class Future[A] {
   import Future._
 
   def flatMap[B](f: A => Future[B]): Future[B] = this match {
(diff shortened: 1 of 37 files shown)


jdegoes:
Nice work! I'd be in favor of not having the coercions, or at least adding `Liskov` to make it safe.

TomasMikula:
The coercions are safe. Notice that
 - `coerceRight` is defined only on `-\/`,
 - `coerceLeft` is defined only on `\/-`,

so we are only ever coercing the absent value.

jdegoes:
@TomasMikula Ah, I missed that. Looks great! 👍

fommil:
presumably the `\/.left` / `\/.right` constructors (and their syntax) are now much preferred over the direct `-\/` / `\/-` case class apply?

I really wish we had proper sum types...

TomasMikula:
And even syntax extensions like `a.left[B]`. Though users may also be forced to remove variance from their own data types containing `\/`. I would be surprised if anyone automated that.

concurrent/src/main/scala/scalaz/concurrent/Future.scala
```diff
@@ -51,7 +51,7 @@ import scala.concurrent.duration._
  * `Future[Throwable \/ A]` with a number of additional
  * convenience functions.
  */
-sealed abstract class Future[+A] {
+sealed abstract class Future[A] {
```
emilypi: 💯

PR #1612, opened 2018-01-05
Add Tannen

puffnfresh:
Composes a Functor on the outside of a Bifunctor.

### DIFF
--- a/core/src/main/scala/scalaz/Tannen.scala
+++ b/core/src/main/scala/scalaz/Tannen.scala
@@ -8,16 +8,46 @@ final case class Tannen[F[_], G[_, _], A, B](f: F[G[A, B]])
 
 object Tannen extends TannenInstances
 
-sealed abstract class TannenInstances {
+sealed abstract class TannenInstances0 {
+  implicit def TannenFoldable[F[_], G[_, _], E](implicit F: Foldable[F], G: Bifoldable[G]): Foldable[Tannen[F, G, E, ?]] =
+    new Foldable[Tannen[F, G, E, ?]] {
+      def foldMap[A, M](fa: Tannen[F, G, E, A])(f: A => M)(implicit M: Monoid[M]): M =
+        F.foldMap(fa.f)(G.bifoldMap(_)(_ => M.zero)(f))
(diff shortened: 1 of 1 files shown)


core/src/main/scala/scalaz/Tannen.scala
```diff
@@ -0,0 +1,23 @@
...
+/** Composes a Functor on the outside of a Bifunctor.
+  *
+  * [[https://hackage.haskell.org/package/bifunctors-5.5/docs/Data-Bifunctor-Tannen.html#t:Tannen]]
+  */
+final case class Tannen[F[_], G[_, _], A, B](f: F[G[A, B]])
+
+object Tannen extends TannenInstances
+
+sealed abstract class TannenInstances {
+  implicit def TannenFunctor[F[_], G[_, _], E](implicit F: Functor[F], G: Functor[G[E, ?]]): Functor[Tannen[F, G, E, ?]] =
```
puffnfresh: For some reason, bifunctors has this as:

```haskell
(Functor f, Bifunctor p) => Functor (Tannen f p a)
```

I've relaxed Bifunctor to Functor. This adds a flexible context, but I don't think that'd be a problem. Is there a reason why bifunctors does what it does?

puffnfresh: I changed to what bifunctors does because otherwise the instance is the same as:

```haskell
(Functor f, Functor g) => Functor (Compose f g)
```

TomasMikula:
I'm curious, what's the origin of the name `Tannen`? (/cc @ekmett)

jbgi:
> I'm curious, what's the origin of the name Tannen? (/cc @ekmett)

$10 on Biff(unctor) Tannen.

puffnfresh:
Haha wow that's right! https://github.com/ekmett/bifunctors/issues/10#issuecomment-46172009

ekmett:
Indeed. Biff was "BIfunctor Functor Functor" and Tannen was added later because I needed a Functor-wrapped Bifunctor and the name didnt roll off the tongue.

In retrospect the structure for Tannen is the same as the structure of a generalized "Cayley" construction or as the identical "Static" arrow construction, but the variance on the bifunctor/profunctor/arrow is different.

PR #1613, opened 2018-01-05
Datatypes for (Co)Free

emilypi:
Here are a few bits I have been working on. `Free`, `Cofree`, `IStream`, and some work on modeling (co)inductive types (contributed by the extremely helpful @alexknvl) based on [this paper](http://www.cse.chalmers.se/~nad/publications/danielsson-altenkirch-mixing.pdf)

### DIFF
--- a/base/shared/src/main/scala/scalaz/control/Inf.scala
+++ b/base/shared/src/main/scala/scalaz/control/Inf.scala
@@ -1,0 +1,43 @@
+package scalaz
+package control
+
+/**
+  * Modelled from http://www.cse.chalmers.se/~nad/publications/danielsson-altenkirch-mixing.pdf
+  */
+sealed abstract class Bool
+sealed abstract class True extends Bool
+sealed abstract class False extends Bool
+
+object Bool {
+  sealed abstract class Value[B <: Bool]
+  implicit object TrueValue extends Value[True]
+  implicit object FalseValue extends Value[False]
+}
+
+final class Inf[+A] private (thunk: () => A) {
+  lazy val force: A = thunk()
+}
+
+object Inf {
+  def apply[A](a: =>A): Inf[A] = new Inf(() => a)
+}
+
+sealed abstract class MaybeInf[B <: Bool, A] {
+  def force: A
+}
+final case class Inductive[A](value: A) extends MaybeInf[True, A] {
+  def force: A = value
+}
+final case class Coinductive[A](thunk: Inf[A]) extends MaybeInf[False, A] {
+  def force: A = thunk.force
+}
+
+object MaybeInf {
+  def inductive[A](a: A): MaybeInf[True, A] = Inductive(a)
+  def coinductive[A](a: Inf[A]): MaybeInf[False, A] = Coinductive(a)
+
+  def delay[B <: Bool, A](value: =>A)(implicit B: Bool.Value[B]): MaybeInf[B, A] = B match {
+    case Bool.TrueValue => Inductive(value)
+    case Bool.FalseValue => Coinductive(Inf(value))
+  }
+}
--- a/base/shared/src/main/scala/scalaz/data/Cofree.scala
+++ b/base/shared/src/main/scala/scalaz/data/Cofree.scala
@@ -1,0 +1,22 @@
+package scalaz
+package data
+
+import scalaz.control.{Coinductive, Inf}
+
+trait CofreeModule {
+  type Cofree[F[_], A]
+
+  def runCofree[F[_], A](f: Cofree[F, A]): F[data.Cofree[F, A]]
+
+  def wrapCofree[F[_], A](a: =>A)(f: =>F[data.Cofree[F, A]]): Cofree[F, A]
+}
+
+private[data] object CofreeImpl extends CofreeModule {
+  type Cofree[F[_], A] = (Coinductive[A], Coinductive[F[data.Cofree[F, A]]])
+
+  def runCofree[F[_], A](f: Cofree[F, A]): F[data.Cofree[F, A]] =
+    f._2.force
+
+  def wrapCofree[F[_], A](a: =>A)(f: =>F[data.Cofree[F, A]]): Cofree[F, A] =
+    (Coinductive(Inf(a)), Coinductive(Inf(f)))
+}
--- a/base/shared/src/main/scala/scalaz/data/Free.scala
+++ b/base/shared/src/main/scala/scalaz/data/Free.scala
@@ -1,0 +1,18 @@
+package scalaz
+package data
+
+trait FreeModule {
+  type Free[F[_], A]
+
+  def runFree[F[_], A](f: Free[F, A]): A \/ F[data.Free[F, A]]
+
+  def wrapFree[F[_], A](f: A \/ F[data.Free[F, A]]): Free[F, A]
+}
+
+private[data] object FreeImpl extends FreeModule {
+  type Free[F[_], A] = A \/ F[data.Free[F, A]]
+
+  def runFree[F[_], A](f: Free[F, A]): A \/ F[data.Free[F, A]] = f
+
+  def wrapFree[F[_], A](f: A \/ F[data.Free[F, A]]): Free[F, A] = f
+}
--- a/base/shared/src/main/scala/scalaz/data/IStream.scala
+++ b/base/shared/src/main/scala/scalaz/data/IStream.scala
@@ -1,0 +1,19 @@
+package scalaz
+package data
+
+sealed abstract class IStreamModule {
+  type IStream[A]
+
+  def scons[A](a: =>A)(as: =>IStream[A]): IStream[A]
+  def uncons[A](as: =>IStream[A]): (A, IStream[A])
+}
+
+private[data] object IStreamImpl extends IStreamModule { self =>
+  type IStream[A] = Cofree[(A, ?), Unit]
+
+  def scons[A](a: =>A)(as: =>IStream[A]): IStream[A] =
+    Cofree.wrapCofree[(A, ?), Unit](())((a, as))
+
+  def uncons[A](as: =>IStream[A]): (A, IStream[A]) =
+    Cofree.runCofree[(A, ?), Unit](as)
+}
--- a/base/shared/src/main/scala/scalaz/data/IStream.scala
+++ b/base/shared/src/main/scala/scalaz/data/IStream.scala
@@ -1,7 +1,7 @@
 package scalaz
 package data
 
-sealed abstract class IStreamModule {
+sealed trait IStreamModule {
   type IStream[A]
 
   def scons[A](a: =>A)(as: =>IStream[A]): IStream[A]
--- a/base/shared/src/main/scala/scalaz/data/IStream.scala
+++ b/base/shared/src/main/scala/scalaz/data/IStream.scala
@@ -1,7 +1,7 @@
 package scalaz
 package data
 
-sealed trait IStreamModule {
+trait IStreamModule {
   type IStream[A]
 
   def scons[A](a: =>A)(as: =>IStream[A]): IStream[A]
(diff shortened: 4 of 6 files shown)


NeQuissimus:
Could I ask for two things?

- Can you hook the types into the `Prelude`?
- Could you provide some documentation in the `examples` project?

emilypi:
@NeQuissimus Sure thing

base/shared/src/main/scala/scalaz/control/Inf.scala
```diff
@@ -0,0 +1,43 @@
+package scalaz
+package control
+
+/**
+  * Modelled from http://www.cse.chalmers.se/~nad/publications/danielsson-altenkirch-mixing.pdf
+  */
+sealed abstract class Bool
+sealed abstract class True extends Bool
```
TomasMikula: This Scaladoc comment will apply to `Bool`, which you probably did not intend.

base/shared/src/main/scala/scalaz/control/Inf.scala
```diff
@@ -0,0 +1,43 @@
...
+sealed abstract class False extends Bool
+
+object Bool {
+  sealed abstract class Value[B <: Bool]
```
TomasMikula: Should subtyping get in the way (or just get annoying to always mention the bound `<: Bool`), I think everything should work even without this bound (and without the superclass `Bool`).

base/shared/src/main/scala/scalaz/data/Cofree.scala
```diff
@@ -0,0 +1,22 @@
...
+trait CofreeModule {
+  type Cofree[F[_], A]
+
+  def runCofree[F[_], A](f: Cofree[F, A]): F[data.Cofree[F, A]]
+
+  def wrapCofree[F[_], A](a: =>A)(f: =>F[data.Cofree[F, A]]): Cofree[F, A]
+}
+
+private[data] object CofreeImpl extends CofreeModule {
+  type Cofree[F[_], A] = (Coinductive[A], Coinductive[F[data.Cofree[F, A]]])
```
TomasMikula: Some would say that using a type constructor (`Coinductive`) as its own type is bad practice. Don't you really want `Inf` instead of `Coinductive`?

base/shared/src/main/scala/scalaz/data/Free.scala
```diff
@@ -0,0 +1,18 @@
...
+trait FreeModule {
+  type Free[F[_], A]
+
+  def runFree[F[_], A](f: Free[F, A]): A \/ F[data.Free[F, A]]
+
+  def wrapFree[F[_], A](f: A \/ F[data.Free[F, A]]): Free[F, A]
+}
+
+private[data] object FreeImpl extends FreeModule {
+  type Free[F[_], A] = A \/ F[data.Free[F, A]]
```
TomasMikula: This representation of `Free` is known to be non-stack-safe (and inefficient).

base/shared/src/main/scala/scalaz/data/IStream.scala
```diff
@@ -0,0 +1,19 @@
...
+
+trait IStreamModule {
+  type IStream[A]
+
+  def scons[A](a: =>A)(as: =>IStream[A]): IStream[A]
+  def uncons[A](as: =>IStream[A]): (A, IStream[A])
+}
+
+private[data] object IStreamImpl extends IStreamModule { self =>
+  type IStream[A] = Cofree[(A, ?), Unit]
```
TomasMikula: Why not `Cofree[Id, A]`?

base/shared/src/main/scala/scalaz/control/Inf.scala
```diff
@@ -0,0 +1,43 @@
...
+  implicit object TrueValue extends Value[True]
+  implicit object FalseValue extends Value[False]
+}
+
+final class Inf[+A] private (thunk: () => A) {
+  lazy val force: A = thunk()
+}
```
TomasMikula: This type seems redundant. (What does it give us compared to `Function0[A]`?)

base/shared/src/main/scala/scalaz/control/Inf.scala
emilypi: I've fixed this with more documentation in updates for tomorrow, thanks :)

base/shared/src/main/scala/scalaz/data/Cofree.scala
emilypi: In talks with @edmundnoble last night we've decided to implement it it in terms of `Fix` and `EnvT` intead a la 

```scala

trait CofreeModule {
  type Cofree[F[_], A]

  def runCofree[F[_], A](f: Cofree[F, A]): (A, F[Cofree[F, A]])

  def wrapCofree[F[_], A](a: =>A)(f: =>F[Fix[EnvT[A, F, ?]]]): Cofree[F, A]
}

private[data] object CofreeImpl extends CofreeModule {
  final class EnvT[A, T[_], X](val run: (A, T[X]))

  type Cofree[F[_], A] = Fix[EnvT[A, F, ?]]

  def runCofree[F[_], A](f: Cofree[F, A]): (A, F[Cofree[F, A]]) =
    Fix.unfix[EnvT[A, F, ?]](f).run

  def wrapCofree[F[_], A](a: =>A)(f: => F[Cofree[F, A]]): Cofree[F, A] =
    Fix.fix[EnvT[A, F, ?]](new EnvT((a, f)))
}

```

base/shared/src/main/scala/scalaz/data/Free.scala
emilypi: We'll fix this then. Do you recommend a specific paper/project with a more efficient representation?

base/shared/src/main/scala/scalaz/data/IStream.scala
emilypi: We were literally talking about that as you commented! It's `Cofree[Id, A]` now. I prefer it this way as well.

base/shared/src/main/scala/scalaz/data/Free.scala
TomasMikula: I think scalaz 7 is a good starting point, i.e.

```scala
A | F[A] | (Free[F, X], X => Free[F, A])
```
(where the case `F[A]` represents _embedding_ `F[A]` into `Free[F, A]`, and the other two cases represent the _free versions of operations_ `pure` and `flatMap`, respectively).

In addition to being stack-safe, it also avoids requiring `Functor[F]` for monad operations on `Free[F, A]`.

base/shared/src/main/scala/scalaz/data/IStream.scala
```diff
@@ -0,0 +1,22 @@
...
+
+trait IStreamModule {
+  type IStream[A]
+
+  def scons[A](a: =>A)(as: =>Id[IStream[A]]): IStream[A]
+  def uncons[A](as: =>IStream[A]): (A, Inf[IStream[A]])
+}
+
+private[data] object IStreamImpl extends IStreamModule { self =>
+  type IStream[A] = data.Cofree[Id, A]
```
TomasMikula: This representation of `IStream` will always evaluate at least one element. Was that intended?

emilypi: We could wrap it in `Inf`, that'd be an easy change, and probably more accurate

base/shared/src/main/scala/scalaz/data/IStream.scala
```diff
@@ -0,0 +1,22 @@
+package scalaz
+package data
+
+import scalaz.control.Inf
+
+
+trait IStreamModule {
+  type IStream[A]
+
+  def scons[A](a: =>A)(as: =>Id[IStream[A]]): IStream[A]
```
TomasMikula: If `IStream` already encapsulates laziness, then there should be no need to pass it by-name. Also, why wrapped in `Id`?

base/shared/src/main/scala/scalaz/data/IStream.scala
```diff
@@ -0,0 +1,22 @@
...
+  def uncons[A](as: =>IStream[A]): (Inf[A], Inf[IStream[A]])
```
TomasMikula: `uncons` should return an actual `A` (as opposed to a thunk of `A`).

Again, if `Stream` already encapsulates laziness, there should be no need to wrap it in `Inf`.

base/shared/src/main/scala/scalaz/data/IStream.scala
emilypi: True about the by-name, error on the `Id`. I've removed it.

base/shared/src/main/scala/scalaz/data/IStream.scala
```diff
@@ -0,0 +1,19 @@
...
+
+trait IStreamModule {
+  type IStream[A]
+
+  def scons[A](a: A)(as: IStream[A]): IStream[A]
+  def uncons[A](as: IStream[A]): (A, IStream[A])
+}
+
+private[data] object IStreamImpl extends IStreamModule {
+  type IStream[A] = data.Cofree[Id, A]
```
TomasMikula: With the current definitions of `Cofree` and `EnvT`, this looks like constructing an infinite stream eagerly.

TomasMikula:
What does "I" in `IStream` stand for? Is it "infinite" or "invariant"? Was it intentional for every `IStream` to be infinite, as opposed to just potentially infinite?

emilypi:
@TomasMikula 

It's being reworked, but yes, "Invariant" was the idea - the intention is to make each `IStream` infinite. I simply haven't had time to work on it yet.

TomasMikula:
Wi will also need a name for the potentially finite stream (note that `Stream` and scalaz 7 `StreamT` are potentially finite). Even if we chose to model the end of stream as just an effect on the (infinite-) stream transformer, it would be useful to have a name for this common case. So we should plan the names ahead.

base/shared/src/main/scala/scalaz/data/Freer.scala
```diff
@@ -0,0 +1,86 @@
...
+  *
+  */
+sealed abstract class Freer[F[_], A] {
+
+  def map[B](f: A => B): Freer[F, B] =
+    flatMap[B](a => Freer.pure[F, B](f(a)))
+
+  def flatMap[B](k: A => Freer[F, B]): Freer[F, B] =
+    this match {
+      case Freer.Pure(a) => k(a)
```
TomasMikula: This eagerness is tempting, but not stack-safe.

Consider

```scala
Pure(a).flatMap(x => Pure(x).flatMap(x => Pure(x).flatMap(x => Pure(x).flatMap(...))))
```

base/shared/src/main/scala/scalaz/data/Freer.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+object Freer extends FreerInstances {
+  private[Freer] type =>:[F[_], A, B] = Kleisli[Freer[F, ?], A, B]
+  private[Freer] type Quiver[F[_], A, B] = ACatenable1[=>:[F, ?, ?], A, B]
+
+  final case class Pure[F[_], A] private (a: A) extends Freer[F, A]
+  final case class LiftF[F[_], A] private (fa: F[A]) extends Freer[F, A]
+  final case class Impure[F[_], E, A] private (f: Freer[F, E], q: Quiver[F, E, A]) extends Freer[F, A] {
```
TomasMikula: I don't think we can make any use of the extra reflection here (a list of functions is as opaque as a single function), and the extra level of indirection will only make it slower than your other `Free` encoding.

base/shared/src/main/scala/scalaz/data/Freer.scala
emilypi: That's definitely the next pass over this code - making sure everything is stacksafe. I'm noticing the implementation of `foldMap` in both `Free` and this one aren't stacksafe either.

PR #1614, opened 2018-01-07
Add Store monad and State comonad

japgolly:
I read http://comonad.com/reader/2018/the-state-comonad/ and decided to quickly bang this out.

I also tried generising for `{State,Store}T` but they'd both need typeclasses over `F[_]` on both sides of the dual line. Specifically, for a `Monad[StoreT[F, S, ?]]`, `peek` needs a `Comonad[F]` where as `Monad#point` needs `Point[F]`. And similarly, a `Comonad[StateT[F, S, ?]]` needs both `Bind[F]` and `copoint`.


xuwei-k:
Could you add some tests?

japgolly:
Done. Thanks for reminding me, the tests highlighted that I'd made a small mistake in the implementation.

vmarquez:
👍

PR #1620, opened 2018-01-20
Add `with this.type` refinement to subclass members

edmundnoble:
Why remove subtyping from the type class hierarchy? Because it gives us ambiguous implicits.
Why keep subtyping in the type class hierarchy? Because it's efficient.

The moment someone implements `def semigroup: Semigroup[A]` in the `Monoid` class with anything but `= this`, all of Scalaz's methods accessing that method get deoptimized and the call is no longer inlined.

All instances must be *implemented* using subtyping, to keep the guarantee that our code doesn't become ridiculously slow because the JIT can't see through it. But subtyping in the *interface* is an abomination.

This PR enshrines these things in the types.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/ConstInstances.scala
+++ b/base/shared/src/main/scala/scalaz/data/ConstInstances.scala
@@ -3,7 +3,6 @@ package data
 
 import typeclass.{FoldableClass, TraversableClass}
 
-import Prelude._
 import FoldableClass._
 import TraversableClass._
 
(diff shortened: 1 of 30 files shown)


base/shared/src/main/scala/scalaz/data/ConstInstances.scala
```diff
@@ -21,30 +20,54 @@ trait ConstInstances {
     override def toList[A](fa: Const[R, A]): List[A] = Nil
   }
 
-  implicit def constFunctor[R]: Functor[Const[R, ?]] = new Functor[Const[R, ?]] {
+  trait ConstFunctor[R] extends Functor[Const[R, ?]] {
```
hrhino: There remains, unless I've missed it, an outer pointer per inner trait. Not sure how much of a difference that makes.

base/shared/src/main/scala/scalaz/data/ConstInstances.scala
```diff
@@ -21,30 +20,54 @@ trait ConstInstances {
...
     def map[A, B](fa: Const[R, A])(f: A => B): Const[R, B] =
       fa.retag[B]
   }
 
-  implicit def constApply[R](implicit R: Semigroup[R]): Apply[Const[R, ?]] = new Apply[Const[R, ?]] {
-    def functor: Functor[Const[R, ?]] = Const.constFunctor[R]
+  trait ConstApply[R] extends Apply[Const[R, ?]] with ConstFunctor[R] {
```
hrhino: Why not so for `Applicative`?

base/shared/src/main/scala/scalaz/typeclass/ApplyClass.scala
```diff
@@ -2,5 +2,5 @@ package scalaz
 package typeclass
 
 trait ApplyClass[F[_]] extends Apply[F] with FunctorClass[F] {
-  implicit final def apply: Apply[F] = this
+  implicit final def apply: Apply[F] with this.type = this
```
hrhino: I disagree with this being `implicit`, but you didn't do that.

base/shared/src/main/scala/scalaz/typeclass/BindClass.scala
```diff
@@ -2,7 +2,7 @@ package scalaz
 package typeclass
 
 trait BindClass[M[_]] extends Bind[M] with ApplyClass[M] {
-  final def bind: Bind[M] = this
+  final def bind: Bind[M] with this.type = this
```
hrhino: I would love to make this a wart or something checkable.

Someone make and approve a scalaz compiler plugin and it will be so.

jbgi:
Alternatively, this could be a mixin constraint:
```scala
trait Monoid[A] { self: Semigroup[A] =>
  final def semigroup: Semigroup[A] = this
```

base/shared/src/main/scala/scalaz/data/ConstInstances.scala
edmundnoble: I don't see how.

base/shared/src/main/scala/scalaz/data/ConstInstances.scala
edmundnoble: Huh. Good point.

base/shared/src/main/scala/scalaz/typeclass/ApplyClass.scala
edmundnoble: Lol wtf. Fixing that.

edmundnoble:
@jbgi That's a very interesting idea. I'll explore it.

jbgi:
@edmundnoble Found another approach: https://github.com/edmundnoble/scalaz/pull/1
=> should allows to remove the unnecessary methods calls.

edmundnoble:
Closed in favor of #1623. #1623 makes us not even *need* to do these method calls unless we call methods with less specific constraints (e.g., calling a `m[A: Semigroup]` with a `Monoid[A]` in scope).

PR #1621, opened 2018-01-20
Rename Show to Debug

edmundnoble:
I don't want people using `Show` for any business logic. `Debug` is for `debug`ging. Make your own type class if you want it to have meaning.


hrhino:
Move `Show.md` to `Debug.md` and you will have found da wae.

example/src/main/tut/typeclass/Debug.md
```diff
@@ -1,11 +1,11 @@
...
-title:  "Show"
+title:  "Debug"
 ---
 
-# Show
+# Debug
 
-`Show` presents a safe and explicit alternative to `Object.toString`.
+`Debug` presents a safe and explicit alternative to `Object.toString`.
 By providing instances of this typeclass, a type explicitly defines that it can be converted into human-readable output.
```
hrhino: I think you could explicitly qualify that this typeclass isn't to be used for general showing around of arbitrary values.

edmundnoble: I think the "human-readable output" is pretty good.

PR #1623, opened 2018-01-21
Type classes: expose super class functions in sub classes. (WIP)

jbgi:
This is an alternative approach to #1620 achieving the goal of ensuring inheritance-based implementations of type classes, here via seal of type class + Template.

This PR also expose super class functions in sub classes: this avoid going through no-op methods to switch to super classes in many cases.

WIP: Monoid/Semigroup and Functor->Monad only. For early review. Remaining type classes will follow.


jdegoes:
@jbgi Looks good to me! /cc @edmundnoble

TomasMikula:
👍  I like it. It even seems clearer to me than the previous approach.

TomasMikula:
Something for future thought: Maybe we could save some `Template`s if we defined the actual typeclasses to be used as

```scala
type Monad[F[_]] = InstanceOf[Monad.Class[F]]
```

where `InstanceOf` would declared as

```scala
type InstanceOf[A] <: A
```

and (privately) implemented as

```scala
type InstanceOf[A] = A
```

Now implementations of `Monad` wouldn't have to also extend `Bind`, `Applicative`, `Apply`, `Functor`. They would just extend `Monad.Class` and be "tagged" as `Monad` by zero-cost wrapping in `InstanceOf`.

(Implicits for `Monad`, if any, would now have to go in the companion of `Monad.Class`.)

jbgi:
Interesting, need to try!

TomasMikula:
Travis says that some tuts need to be fixed.

jbgi:
@TomasMikula quick update: I tested your suggested encoding this week-end. Works very well and avoid quite a bit of boilerplate classes. Currently preparing a replacement for this PR use it. (also I am thinking that we could also use it in 7.3.x ...)

TomasMikula:
Good to hear! Though I think it's fine to finish this PR (by fixing the build) and do the refactoring in the next round.

jbgi:
closing in favor of #1647.

PR #1626, opened 2018-01-24
[WIP] cofoldable and cofoldable1

vmarquez:
Something @jdegoes and I were talking about a couple months ago, finally had time to play around a bit more.  Would love some feedback if this would be useful.  I'd also like to put it into Scalaz8 if we agree it's the right thing.   @edmundnoble @fommil and anyone else, thoughts?


core/src/main/scala/scalaz/Cofoldable.scala
```diff
@@ -0,0 +1,55 @@
+package scalaz
+
+trait Cofoldable[F[_], A] extends Cofoldable1[F, A] {
+  
+  def build[B](b: B)(f: B => Option[(A, B)]): F[A]
+  
+  def build1[B](b: B)(f: B => (A, Option[B])): F[A] =
+    build(b)(b => f(b) match {
```
tonymorris: `val x = build(b); x.map(bb => Some((x._1, bb))`

core/src/main/scala/scalaz/Cofoldable1.scala
```diff
@@ -0,0 +1,29 @@
+package scalaz
+
+trait Cofoldable1[F[_], A] {
+  
+  def build1[B](b: B)(f: B => (A, Option[B])): F[A]
```
tonymorris: Noice.

tonymorris:
I suggest some more library support. Top of my head:

    def infinite[A](start: A, f: A => A)(implicit F: Cofoldable[F]): F[A] = …

Looks good!

xuwei-k:
please update [`project/GenTypeClass.scala`](https://github.com/scalaz/scalaz/blob/series/7.3.x/project/GenTypeClass.scala)

jbgi:
Given #1602, we could also have the following support:
```scala
implicit def reducingCofoldable1[A, F[_]](implicit R: Reducer[A, F[A]]): Cofoldable1[F, A]

implicit def reducingCofoldable[A, F[_]](implicit R: Reducer[A, F[A]], M: Monoid[F[A]]): Cofoldable[F, A]
```
(not sure if those methods should be implicit though)

core/src/main/scala/scalaz/Cofoldable1.scala
```diff
@@ -0,0 +1,29 @@
+package scalaz
+
+trait Cofoldable1[F[_], A] {
```
jbgi: why not the more general definition:
```scala
trait Cofoldable1[F, A] {
  def build1[B](b: B)(f: B => (A, Option[B])): F
}
```
? 
type-inference?
(also it would allows `Reducer[C, M] extends Cofoldable1[M, C]`)

fommil:
Initially I thought this was going to be some generalisation of corecursive structures but this is much better :-) CanActuallyBuildFrom

core/src/main/scala/scalaz/Cofoldable.scala
```diff
@@ -0,0 +1,55 @@
+package scalaz
+
+trait Cofoldable[F[_], A] extends Cofoldable1[F, A] {
+  
+  def build[B](b: B)(f: B => Option[(A, B)]): F[A]
```
jbgi: I think it should be 
```scala
def build(init: S)(step: S => Maybe[(S, A)]): F[A]
```
to be coherent with [`CorecursiveList`](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/CorecursiveList.scala#L39) (that would be the canonical instance of `Cofoldable`)
(Alternatively CorecursiveList would need to be modified to match this method)

jdegoes:
This looks good but I'm not sure it's dual to `Foldable`, so I'd recommend the name `Unfoldable`.

fommil:
My eyes are acting up today so it's hard to give this a review... is this a typeclass generalisation of corecursive data structures like corecursivelist?

core/src/main/scala/scalaz/Cofoldable.scala
vmarquez: Good idea, I can change that.

core/src/main/scala/scalaz/Cofoldable1.scala
vmarquez: Let me think about that one...

vmarquez:
@jdegoes I'm not sure either, but it looks...pretty close?

foldRight is
`F[A]   =>           B =>A   =>  B => B`
build is 
`B => B => Option(B => A) => F[A]`

or 
'F[A] <= Option(B <= A) <= B <= B` .

*shrug*

@fommil this is the generalization of anything that can have something 'put into', and it has laws.  So, Set, ISet, List, Vector, DList, etc. 

It also lets us derive things like 'partition' (which I renamed here, but we can use partition) and 'filter', though we'd want to specialize for performance reasons.   This means we can abstract over data structures that we can add to in a lawful way.  So between this + Foldable, we'll have the Scalaz version of Scala's `Seq`.

fommil:
Sweet. All we need now is CanActuallyBuildFrom

jbgi:
this is definitely useful!

For the record, do we want to be able to unfold more complex data structure, like `Tree`s?
In which case we will need [something more evolved](https://hackage.haskell.org/package/unfoldable), but that can be kept for a future evolution.

tonymorris:
`unfoldr` is dual to `foldr`.

jbgi:
> unfoldr is dual to foldr.

by the way, should the default `Cofoldable`/`Unfoldable` instance for `IList` do an `unfoldr` or an `unfoldl`? I believe the later is what is currently implemented...

vmarquez:
@tonymorris @jdegoes ah, yeah, then if I'm not mistaken, I did a more general version of unfoldr (though I did the unfoldl here, I can can change). 

`unfoldr :: (b -> Maybe (a, b)) -> b -> [a]`

vmarquez:
@jbgi I didn't dig into the package much, but the build method as is should work with trees, no?  build1 will work for CoFree, OneAnd, etc. too.  Did you mean in a more efficient way or something else?  Not sure I follow.

jbgi:
@vmarquez yes, it should work for `Tree`, did not dig too much either before commenting sorry.
Now I understand that the main difference with the haskell package, is that it separate the responsibility of the unfold into two separate type-classes, so potentially allowing better code reuse (at the cost of clarity I think). Probably not worth it.

vmarquez:
@xuwei-k some renewed interest in this but I don't know why the TC gen is failing, any pointers on how I could fix that?

project/GenTypeClass.scala
```diff
@@ -93,8 +93,10 @@ object TypeClass {
   lazy val monadTell = TypeClass("MonadTell", |*->*|->*, extendsList = Seq(monad))
   lazy val monadReader = TypeClass("MonadReader", |*->*|->*, extendsList = Seq(monad), createSyntax = false)
   lazy val comonadStore = TypeClass("ComonadStore", |*->*|->*, extendsList = Seq(comonad), createSyntax = false)
-
+  
   lazy val bindRec = TypeClass("BindRec", *->*, extendsList = Seq(bind))
+  lazy val cofoldable = TypeClass("Cofoldable", |*->*|->*, extendsList = Seq(cofoldable1), createSyntax = true)
```
xuwei-k: Why not `*->*`?

vmarquez: The type is `trait Cofoldable[F[_], A] extends Cofoldable1[F, A]` seems the same as say, MonadTell, which also has `|*->*|->*`

edmundnoble: Agreed, this seems wrong to me. It should just be `F[_]`.

jbgi: There could be a case for `Cofoldable[F, A]` (so that is extends `Reducer[F, A]`) and then we could have instances like `Cofoldable[String, Char]`. Also would avoid duplication between the two type classes.

vmarquez:
Let's look at what we gain and lose from having `F[_], A` vs. `F[_]`.  The former we gain *laws*.   The downside is some awkwardness when working with say, Maps.   Any others?

vmarquez:
@jbgi I don't really see how we'd get the same ability to derive the methods I was able to with the `F[_], A` encoding.   It's possible I'm not thinking of a way to do it though.

jbgi:
@vmarquez the way is to move the derived functions to the companion object: https://github.com/vmarquez/scalaz/pull/2
Also it would be nice for `unfoldr` to have the same signature in `Cofoldable` and in `Reducer`.

But, above all, we probably need a coherent story between this PR and the one merged in scalaz8. Maybe one way would be to have both `Cofoldable[F[_]]` and `Unfoldable[F, A]` (renaming scalaz8 `Unfoldable` into `Cofoldable`).

edmundnoble:
Just because we gain a law, doesn't mean we're doing something good. Laws are only necessary to restrict very general types; with more specific types, we need no laws. Being generic over `A` in the method and not the type class gives the user much more power; if we want to have this genericity in the method and not the type class, we should do the same for `Foldable` and `Traversable`.

The version of `Unfoldable` with `A` attached is well known in Haskell; this is just something from `mono-traversable` which provides monomorphic versions of `Foldable`, `Traversable`, `Functor`, etc. I am not opposed to having these, but only as *separate* type classes, hopefully with the `Mono` prefix because having two things called `Unfoldable` and `Cofoldable` is just crazytalk to me.

PR #1633, opened 2018-01-28
Cord, Biconst, Catenable1, Identity

edmundnoble:
Also added just a couple other things, like IList methods, and a replacement of the summon methods in Prelude with companion aliases because otherwise we literally cannot refer to type class companions.

Edit: I also want to note that this Cord implementation was heavily, *heavily* inspired by @alexknvl's suggested implementation in the gitter.


base/shared/src/main/scala/scalaz/data/Cord.scala
```diff
@@ -0,0 +1,65 @@
...
+}
+
+object CordImpl extends CordModule {
+  // Cord = (String | Concat, Int)
+  // Concat = (String | Concat, String | Concat)
+
+  type Cord = CordRepr
+
+  final class CordRepr(val under: AnyRef, val length: Int)
+  final class Concat(val left: AnyRef, val right: AnyRef)
```
jdegoes: Awesome.

base/shared/src/main/scala/scalaz/data/IList.scala
```diff
@@ -2,12 +2,21 @@ package scalaz
...
 trait IListModule {
   type IList[A]
 
   def uncons[A](as: IList[A]): Maybe2[A, IList[A]]
 
+  def cons[A](a: A, list: IList[A]): IList[A]
+
+  def concat[A](as1: IList[A], as2: IList[A]): IList[A]
+
+  def empty[A]: IList[A]
```
jdegoes: `empty` and `concat` are very valuable identifiers. Should they be ceded to `IList`, or should we make users utilize `Monoid` functions for `IList`?

base/shared/src/main/scala/scalaz/data/IList.scala
```diff
@@ -17,6 +26,33 @@ private[data] object IListImpl extends IListModule {
   def uncons[A](as: IList[A]): Maybe2[A, IList[A]] =
     Fix.unfix[Maybe2[A, ?]](as)
 
+  def cons[A](a: A, list: IList[A]): IList[A] =
+    Fix.fix[Maybe2[A, ?]](Maybe2.just2(a, list))
+
+  def concat[A](i1: IList[A], i2: IList[A]): IList[A] = {
+    @tailrec
+    def go(as1: IList[A], as2: IList[A]): IList[A] = {
+      Maybe2.toOption2(uncons(as1)) match {
```
jdegoes: Will be interesting to see how stupid `scalac` will be with this pattern match.

base/shared/src/main/scala/scalaz/data/Identity.scala
```diff
@@ -1,4 +1,34 @@
...
+
+    def subst[A]: Identity[A] Is A
+
+    implicit val monad: Monad[Identity]
+}
+
+private[data] object IdentityImpl extends IdentityModule {
+    type Identity[A] = A
+
+    def runIdentity[A](identity: A): A = identity
```
jdegoes: `def runIdentity[A](identity: Identity[A]): A = ...`

base/shared/src/main/scala/scalaz/typeclass/Debug.scala
```diff
@@ -1,19 +1,28 @@
...
+/** A typeclass describing types which can be meaningfully represented as a `String`, for debugging purposes.
   */
 trait Debug[A] {
   /** Produce a `String` representation of `a`.
     */
-  /* todo: when we have a `Cord`, change this to return that instead,
-   * todo: and add a `def debugs(a: A): String = debug(a).toString`
-   */
-  def debug(a: A): String
+  def debug(a: A): Cord = Cord(debugs(a))
```
jdegoes: 🎉

base/shared/src/main/scala/scalaz/typeclass/Foldable.scala
```diff
@@ -11,4 +11,8 @@ trait Foldable[F[_]] {
 
   // TODO Use IList (`toIList`)
   def toList[A](fa: F[A]): List[A] = foldLeft(fa, List[A]())((t, h) => h :: t).reverse
-}
\ No newline at end of file
+}
+
+object Foldable {
+  def apply[F[_]](implicit F: Foldable[F]): Foldable[F] = F
```
jdegoes: I like this change. 👍

benchmarks/src/main/scala/scalaz/data/StringBalancedConcatFoldBenchmark.scala
```diff
@@ -0,0 +1,50 @@
+package scalaz
```
jdegoes: Could be worth including Cats for reference like the `IO` benchmarks.

edmundnoble: cats doesn't have a Cord; I will include Scalaz 7's Cord though.

jdegoes: It has `Show` though. The benchmark could compare `Show` vs `Show` (7) vs `Debug` (8).

edmundnoble: Good point.

base/shared/src/main/scala/scalaz/data/IList.scala
edmundnoble: I have an idea here for later: we stop providing the `*Functions` traits in the prelude, and instead we put a bunch of functions inside a new `Z` object.

edmundnoble:
This just in: we're faster than `StringBuilder` for strings that do *not* fit in cache.

base/shared/src/main/scala/scalaz/typeclass/Debug.scala
```diff
@@ -1,19 +1,28 @@
...
+  def debugs(a: A): String = Cord.fold(debug(a))
```
jbgi: At least on of the `debug[s]` methods must be overridden: would be safer to use the `Alt` trait pattern and let the two methods abstract.

edmundnoble: Gotcha.

jdegoes:
@edmundnoble Can you update this so we can get it merged in?

base/shared/src/main/scala/scalaz/data/Cord.scala
```diff
@@ -0,0 +1,70 @@
...
+  val empty: Cord = new CordRepr("", 0, 1)
+
+  def length(cord: Cord): Int = cord.length
+
+  def unsafeAppendToH(rights: Array[AnyRef], out: Array[Char], cord: AnyRef): Unit = {
+    var current = cord
+    var stackPtr = 0
+    var outputPtr = 0
+    while (current != null) {
+      if (current.getClass == classOf[String]) {
```
NeQuissimus: Is there an advantage of this over `current.isInstanceOf[String]`?

alexknvl: Doesn't walk the inheritance tree (in general, I think in this case it will just check if the class is `final` and then do `_.getClass == classOf[String]`). I think we tested it and it was a bit faster.

jdegoes:
@edmundnoble Just a few conflicts to resolve and we can merge this in. Really looking forward to a `Cord` this fast.

edmundnoble:
Closing this in favor of smaller PRs.

PR #1635, opened 2018-02-01
Scala 2.13.0-M3

xuwei-k:


core/src/main/scala/scalaz/EitherT.scala
```diff
@@ -324,10 +324,6 @@ sealed abstract class EitherTInstances2 extends EitherTInstances3 {
 }
 
 sealed abstract class EitherTInstances1 extends EitherTInstances2 {
-  implicit def eitherTMonad[F[_], L](implicit F0: Monad[F]): Monad[EitherT[F, L, ?]] =
```
xuwei-k: This is unnecessary because same restriction of `eitherTMonadError`

https://github.com/scalaz/scalaz/blob/b9666d892ef287657156dd9b2c62bd3945c02817/core/src/main/scala/scalaz/EitherT.scala#L313

core/src/main/scala/scalaz/Free.scala
```diff
@@ -386,15 +386,12 @@ object Trampoline extends TrampolineInstances {
 }
 
 sealed trait TrampolineInstances {
-  implicit val trampolineInstance: Monad[Trampoline] with Comonad[Trampoline] with BindRec[Trampoline] =
-    new Monad[Trampoline] with Comonad[Trampoline] with BindRec[Trampoline] {
-      override def point[A](a: => A) = return_[Function0, A](a)
-      def bind[A, B](ta: Trampoline[A])(f: A => Trampoline[B]) = ta flatMap f
+  implicit val trampolineInstance: Comonad[Trampoline] =
```
xuwei-k: avoid `ambiguous implicit` error

core/src/main/scala/scalaz/std/Future.scala
```diff
@@ -65,8 +65,10 @@ private class FutureInstance(implicit ec: ExecutionContext) extends Nondetermini
   def raiseError[A](e: Throwable): Future[A] =
     fail(e)
 
-  def handleError[A](fa: Future[A])(f: Throwable => Future[A]): Future[A] =
-    fa.recoverWith(PartialFunction(f))
+  def handleError[A](fa: Future[A])(f: Throwable => Future[A]): Future[A] = {
+    val pf: PartialFunction[Throwable, Future[A]] = { case x => f(x) }
```
xuwei-k: https://github.com/scala/scala/commit/0cddad7946b3f96aaa35e780ffb10e64350509ae

Issue #1629, opened 2018-01-24
scalaz doesn't compile in what will soon become Scala 2.13.0-M3


SethTisue:
the compilation errors are in the `coreJVM` project. the full log is at https://scala-ci.typesafe.com/view/scala-2.13.x/job/scala-2.13.x-integrate-community-build/902/consoleFull

sample error (from a bunch of similar ones):

```
[scalaz] [error]  both method writerTraverse in class WriterTInstances1 of type [W]=> scalaz.Traverse[[β$19$]scalaz.WriterT[[X]X,W,β$19$]]
[scalaz] [error]  and method writerTTraverse in class WriterTInstances0 of type [F[_], W](implicit F0: scalaz.Traverse[F])scalaz.Traverse[[γ$22$]scalaz.WriterT[F,W,γ$22$]]
[scalaz] [error]  match expected type scalaz.Functor[[β$6$]scalaz.WriterT[[X]X,E,β$6$]]
[scalaz] [error]     new Adjunction[Writer[E, ?], Reader[E, ?]] {
[scalaz] [error]         ^
```

I'm not sure what this is about. I don't understand your source code, you don't know what changed in 2.13... we'll have to try to meet in the middle somewhere. getting the ball rolling with this ticket

until this is resolved, I have to remove scalaz from the 2.13 community build (this is happening at https://github.com/scala/community-builds/pull/673), which takes out a ton of downstream projects, too.

SethTisue:
P.S. the errors look similar to errors in Shapeless, seen in the same log, so it's possible that when @milessabin looks into that, what he learns will be relevant here as well

TomasMikula:
My guess is that it is because of https://github.com/scala/scala/pull/6069, which thanks to better unification resolves more implicits.

It will be problematic to make the same code compile with both Scala 2.11 and Scala 2.13. 😕

SethTisue:
but #6069 was in 2.12.4...?

TomasMikula:
Yes, but only under the flag `-Xsource:2.13`, which I don't see enabled in scalaz build.

SethTisue:
ahhhhh, right.

TomasMikula:
In the past, we enabled `-Ypartial-unification` and dropped 2.10 support in [one commit](https://github.com/scalaz/scalaz/pull/1367).

I personally would be fine with dropping 2.11 support in scalaz 7.3.x. But then, the community build uses 7.2.x anyway.

[...]

PR #1632, opened 2018-01-27
Resolve ambiguous implicits for Scala 2.13.0-M3

joroKr21:
In preparation. There may be more but this is what the tests caugth.

PR #1636, opened 2018-02-01
[7.2.x] Scala 2.13.0-M3

xuwei-k:
<del>partially(?)</del>  fix https://github.com/scalaz/scalaz/issues/1629
(can't build some modules due to https://github.com/rickynils/scalacheck/issues/382 )


core/src/main/scala/scalaz/Adjunction.scala
```diff
@@ -117,7 +117,7 @@ sealed abstract class AdjunctionInstances {
     }
 
   implicit def writerReaderAdjunction[E]: Adjunction[Writer[E, ?], Reader[E, ?]] =
-    new Adjunction[Writer[E, ?], Reader[E, ?]] {
+    new Adjunction[Writer[E, ?], Reader[E, ?]]()(WriterT.writerTraverse[E], implicitly) {
```
joroKr21: The real solution is to reorder the implicits in `WriterT`. I wonder why this didn't come up in #1632 (maybe something changed between 7.2.x and 7.3.x). I can help with this after scalacheck is out.

core/src/main/scala/scalaz/Free.scala
```diff
@@ -357,10 +357,14 @@ object Trampoline extends TrampolineInstances {
 
   def suspend[A](a: => Trampoline[A]): Trampoline[A] =
     Free.suspend(a)
+
+  implicit def trampolineComonad: Comonad[Trampoline] =
+    trampolineInstance
```
joroKr21: Neat, but why is it defined twice?

xuwei-k: removed

core/src/main/scala/scalaz/Adjunction.scala
xuwei-k: fixed

tests/src/test/scala/scalaz/FreeTTest.scala
```diff
@@ -115,7 +115,7 @@ object FreeTTest extends SpecLite {
 
   private def compilationTest = {
     val a: String \/ Int = \/-(42)
-    val b: FreeT[Maybe, String \/ ?, Int] = FreeT.liftMU(a)
+    val b: FreeT[Maybe, String \/ ?, Int] = FreeT.liftMU[Maybe, String \/ Int](a)
```
xuwei-k: :confused:

PR #1637, opened 2018-02-02
Make \/.left/right methods (instead of functions).

### LINKED ISSUE #1611: \/.left / \/.right allocate an extra function
`\/.left(a)`/`\/.right(b)` allocate a function only to apply it immediately. Is there a reason why not to change them to methods?

They used to be that way and were changed in https://github.com/scalaz/scalaz/commit/038937f338025f4c749318c2f50aaacad54f842c, for reasons unknown to me.

TomasMikula:
To avoid allocating an extra function object.

Resolves #1611.

### DIFF
--- a/core/src/main/scala/scalaz/Either.scala
+++ b/core/src/main/scala/scalaz/Either.scala
@@ -370,12 +370,12 @@ final case class \/-[A, B](b: B) extends (A \/ B) {
 object \/ extends DisjunctionInstances {
 
   /** Construct a left disjunction value. */
-  def left[A, B]: A => A \/ B =
-    -\/(_)
+  def left[A, B](a: A): A \/ B =
+    -\/(a)
 
   /** Construct a right disjunction value. */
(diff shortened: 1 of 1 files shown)


jbgi:
on the other hand, when targeting java8, it should compile to a unique lambda instance (since no capture).

TomasMikula:
Can you explain that more?

jbgi:
@TomasMikula:
```scala
scala> val f1 = \/.left[String, Int]
f1: String => String \/ Int = scalaz.$bslash$div$$$Lambda$5581/483161376@6e8cdae9

scala> val f2 = \/.left[Int, String]
f2: Int => Int \/ String = scalaz.$bslash$div$$$Lambda$5581/483161376@6e8cdae9

scala> f1 eq f2
res0: Boolean = true
```

TomasMikula:
Btw, see the new PR #1642

TomasMikula:
Closing in favor of one of #1642, #1645.

PR #1642, opened 2018-02-03
Smart constructors for -\/, \/-.

TomasMikula:


core/src/main/scala/scalaz/Either.scala
```diff
@@ -370,10 +370,18 @@ final case class \/-[A, B](b: B) extends (A \/ B) {
 object \/ extends DisjunctionInstances {
 
   /** Construct a left disjunction value. */
+  def inl[A, B](a: A): A \/ B =
```
jbgi: I'd rather see `left` and `right` spelled out.

jbgi:
alternatively, what about
```
object -\/ {
  def apply[A, B](a: A): A \/ B = new -\/(a)
}
```
This would preclude backports targeting scala 2.10 though.

TomasMikula:
I didn't know you could do that. See #1645

TomasMikula:
Closing in favor of #1645.

PR #1643, opened 2018-02-03
Simplified StringFunctions

poslegm:
Unused predicate removed from `asNumber` function. This parameter is not used inside the function. `intChars` predicate is sufficient for handling number parsing errors because too big Float and Double parse to `Infitinity` and are handled in a separate code.

### DIFF
--- a/core/src/main/scala/scalaz/std/String.scala
+++ b/core/src/main/scala/scalaz/std/String.scala
@@ -47,30 +47,30 @@ trait StringFunctions {
   def charsNel(s:String, e: => NonEmptyList[Char]) : NonEmptyList[Char] = charsNel(s) getOrElse e
 
   // Parsing functions.
-  private def intChars(s: String): Boolean = (s.length > 0) && (s.forall(c => c.isDigit || c == '-'))
+  private def intChars(s: String): Boolean = s.length == 1 && s.charAt(0).isDigit || ((s.length > 1) && s.stripPrefix("-").forall(_.isDigit))
 
-  private def asNumber[T](f: String => T, r: Tuple2[T, T], s: String)(implicit t: ClassTag[T]): Validation[String, T] =
+  private def asNumber[T](f: String => T, lowerBound: T, upperBound: T, s: String)(implicit t: ClassTag[T]): Validation[String, T] =
     try {
       Success(f(s))
     } catch {
-      case _: NumberFormatException if intChars(s) => Failure(s"${s} is outside of range for ${t} (${r._1} - ${r._2})")
+      case _: NumberFormatException if intChars(s) => Failure(s"${s} is outside of range for ${t} (${lowerBound} - ${upperBound})")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/scalaz/std/String.scala
```diff
@@ -48,9 +48,8 @@ trait StringFunctions {
 
   // Parsing functions.
   private def intChars(s: String): Boolean = (s.length > 0) && (s.forall(c => c.isDigit || c == '-'))
```
jbgi: If you will, `intChars` would need to be fixed, to accept `-` only if it appears as first char of a multi chars string.
I think a correct impl could be 
`s.length == 1 && s.charAt(0).isDigit || ((s.length > 1) && s.stripPrefix("-").forall(_.isDigit))`

core/src/main/scala/scalaz/std/String.scala
```diff
@@ -59,19 +58,19 @@ trait StringFunctions {
       case NonFatal(e) => Failure(e.getMessage)
```
jbgi: This is more of an open question: do we want `parse*` methods to catch exceptions like `NullPointerException`?
Personally I'd prefer we remove the `NonFatal(e)` cases, and deal with null input more explicitly (eg. via `scalaz.Maybe#fromNullable`) if needed.

core/src/main/scala/scalaz/std/String.scala
```diff
@@ -48,9 +48,8 @@ trait StringFunctions {
 
   // Parsing functions.
   private def intChars(s: String): Boolean = (s.length > 0) && (s.forall(c => c.isDigit || c == '-'))
-  private def floatChars(s: String): Boolean = (s.length > 0) && (s.forall(c => c.isDigit || c == '-' || c == '.'))
```
jbgi: `floatChars`, in the same maner as `intChars`, need to be fixed, but not removed.

core/src/main/scala/scalaz/std/String.scala
```diff
@@ -48,9 +48,8 @@ trait StringFunctions {
...
 
-  private def asNumber[T](f: String => T, b: String => Boolean, r: Tuple2[T, T], s: String)(implicit t: ClassTag[T]): Validation[String, T] =
+  private def asNumber[T](f: String => T, r: Tuple2[T, T], s: String)(implicit t: ClassTag[T]): Validation[String, T] =
```
jbgi: the `b` parameter must not be removed. It need to be used in place of `intChars`. (so that `floatChars` can be use as an alternative, by `parseDouble` and `parseFloat`)

core/src/main/scala/scalaz/std/String.scala
poslegm: Done

core/src/main/scala/scalaz/std/String.scala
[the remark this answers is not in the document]
poslegm: Done

PR #1644, opened 2018-02-04
remove "-Xsource:2.12" option if Scala 2.13. fix errors

xuwei-k:
https://github.com/scalaz/scalaz/issues/1629

### DIFF
--- a/tests/src/test/scala/scalaz/FreeTTest.scala
+++ b/tests/src/test/scala/scalaz/FreeTTest.scala
@@ -134,7 +134,7 @@ object FreeTTest extends SpecLite {
 
   private def compilationTest = {
     val a: String \/ Int = \/-(42)
-    val b: FreeT[Maybe, String \/ ?, Int] = FreeT.liftMU(a)
+    val b: FreeT[Maybe, String \/ ?, Int] = FreeT.liftMU[Maybe, String \/ Int](a)
   }
 
   object instances {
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/scalaz/FreeTTest.scala
```diff
@@ -134,7 +134,7 @@ object FreeTTest extends SpecLite {
 
   private def compilationTest = {
     val a: String \/ Int = \/-(42)
-    val b: FreeT[Maybe, String \/ ?, Int] = FreeT.liftMU(a)
+    val b: FreeT[Maybe, String \/ ?, Int] = FreeT.liftMU[Maybe, String \/ Int](a)
```
TomasMikula: Wasn't this necessary only for Scala 2.10?

xuwei-k: ¯\_(ツ)_/¯


- https://travis-ci.org/xuwei-k/scalaz/builds/337387331
- https://travis-ci.org/xuwei-k/scalaz/jobs/337387338#L1015

```
[error] /home/travis/build/xuwei-k/scalaz/tests/src/test/scala/scalaz/FreeTTest.scala:137:57: polymorphic expression cannot be instantiated to expected type;
[error]  found   : [S[_]]scalaz.FreeT[S,M.M,M.A]
[error]  required: scalaz.FreeT[scalaz.Maybe,[β$0$]String \/ β$0$,Int]
[error]     val b: FreeT[Maybe, String \/ ?, Int] = FreeT.liftMU(a)
[error]                                                         ^
```


travis-matrix

TomasMikula: A regression in 2.13.0-M3? Has it been reported?

Anyway,

```scala
val b: FreeT[Maybe, String \/ ?, Int] = FreeT.liftM(a)
```

seems to work across 2.11, 2.12, 2.13.....

xuwei-k: not yet reported

PR #1647, opened 2018-02-08
Type classes: use standard inheritance but hide it

jbgi:
Subtyping is hiden from type system (and implicit resolution) through invariant abstract type (zero-cost wrapper) in `InstanceOfModule`.

See [Original idea by @TomasMikula](https://github.com/scalaz/scalaz/pull/1623#issuecomment-359960281).

Ease of use of traditional encoding is thus preserved at the minor cost of wrapping instances construction in an `instanceOf` method call.

Remove all `Template` classes. This also force type class instance to use inheritance ([to preserve performance](https://github.com/scalaz/scalaz/pull/1620#issue-164107195)).


NeQuissimus:
This is certainly interesting :)

TomasMikula:
Looking good!

SystemFw:
Really nice

andyscott:
I also really like this. Nice work!

fosskers:
Benchmarks?

jdegoes:
@jbgi Nice work! Worth documenting in the Scalaz 8 design guide.

PR #1607, opened 2018-01-03
[WIP] Document type classes

NeQuissimus:
The goal of this PR is to document all the (currently existing) type classes into the `examples` module.

- [X] What is a type class? / index
- [ ] Applicative
- [ ] Apply
- [ ] Bind
- [ ] Category
- [ ] Choice
- [ ] Cobind
- [ ] Comonad
- [ ] Compose
- [ ] Contravariant
- [ ] Foldable
- [x] Functor
- [ ] InvariantFunctor
- [ ] IsContravariant
- [ ] IsCovariant
- [ ] Monad
- [ ] Monoid
- [ ] Phantom
- [ ] Profunctor
- [ ] Semigroup
- [ ] Show
- [ ] Strong
- [ ] Traversable

PR #1648, opened 2018-02-15
Document (very few) typeclasses

NeQuissimus:
Replacing #1607 with in-repo PR rather than fork. This way we can easily have other people contribute.

The goal of this PR is to document all the (currently existing) type classes into the `examples` module.

- [X] What is a type class? / index
- [ ] Applicative
- [ ] Apply
- [ ] Bind
- [ ] Category
- [ ] Choice
- [ ] Cobind
- [ ] Comonad
- [ ] Compose
- [ ] Contravariant
- [ ] Debug
- [ ] Foldable
- [X] Functor
- [ ] InvariantFunctor
- [ ] IsContravariant
- [ ] IsCovariant
- [ ] Monad
- [ ] Monoid
- [ ] Phantom
- [ ] Profunctor
- [ ] Semigroup
- [ ] Strong
- [ ] Traversable

### DIFF
--- a/example/src/main/resources/microsite/data/menu.yml
+++ b/example/src/main/resources/microsite/data/menu.yml
@@ -13,9 +13,11 @@ options:
     - title: Apply
       url: typeclass/Apply.html
       menu_section: typeclass
+    - title: Debug
+      url: typeclass/Debug.html
+      menu_section: typeclass
     - title: Functor
       url: typeclass/Functor.html
-      menu_section: typeclass
       menu_section: typeclass
     - title: Monad
       url: typeclass/Monad.html
@@ -24,10 +26,7 @@ options:
       url: typeclass/Monoid.html
       menu_section: typeclass
     - title: Semigroup
-      menu_section: typeclass
       url: typeclass/Semigroup.html
-    - title: Show
-      url: typeclass/Show.html
       menu_section: typeclass
 
   - title: Data classes
--- a/example/src/main/resources/microsite/data/menu.yml
+++ b/example/src/main/resources/microsite/data/menu.yml
@@ -16,9 +16,6 @@ options:
     - title: Functor
       url: typeclass/Functor.html
       menu_section: typeclass
-    - title: Semigroup
-      menu_section: typeclass
-      url: typeclass/Semigroup.html
       menu_section: typeclass
     - title: Monad
       url: typeclass/Monad.html
@@ -26,6 +23,9 @@ options:
     - title: Monoid
       url: typeclass/Monoid.html
       menu_section: typeclass
+    - title: Semigroup
+      menu_section: typeclass
+      url: typeclass/Semigroup.html
     - title: Show
       url: typeclass/Show.html
       menu_section: typeclass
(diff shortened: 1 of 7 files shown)


NeQuissimus:
I am declaring "PR bankruptcy" on this one... I simply don't get around to working on this... We should still do the little bit of work I have done...

jdegoes:
@NeQuissimus Don't give up yet! 😉

NeQuissimus:
@jdegoes I am not giving up, I am **decomposing** into multiple PRs :D

alexknvl:
:+1: Smaller PRs are better for various reasons IMHO.

example/src/main/tut/typeclass/index.md
```diff
@@ -4,3 +4,81 @@ title:  "Type classes"
...
+val bo = BusinessObject(1234L, 1234567L)
+
+businessObjectDebug.debug(bo)
+```
+
+However, this is not too convenient and very specific to our `BusinessObject`. Would it not be nicer to be able to say `bo.debug`?
+
+Luckily, **Scalaz** makes all this very easy.
+
+We need something similar to the following:
```
alexknvl: Not entirely sure if this is the right place to explain syntax enrichment :) Have you considered something along the lines of:

> However, this is not too convenient and very specific to our `BusinessObject`. Would it not be nicer to be able to say `bo.debug`? By default, Scalaz enriches all objects `o` that have an instance of typeclass `Debug` with a method `o.debug`:

and then continue with the `BusinessObject(1234L, 1234567L).debug` snippet.

alexknvl: AAh, I see. This is not `Debug` specific, my mistake :smile_cat:

example/src/main/resources/microsite/data/menu.yml
```diff
@@ -13,7 +13,7 @@ options:
     - title: Apply
       url: typeclass/Apply.html
       menu_section: typeclass
-    - tiel: Bifunctor
+    - title: Bifunctor
```
hrhino: *whistle, whistle* nothing to see here, right?

example/src/main/tut/typeclass/index.md
```diff
@@ -4,3 +4,81 @@ title:  "Type classes"
 ---
 
 # Type classes
+
+_In computer science, a type class is a type system construct that supports ad hoc polymorphism.
```
hrhino: "In computer science" is probably implied.

example/src/main/tut/typeclass/index.md
```diff
@@ -4,3 +4,81 @@ title:  "Type classes"
...
+
+Luckily, most of this ships in Scalaz by default.
+
+```tut:reset
+import scalaz.Scalaz._
+import scalaz.typeclass.DebugClass
+
+final case class BusinessObject(id: Long, value: Long)
+
+implicit val businessObjectDebug: Debug[BusinessObject] = instanceOf[DebugClass[BusinessObject]](b => s"BO[${b.id} = ${b.value}]")
```
hrhino: at some point we should explain `instanceOf` (but not here).

PR #1649, opened 2018-02-15
execute scalafmtCheck for build files and test files

xuwei-k:


TomasMikula:
Any instructions on what to do when that check fails?

All I get is

```
/home/travis/build/scalaz/scalaz/base/shared/src/main/scala/scalaz/data/AFix.scala isn't formatted properly!
```

Can I get it to output what exactly it doesn't like?

fommil:
just run the formatter I guess, `scalafmt`

Issue #1627, opened 2018-01-24
Replace usages of Seq with type class constrained variables (preferably) or IList


jbgi:
`Foldable`/`IList` => Good
`Seq` => Bad

in particular, `Nondeterminism`, `Future` and `Task` make heavy use of `Seq` in their API and need to be fixed.

jaapterwoerds:
I'm willing to take a stab at this.

PR #1664, opened 2018-03-05
1627 replace seq

jaapterwoerds:
As suggested in #1627, I replaced the usage of Seq by IList.

### DIFF
--- a/concurrent/src/main/scala/scalaz/concurrent/Future.scala
+++ b/concurrent/src/main/scala/scalaz/concurrent/Future.scala
@@ -248,7 +248,7 @@ object Future {
       fa flatMap f
     def point[A](a: => A): Future[A] = delay(a)
 
-    def chooseAny[A](h: Future[A], t: Seq[Future[A]]): Future[(A, Seq[Future[A]])] = {
+    def chooseAny[A](h: Future[A], t: IList[Future[A]]): Future[(A, IList[Future[A]])] = {
       Async { cb =>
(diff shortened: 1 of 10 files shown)


tonymorris:
:+1: 

It might also be worth writing a document making it clear why `Seq` has no use-cases.

hrhino:
Would it make sense to replace some of those with arbitrary `Foldable`/`Traversable` types? An `IList` is `O(n)` to construct if you happen to only have a `Foldable` lying around.

jaapterwoerds:
@tonymorris Good point, but I'm not sure where this would fit in the project and or if it it should be part of this PR. I am for giving the good example in the Scalaz code base and example and test and remove undesirable structures further.

@hrhino I will take a look to see if I can weaken the required abstraction in some cases.

tonymorris:
@jaapterwoerds yeah not in this PR. Perhaps open an issue for creating a document that specifies "the useful subset of scala" so that others can see it, and understand some of the reasoning.

Just a suggestion.

hrhino:
FWIW just this week I wanted to `ISet.unions` an `ISet` of `ISet`s. (Obviously it can't make a `Monad` so `flatMap` in general was out.) Making an `IList` of it would have been wasteful.

fommil:
are we good to merge?

jaapterwoerds:
@hrhino I've spent some time trying to figure out if I could replace IList by Foldable or Traversable but could not find any way. 

@fommil I don't have anything else to add so we're good AFAICS

PR #1677, opened 2018-03-26
Add a "by-name" variant of Endo

jbgi:
and use it when implementing `foldRight` in term of `foldMap`, so that "by-name" evaluation semantic is preserved, potentially avoiding stack overflows.

### DIFF
--- a/core/src/main/scala/scalaz/Endo.scala
+++ b/core/src/main/scala/scalaz/Endo.scala
@@ -15,15 +15,39 @@ final case class Endo[A](run: A => A) {
   final def andThen(other: Endo[A]): Endo[A] = other compose this
 }
 
+/** Endomorphisms using by-name evaluation.  They have special properties among functions, so
+  * are captured in this newtype.
+  *
+  * @param run The captured function.
+  */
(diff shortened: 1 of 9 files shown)


core/src/main/scala/scalaz/Endo.scala
```diff
@@ -15,16 +15,37 @@ final case class Endo[A](run: A => A) {
...
+
+  /** Call `other` with my result. */
+  final def andThen(other: EndoByName[A]): EndoByName[A] = other compose this
+}
+
 object Endo extends EndoInstances {
   /** Alias for `Endo.apply`. */
   final def endo[A](f: A => A): Endo[A] = Endo(f)
 
+  /** Alias for `Endo.apply`. */
```
hrhino: nit: no it's not

jbgi:
Alternatively, we could make current `Endo` use by-name evaluation; probably would not be a huge lost.

TomasMikula:
If this buys us some stack-safety or laziness, it would be good to add a test that demonstrates it (i.e. that would fail now and passes after this change), so that we don't lose it in the future.

jbgi:
@TomasMikula added the test.

PR #1681, opened 2018-03-28
Reducer: support unfolds without needing a monoid.

TomasMikula:
- Added `unfoldlOpt`, `unfoldrOpt`: unfolds that don't require monoid.
 - Removed `final` modifier from `unfoldl`.


jbgi:
nice! could you also add the corresponding forwards to `ReducerOps`? (probably better as a separate commit as those cannot be backported to 7.2.x)

TomasMikula:
Added.

PR #1683, opened 2018-03-28
One-sided reducers (LeftReducer, RightReducer)

TomasMikula:


jbgi:
But we can always implement `LeftReducer` / `RightReducer` / `Reducer` as soon as we have the semigroup + unit.
Do we want to only provide `LeftReducer` for linked list types? if not they having only `Reducer` is enough.

TomasMikula:
Yes, the idea is to provide only efficient operations. I'm not completely sure what implications it would have for current users of list reducer.

TomasMikula:
I guess such change (only `LeftReducer` for lists) should be part of this PR.

jbgi:
maybe we could have `FastLeftReducer`/`FastRightReducer` extends `Reducer` and `FastReducer` extends both `FastLeftReducer`/`FastRightReducer`?

TomasMikula:
Yeah, that sounds good.

jbgi:
Or maybe simply a `Slow` tag and implicit conversions from LeftReducer/RightReducer to a tagged Reducer. Wdyt?

TomasMikula:
Even a `Slow` reducer would typically be fast in one direction, so maybe `Right` and `Left` tags instead of `Slow`.

jbgi:
Not sure the distinction matter: if a method need a Reducer constraint, (as opposed to just LeftReducer or RightReducer) it means that if one of cons/snoc is slow, overall execution will be slow.

TomasMikula:
From `Reducer @@ Left` one could extract a fast `LeftReducer`. But not from `Reducer @@ Slow`.

jbgi:
Just to be sure we are talking about the same thing, I was thinking of 
```scala
implicit def slowReducer[C, M](implicit lr: LeftReducer[C, M]) : Reducer[C, M @@ Slow]
```
The tagged type being opaque being opaque we cannot recover the original instance (whatever the tag name), nor can I think of a case where we would want to.

TomasMikula:
Right, we can't get back the original `lr`, but we could use the resulting `Reducer[C, M @@ Slow]` as a fast `LeftReducer`.

jbgi:
But Reducer already extends LeftReducer, there is no notion of fast or tagged at this point (in the method with the Reducer constraint): only at call site it is decided to tag to force usage of api despite slow implementation.

jbgi:
So, let's recap:

`Reducer` is a type class introduced to improve algorithms efficiency.

`LeftReducer` and `RightReducer` could allow to better document the efficiency conditions of the said algorithms.
This is nice, but is it enough value to warrant the complexity of adding two new type classes?
After all the documentation is purely outside of the type-system so relatively weak in my opinion, as the two type classes do not bring any more free-theorems over `Reducer`, 

Also, many types that would only expose a `LeftReducer` should still expose a `snoc` for the sake of (marginal) efficiency: ie. `snoc(as: List[A], a: A)` is still a little bit more efficient than `append(as: List[A], a :: Nil : List[A])`.

Another concern is the interaction with `ApplicativePlus` when [deriving `Reducer` from `ApplicativePlus`](https://github.com/scalaz/scalaz/issues/1717)).
To be coherent we would need a `LeftPlus` and a `RightPlus`, in order to derive the corresponding `RightReducer` or `LeftReducer`.
But  given the number of sub-classes of `Plus`, introducing `LeftPlus`/`RightPlus` would not be viable.

So after pondering I think `LeftReducer` and `RightReducer` are not worth the added complexity. But I am still curious of other opinions.

TomasMikula:
> After all the documentation is purely outside of the type-system so relatively weak in my opinion

So is the difference between `cons` and `snoc`, or the difference between `foldLeft` and `foldRight`.

We could also have just one typeclass, say, `SemiReducer`.

`cons` and `snoc`, or `LeftReducer` and `RightReducer`, become discernible at the type level in the type-aligned case:

```scala
def cons(f: F[A, B], g: G[B, C]): G[A, C]
def snoc(g: G[A, B], f: F[B, C]): G[A, C]
```

> Also, many types that would only expose a `LeftReducer` should still expose a `snoc` for the sake of (marginal) efficiency: ie. `snoc(as: List[A], a: A)` is still a little bit more efficient than `append(as: List[A], a :: Nil : List[A])`.

Bad example: the `snoc` will start by transforming that `a` into `a :: Nil`, but I get your point.
Some slow `Reducer` might be provided, but maybe not implicitly.

> To be coherent we would need a `LeftPlus` and a `RightPlus`, in order to derive the corresponding `RightReducer` or `LeftReducer`.

I don't get the coherence problem, but we are not obliged to derive `LeftReducer`/`RightReducer` from `ApplicativePlus`.

PR #1685, opened 2018-03-29
Unfolds and early termination

TomasMikula:
Iterated multiplication may terminate when 0 is encountered.

This PR adds methods to `Semigroup`, `Monoid`, `Plus`, `PlusEmpty` that may take advantage of this fact.

It is demonstrated on a few cases:
 - `Int` multiplication
 - `Boolean` conjunction and disjunction
 - `Maybe` "addition" (`Plus`)

Finally, these potentially short-circuiting methods can be used to implement `foldMap` (and `fold`, `psum`, `psumMap`) on many data structures. As a demonstration, it is done for `Stream`. As a result, we get **`foldMap` that is both stack-safe _and_ short-circuiting.**

Consequently, we may deprecate `msuml`, since there's no more need for it (`psum` used to be short-circuiting and `msuml` stack-safe, but now `psum` is both).

### DIFF
--- a/core/src/main/scala/scalaz/Foldable.scala
+++ b/core/src/main/scala/scalaz/Foldable.scala
@@ -283,14 +283,13 @@ trait Foldable[F[_]]  { self =>
   def suml1Opt[A](fa: F[A])(implicit A: Semigroup[A]): Option[A] =
     foldLeft1Opt(fa)(A.append(_, _))
 
-  // should be revisited by https://github.com/scalaz/scalaz/issues/1663
   /**
    * Map elements to `G[B]` and sum using a polymorphic monoid ([[PlusEmpty]]).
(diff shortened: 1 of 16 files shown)


TomasMikula:
@fommil Even if there was an implicit `Monoid[G[A]]` different from the one implied by `Plus[G]`, in this case we want the one implied by `Plus[G]`, anyway. There's no danger of mixing up different monoids.

fommil:
Now. If we can get @jdegoes IO into scalaz 7.3 I think it'll just about tide me over and we can start to focus on 8.0 :smile:

jdegoes:
@TomasMikula Superb work, as always!

@fommil Nearly there. Another week's worth of work and it should be ready to backport to 7.3. Though I'm not volunteering for that job. 😆

PR #1688, opened 2018-03-29
Message first, version second in @deprecation messages.

hrhino:
Otherwise you get messages like:

```
there was one deprecation warning (since prefer leftT; this method's meaning will change in scalaz 7.3)
```

... which I doubt was the intention.


hrhino:
The alternative reading is encouraged by

```scala
@deprecated("7.2.19", since = "you really should prefer leftT; ...")
```

but that's frivolous.

fommil:
oops :blush:

PR #1689, opened 2018-03-30
Applicative short-circuiting

TomasMikula:
Some `Applicatives` have absorbing values (such as `Option`'s `None`) that may be used to terminate early.

This PR adds a method to `Apply` that may be overridden to take advantage of such short-circuiting.

```scala
trait Apply[F[_]] {
  def unfoldrOpt[S, A, B](seed: S)(f: S => Maybe[(F[A], S)])(implicit R: Reducer[A, B]): Maybe[F[B]]
}
```

This PR then overrides this method for `Apply[Maybe]`. (This should be done for other Applicatives with absorbing elements as well.)

Finally, `List` traversal is implemented using this potentially short-circuiting method, resulting in **traversal that is both stack-safe and short-circuiting** (for short-circuiting Applicatives), resolving an old dilemma. (This should be done for other Traversables as well.)

~~(Note that the first 6 commits are from #1685.)~~

### DIFF
--- a/core/src/main/scala/scalaz/Apply.scala
+++ b/core/src/main/scala/scalaz/Apply.scala
@@ -1,6 +1,8 @@
 package scalaz
 
 ////
+import scala.annotation.tailrec
+
 /**
  * [[scalaz.Applicative]] without `point`.
  *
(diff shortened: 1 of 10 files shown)


fommil:
Great stuff! This is the kind of thing that makes me excited for 7.3

PR #1692, opened 2018-04-03
deprecate impure methods that expose Throwable

fommil:
Continuing the discussion from #1687 

These methods either

- only turn partial functions into slightly less partial functions (which is a fairly useless replacement for a `try` / `catch` block)
- or, return `Throwable` values. `Throwable` is not referentially transparent, even for deterministic failures, thanks to the stacktrace (this is actually a really big deal, I've worked in at least two places that tried to cache/intern something similar to `Throwable \/ A` and it led to a lot of pain and suffering).

Generally it is my understanding that we plan on having more principled FP in scalaz 7.3+ and that there is little room for anything that compromises purity in the API for beginner convenience. However, we should not leave users without an option... and there are several presented. I would wager that most uses of these deprecated methods would be best replaced with an IO. If the user wants the same semantics as in the deprecated methods, `Try` is in the stdlib. Or they can call `attempt` with `identity` (but it is not recommended).

We should consider introducing data structures to assist dealing with legacy throwables and nulls. For now, IO is all we have, and the new `attempt` functions I introduced are a compromise to allow the use of **deterministic** legacy blocks of code.

### DIFF
--- a/core/src/main/scala/scalaz/Catchable.scala
+++ b/core/src/main/scala/scalaz/Catchable.scala
@@ -16,6 +16,7 @@ package scalaz
  * total.
  */
 ////
+@deprecated("No laws, and Throwable is not referentially transparent. Prefer MonadError", "7.2.21")
 trait Catchable[F[_]]  { self =>
   ////
 
--- a/core/src/main/scala/scalaz/Either.scala
+++ b/core/src/main/scala/scalaz/Either.scala
@@ -368,12 +368,14 @@ object \/ extends DisjunctionInstances {
   def fromEither[A, B](e: Either[A, B]): A \/ B =
     e fold (left, right)
 
+  @deprecated("Throwable is not referentially transparent, use attempt", "7.2.21")
   def fromTryCatchThrowable[T, E <: Throwable](a: => T)(implicit nn: NotNothing[E], ex: ClassTag[E]): E \/ T = try {
     \/-(a)
   } catch {
     case e if ex.runtimeClass.isInstance(e) => -\/(e.asInstanceOf[E])
   }
 
+  @deprecated("Throwable is not referentially transparent, use attempt", "7.2.21")
   def fromTryCatchNonFatal[T](a: => T): Throwable \/ T = attempt(a)(identity)
 
   /**
--- a/core/src/main/scala/scalaz/EitherT.scala
+++ b/core/src/main/scala/scalaz/EitherT.scala
@@ -279,6 +279,7 @@ object EitherT extends EitherTInstances {
   def fromEither[F[_], A, B](e: F[Either[A, B]])(implicit F: Functor[F]): EitherT[F, A, B] =
     apply(F.map(e)(_ fold (\/.left, \/.right)))
 
+  @deprecated("Throwable is not referentially transparent, use \\/.attempt", "7.2.21")
   def fromTryCatchThrowable[F[_], A, B <: Throwable](a: => F[A])(implicit F: Applicative[F], nn: NotNothing[B], ex: ClassTag[B]): EitherT[F, B, A] =
     try {
       rightT(a)
@@ -286,6 +287,7 @@ object EitherT extends EitherTInstances {
       case e if ex.runtimeClass.isInstance(e) => leftT(F.point(e.asInstanceOf[B]))
     }
 
+  @deprecated("Throwable is not referentially transparent, use \\/.attempt", "7.2.21")
   def fromTryCatchNonFatal[F[_], A](a: => F[A])(implicit F: Applicative[F]): EitherT[F, Throwable, A] =
     try {
       rightT(a)
--- a/core/src/main/scala/scalaz/Maybe.scala
+++ b/core/src/main/scala/scalaz/Maybe.scala
@@ -176,12 +176,14 @@ object Maybe extends MaybeInstances {
   final def fromOption[A](oa: Option[A]): Maybe[A] =
     std.option.cata(oa)(just, empty)
 
+  @deprecated("Not referentially transparent, use attempt", "7.2.21")
   def fromTryCatchThrowable[T, E <: Throwable](a: => T)(implicit nn: NotNothing[E], ex: ClassTag[E]): Maybe[T] = try {
     just(a)
   } catch {
     case e if ex.runtimeClass.isInstance(e) => empty
   }
 
+  @deprecated("NonFatal is too forgiving, use attempt", "7.2.21")
   def fromTryCatchNonFatal[T](a: => T): Maybe[T] = try {
     just(a)
   } catch {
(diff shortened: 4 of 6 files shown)


TomasMikula:
It is not obvious to me that `attempt` is always an improvement over `fromTryCatchThrowable`.

> only turn partial functions into slightly less partial functions

One might say that `attempt` encourages partiality:

```scala
attempt(foo) {
  case e: SomeException => handle(e)
}
```

This only handles `SomeException` and anything else will be converted to a `MatchError`. This seems worse than

```scala
fromTryCatchThrowable[A, SomeException](foo).leftMap(e => handle(e))
```

where the original uncaught exception is propagated unchanged.

> or, return `Throwable` values.

As you noted, the user can easily call `attempt` with `identity` handler and retain the `Throwable`.

---

I don't care much about this issue, but migration from `fromTryCatchThrowable` to `attempt` does feel somewhat arbitrary to me.

fommil:
```scala
attempt(foo) {
  case e: SomeException => handle(e)
}
```

the problem here is that a partial function has been provided instead of a total function.

fommil:
@TomasMikula yeah, unfortunately. We can only hope the user has scalafix enabled! Actually, I wonder if there is a decent linter rule to turn silent acceptance into a warning when the developer matches on an unsealed interface, but does not have a catch-all clause.

fommil:
I requested it https://github.com/scalacenter/scalafix/issues/670

fommil:
oh great, travis isn't setup to handle revert merges.

fommil:
this entire discussion seems to have lost its steam.

@xuwei-k what are your thoughts? I can't merge anyway because travis building the `/push` job, which fails.

fommil:
btw, there is `scalaz.effect.IoExceptionOr` which is in direction of what @tonymorris was talking about. If we were to move it into `scalaz-core` we could deprecate / remove everything that deals with `Throwable` or `Exception` and standardise on that as the best way to deal with legacy deterministic exceptions. `IO` still preferred for non-deterministic.

japgolly:
> /cc @japgolly

Sorry for the delay. I would like to be involved in a constructive discussion but it's a bit hard to get time (and enough sleep) to spend on this. Having a 2 year-old and a newborn makes just a newborn seem like a holiday. :sob: Let me come back to this over the weekend.

japgolly:
There are three propositions motivating this change, right? Exceptions are side-effects, throwables are non-RT, drive beginners in the right direction. I'll try to address all three:

Exceptions are side-effects. Well, are they and are they always? They can subvert types and control-flow, and affect the stack in opaque ways. In that sense, yes they are. On the other hand, if an exception is thrown deterministically and caught, then the fact that a side-effect occurred can be recovered, and functional-purity can be regained. Simple example: `def x = try ??? catch {case _ => 0}` is pure and RT. Callers of `x` have no idea that an exception occurred. This is in direct contrast to most other side-effects for which IO is appropriate. There's no way you can recover purity from `println` or modifying a global variable because they can be externally observed beyond your ability to control. Exceptions are in the same bucket of crap as is null. Most Java and some non-FP libs use null liberally; null subverts types and breaks parametricity but we handle that by never using nulls ourselves and when we need to interface with crappy code, we use Option. In the same way, most Java and some non-FP libs throw exceptions liberally; exceptions subvert types and break parametricity but we handle that by never throwing exceptions ourselves and when we need to interface with crappy code, we use catch it and use a disjunction. IO would be waaaaaay too strong a solution to otherwise-pure exceptions, just as it would be too strong a solution to null handling. Thus we should allow exception catching methods on `\/` and not push for `IO` as an alternative.

Throwables are non-RT. Yeah sure but not all parts of Throwables are. Firstly, code that doesn't use any non-RT parts of Throwables, is itself RT. Sensible exception handling should catch throwables and extract only the useful bits (if any), instead returning a pure, more descriptive type. The new `attempt` method goes in that direction but only for nonFatal errors, not for all throwables (good for JS) nor certain subtypes (good for Java). Secondly, (and prepare any potential ivory towers (such as I've occupied myself in the past too) for an offence): non-RT isn't universally a bad thing. It's bad in nearly all cases because in nearly all cases, logic and computation will be affected. However when it comes to throwables and in particular, their stack traces, no one (sane) would ever use that data to affect logic or computation. Instead it's just some non-RT data that gets exported to humans via screen or log-service, for informative purposes. Different stack-trace values occurring due to refactoring (RT) won't cause any incorrectness.

Driving beginners in the right direction. Sure. A noble thing to do and I've no problem with it, don't get me wrong. But it shouldn't be at the expense of tools for people who know what they're doing. My arguments above are acceptable motivation for a user both wanting to catch exceptions without `IO`, and even sometimes passing a Throwable around. The methods to do so shouldn't be deprecated, causing the need for people to manually edit their codebase to do the exact same thing with a hoop to jump through. Not only that, the very important part of good FP is creating the tiniest building blocks meaningfully possible, and composing them into higher-level constructs. From my sadly-shitty understanding of category theory, that means finding factorisations and preferring universal constructions. The deprecations were discussing here deprecate the universal form and force everyone to a factored version which to me, is crazy, it should be the other way around. It would be one level of objection if this were all new code that we're talking about, but the fact that in addition to the factorisation argument, users will be forced to change their existing code is a whole other level. (And yes, deprecation is a big deal - many teams enforce zero compilation warnings).

fommil:
Thanks for the detailed response. Are you mostly saying that we should leave the "bad" methods in place that expose Throwable so people who know what they are doing can use them? In that case, why do they not just use try / catch which makes it clearer what they are doing? We'll even do this ourselves in the 7.3 stringops parsers for increased efficiency.

With regards to non-Fatal exceptions, I think we can agree that these are program errors and we can't recover from them. Actually maybe we should put additional matches in to catch Exception things that NonFatal passes through. Catching Throwable is a total moot point in my experience as it kicks in when running out of memory out which point we can't reason about anything: even the code we've written to deal with it will most likely fail.

In 7.3 we could move the data structure from the effect package into core and remove all of these things, put put all the convenience methods we want on its companion with detailed notes.

jdegoes:
>  Actually maybe we should put additional matches in to catch Exception things that NonFatal passes through. 

This seems like it's a good idea but it's actually not, as the `NonFatal` extractor will actually allocate memory, which of course is undefined for out of memory errors; and even logic is undefined for some errors such as internal VM errors.

fommil:
NonFatal is a lie

jdegoes:
@fommil Indeed, and the myth has been heavily repeated so often by Scala gurus it's now considered a fact.

jdegoes:
Oh you can do it the way I did in Scalaz 8, which is `case x : Throwable if nonFatal(x)`, which will not allocate heap but will still increase stack as the method is executed. Plus whatever crap scalac does behind the scenes.

fommil:
I'm not seeing consensus on this in the scalaz7 line, so closing.

PR #1698, opened 2018-04-08
Void and denial inequality.

alexknvl:
The api is currently rather impoverished since the majority of the methods I have in `leibniz` require additional data types to be defined first. `NotIs` could be implemented as a newtype over `(A === B) => Void`, but I would like to first add a `@newtype` macro / compiler plugin to scalaz so that I don't pollute the code with unnecessary details of newtype encoding.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/Void.scala
+++ b/base/shared/src/main/scala/scalaz/data/Void.scala
@@ -1,0 +1,13 @@
+package scalaz
+package data
+
+trait VoidFunctions {
+  def unsafeVoid: Void = throw new Void.Absurd
+}
+
+object Void extends VoidFunctions {
+  private[data] trait Tag
+  type Void <: Nothing with Tag
+
+  private[data] final class Absurd extends RuntimeException
+}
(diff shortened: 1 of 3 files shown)


base/shared/src/main/scala/scalaz/data/Void.scala
```diff
@@ -0,0 +1,13 @@
+package scalaz
+package data
+
+trait VoidFunctions {
+  def unsafeVoid: Void = throw new Void.Absurd
```
jbgi: is that api necessary?

Also I'd have expected `absurd::Void->A` here

alexknvl: It might be useful to assert void in `(Int === String) => Void` as an example. Due to Scala's subtyping rules absurd is the same as identity. `absurd` would likely result in DCE warnings.

jbgi: ah yes, `Nothing` was there but I missed it! ;)

jdegoes:
@alexknvl Without `absurd`, this will not be a useful definition of `Void`.

Why the need to use `Nothing`? This just does lead to DCE warnings that become errors in some projects.

PR #1699, opened 2018-04-09
Port scalaz8 IO to 7.2.x in new module

jmcardon:


build.sbt
```diff
@@ -96,6 +96,22 @@ lazy val concurrent = Project(
...
 )
 
+lazy val ioeffect = Project(
+  id = "ioeffect",
+  base = file("ioeffect")
+).settings(
+  standardSettings,
+  name := IOEffectName,
+  osgiExport("scalaz.ioeffect"),
+  OsgiKeys.importPackage := Seq("javax.swing;resolution:=optional", "*"),
```
hrhino: wait what `javax.swing` what how why?

build.sbt
```diff
@@ -96,6 +96,22 @@ lazy val concurrent = Project(
...
+  libraryDependencies ++=
+  Seq("org.specs2" %% "specs2-core"          % "4.0.0" % "test",
+    "org.specs2" %% "specs2-matcher-extra" % "4.0.0" % "test"),
```
xuwei-k: specs2 4.x does not support Scala 2.10 (and Java 7?). but scalaz 7.2.x support Scala 2.10 and Java 7.

build.sbt
emilypi: @hrhino [https://imgflip.com/i/28029b](because)

fommil:
This is awesome to see! I haven't done a review yet, I will try to find the time but please don't be blocked by me.

I was thinking perhaps it would make sense in a separate repo/artifact so that we don't have to make a scalaz 7.2 release every time it is updated, and also doesn't come with bincompat guarantees. I think if we broke bincompat at this stage of 7.2.x it would be very embarrassing for scalaz even if it's in a new module.

xuwei-k:
> I was thinking perhaps it would make sense in a separate repo/artifact so that we don't have to make a scalaz 7.2 release every time it is updated, and also doesn't come with bincompat guarantees.

👍

jdegoes:
This is great! But: do you want a couple bug fixes and the new bifunctor design (`IO[E, A]`)? Nearly ready to submit that PR to 8.x.

I know a lot of folks are looking forward to using this with 7.x.

emilypi:
@jdegoes 
for sure, send 'em our way

ioeffect/src/main/scala/scalaz/ioeffect/IOInstances.scala
```diff
@@ -0,0 +1,24 @@
...
+//    override final def ap[A, B](ma: IO[A])(mf: IO[A => B]): IO[B] =
+//      ma.flatMap(a => mf.map(f => f(a)))
+
+
+
+
+//////    override final def pure[A](a: A): IO[A] = IO.now(a)
+////
+////    override final def flatMap[A, B](ma: IO[A])(f: A => IO[B]): IO[B] =
+//      ma.flatMap(f)
```
xuwei-k: unnecessary comments?

fommil:
@jdegoes can you please set up a repo `scalaz/backports` or `scalaz/ioeffect` so we can continue development there and close this PR?

Milyardo:
Should `effect` be deleted from 7.3 if `ioeffect` gets published as a separate project?

fommil:
@milyardo, no it will be in 7.3. This is for 7.2.

fommil:
@Milyardo I don't know if deprecation is the right way to go. Many basic uses of IO will remain the same. But if it is a dramatic change, and there is a `scalaz-ioeffect` package available to use in 7.2, I'd be happy with adding deprecation warnings.

fommil:
@jdegoes thanks! Not to be a total PITA but do you mind changing it to `scalaz-ioeffect` to make it nicer when forked?

jdegoes:
@fommil Your wish is my command.

PR #1700, opened 2018-04-09
Iso.

alexknvl:


base/shared/src/main/scala/scalaz/data/Iso.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  final case class Refl[A]() extends Iso[A, A] {
+    def to(a: A): A   = a
+    def from(b: A): A = b
+  }
+
+  def id[A]: Iso[A, A] = Refl[A]()
+
+  object Product {
+    type ⨂[A, B] = (A, B)
```
TomasMikula: Can we add a comment with the unicode code when using special characters?

Also, this is the "big" tensor product, not the binary tensor product (`⊗`), but I guess it's justified by better readability. (The binary one looks unintelligible in my editor.)

base/shared/src/main/scala/scalaz/data/Iso.scala
```diff
@@ -0,0 +1,139 @@
...
+      unsafe({ case (a, b) => (b, a) }, { case (b, a) => (a, b) })
+
+    final def unit[A]: Iso[A, A ⨂ Id] =
+      unsafe({ case a => (a, ()) }, { case (a, ()) => a })
+
+    final def first[A, B, C](iso: Iso[A, C]): Iso[A ⨂ B, C ⨂ B] =
+      iso and id
+
+    final def second[A, B, C](iso: Iso[B, C]): Iso[A ⨂ B, A ⨂ C] =
+      id and iso
```
TomasMikula: Why is `and` on the `Iso` itself and `first` and `second` here?

base/shared/src/main/scala/scalaz/data/Iso.scala
```diff
@@ -0,0 +1,139 @@
...
+    def to(fa: F[A]): F[B]   = F.map(fa)(ab.to)
+    def from(fb: F[B]): F[A] = F.map(fb)(ab.from)
+  }
+
+  def liftCtF[F[_]](implicit F: Contravariant[F]): Iso[F[A], F[B]] = new Iso[F[A], F[B]] {
+    def to(fa: F[A]): F[B]   = F.contramap(fa)(ab.from)
+    def from(fb: F[B]): F[A] = F.contramap(fb)(ab.to)
+  }
+
+  def liftInvF[F[_]](implicit F: InvariantFunctor[F]): Iso[F[A], F[B]] = new Iso[F[A], F[B]] {
```
hrhino: This is `liftInvF` here but `substF` above...

base/shared/src/main/scala/scalaz/data/Iso.scala
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  def flip: Iso[B, A] = new Iso[B, A] {
+    def to(b: B): A              = ab.from(b)
+    def from(a: A): B            = ab.to(a)
+    override val flip: Iso[A, B] = ab
+  }
+
+  def and[I, J](ij: Iso[I, J]): Iso[(A, I), (B, J)] =
+    Iso.unsafe({ case (a, i) => (ab.to(a), ij.to(i)) }, { case (b, j) => (ab.from(b), ij.from(j)) })
```
hrhino: Why do some of these make a new anonymous `Iso` and others use `unsafe`?

base/shared/src/main/scala/scalaz/data/Iso.scala
TomasMikula: Rename `substF` to `substInvF`?

hrhino: 👍. Or `liftF` but I prefer being explicit about the invariance. just `F` makes me think covariant `Functor`.

base/shared/src/main/scala/scalaz/data/Iso.scala
alexknvl: This is mostly for readability of method signatures within those two objects (`Product` and `Coproduct`), I don't intend anyone using `⨂` outside of `Iso`.

TomasMikula: 👍

base/shared/src/main/scala/scalaz/data/Iso.scala
alexknvl: Historical reasons, I'll fix it.

tonymorris:
Can this be generalised to include polymorphic update?

`Iso[S, T, A, B] ~ (S => T, A => B)`

base/shared/src/main/scala/scalaz/data/Iso.scala
```diff
@@ -0,0 +1,149 @@
...
+    final def second[A, B, C](iso: Iso[B, C]): Iso[A ⨂ B, A ⨂ C] =
+      id and iso
+  }
+
+  object Coproduct {
+    private[data] type ⨂[A, B] = \/[A, B]
+    private[data] type Id      = Void
+
+    final def associate[A, B, C]: Iso[A ⨂ (B ⨂ C), (A ⨂ B) ⨂ C] =
+      unsafe(
```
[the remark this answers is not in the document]
fommil: uuuh, we do.... https://gitlab.com/fommil/scalaz-deriving/blob/master/scalaz-deriving/src/main/scala/iotaz/IotaMacros.scala

PR #1702, opened 2018-04-09
Add Show[Array[A]]

NeQuissimus:
Alternative to using `java.util.Arrays.toString`

### DIFF
--- a/core/src/main/scala/scalaz/std/Array.scala
+++ b/core/src/main/scala/scalaz/std/Array.scala
@@ -1,0 +1,10 @@
+package scalaz
+package std
+
+trait ArrayInstances {
+  implicit def arrayShow[A](implicit SA: Show[A]): Show[Array[A]] = new Show[Array[A]] {
+    override def shows(a: Array[A]): String = a.map(SA.shows).mkString("[", ",", "]")
+  }
+}
+
+object array extends ArrayInstances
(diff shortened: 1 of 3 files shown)


core/src/main/scala/scalaz/std/Array.scala
```diff
@@ -0,0 +1,8 @@
+package scalaz
+package std
+
+trait ArrayInstances {
+  implicit def arrayShow[A](implicit SA: Show[A]): Show[Array[A]] = new Show[Array[A]] {
+    override def shows(a: Array[A]): String = a.map(SA.shows).mkString("[", ",", "]")
```
hrhino: The `map` call makes a temp `Array[String]`, which is probably suboptimal. Even using `.iterator` may be faster.

core/src/main/scala/scalaz/std/Array.scala
```diff
@@ -0,0 +1,8 @@
...
+  }
+}
```
hrhino: Add an `object array extends ArrayInstances` for `import scalaz.std.array._`.

xuwei-k:
(rejected) related pull request https://github.com/scalaz/scalaz/pull/448

emilypi:
@NeQuissimus consider a slightly more optimal `shows` implementation (as @hrhino) suggested - take a look at [this](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/foldable.scala#L84)

NeQuissimus:
I can agree with the reasoning in #448

PR #1704, opened 2018-04-09
Better Void.

alexknvl:
https://github.com/scalaz/scalaz/pull/1698#issuecomment-379895959

### DIFF
--- a/base/shared/src/main/scala/scalaz/Prelude.scala
+++ b/base/shared/src/main/scala/scalaz/Prelude.scala
@@ -55,6 +55,7 @@ trait BaseTypeclasses {
 }
 
 trait BaseData {
+  type Void = data.Void.Void
   type Both[A, B]                                      = data.Both[A, B]
   type Forall2[F[_, _]]                                = data.Forall2.Forall2[F]
   type Forall[F[_]]                                    = data.Forall.Forall[F]
@@ -80,13 +81,13 @@ trait BaseData {
 }
 
 trait BaseDataAliases { self: BaseData =>
-  type Void = data.Void.Void
-
   type \/[L, R]    = data.Disjunction.\/[L, R]
   type ===[A, B]   = data.Is[A, B]
   type =!=[A, B]   = data.NotIs[A, B]
   type <~<[-A, +B] = data.As[A, B]
   type >~>[+B, -A] = data.As[A, B]
+
+  val Void: data.Void.type = data.Void
 
   val Forall: data.Forall.type = data.Forall
   val ∀ : data.Forall.type     = data.Forall
@@ -135,7 +136,8 @@ trait AllInstances
     with typeclass.TraversableInstances
 
 trait AllSyntax
-    extends data.AsSyntax
+    extends data.VoidSyntax
+    with data.AsSyntax
     with data.DisjunctionSyntax
     with data.ForallSyntax
     with data.Forall2Syntax
--- a/base/shared/src/main/scala/scalaz/Prelude.scala
+++ b/base/shared/src/main/scala/scalaz/Prelude.scala
@@ -55,7 +55,7 @@ trait BaseTypeclasses {
 }
 
 trait BaseData {
-  type Void = data.Void.Void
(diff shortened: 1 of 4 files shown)


base/shared/src/main/scala/scalaz/Prelude.scala
```diff
@@ -135,7 +136,8 @@ trait AllInstances
     with typeclass.TraversableInstances
 
 trait AllSyntax
-    extends data.AsSyntax
+    extends data.VoidSyntax
```
hrhino: could wind up a few lines down for us alphabetically-inclined folks

base/shared/src/main/scala/scalaz/data/Iso.scala
```diff
@@ -133,11 +131,10 @@ object Iso {
         case -\/(b) => \/-(b)
       })
 
-    @silent
     final def unit[A]: Iso[A, A ⨂ Id] =
       unsafe({ case a => -\/(a) }, {
         case -\/(a)   => a
-        case \/-(n)   => n // Dead code warning.
```
hrhino: 👍

jdegoes:
👍

base/shared/src/main/scala/scalaz/data/Void.scala
```diff
@@ -1,13 +1,45 @@
...
-  type Void <: Nothing with Tag
+trait VoidSyntax {
+  implicit class Ops(v: Void) {
+    def absurd[A]: A = Void.absurd(v)
+  }
+}
+
+// NOTE: this is some next level black compiler magic
+// but without this object syntax doesn't resolve...
+object VoidSyntax extends VoidSyntax
```
hrhino: I'm curious: what's the actual compiler magic?

alexknvl: I am honestly not sure *how* and *why* it works. Took it from @TomasMikula 's `Forall` implementation, maybe he knows :)

TomasMikula: Yeah, it's super confusing...

We are looking for implicits related to `scalaz.data.Void.Void` = `scalaz.data.Void.type#Void`.

The spec [says](https://scala-lang.org/files/archive/spec/2.12/07-implicits.html#implicit-parameters) that the "parts" of this type projection include the parts of `scalaz.data.Void.type`:

 > if T is a type projection S#U, the parts of S as well as T itself;

Then the spec says that the parts of this singleton type `scalaz.data.Void.type` are the parts of the type of `scalaz.data.Void`:

> if T is a singleton type p.type, the parts of the type of p;

The type of `scalaz.data.Void` is `VoidModule with VoidSyntax`.

The parts of `VoidModule with VoidSyntax` is just `VoidModule with VoidSyntax` itself:

> in all other cases, just T itself.

From that we conclude that `VoidModule with VoidSyntax` is a "part" of the type `scalaz.data.Void.Void`.

The class `VoidSyntax` is a base class of `VoidModule with VoidSyntax`, which is a part of `scalaz.data.Void.Void`, thus `VoidSyntax` is "associated" with `scalaz.data.Void.Void`:

> Here, we say a class C is associated with a type T if it is a base class of some part of T.

Finally,

> The implicit scope of a type T consists of all companion modules of classes that are associated with the implicit parameter's type.

which means that the companion of  `VoidSyntax` is included in the search.

hrhino: Ah! that makes sense. I knew about the prefix rules but I didn't notice that `VoidModule` extended `VoidSyntax`.

TomasMikula: It doesn't :). But we could have

```scala
object VoidModule extends VoidSyntax
```

instead of

```scala
object VoidSyntax extends VoidSyntax
```

and then

```scala
val Void: VoidModule = VoidImpl
```

instead of

```scala
val Void: VoidModule with VoidSyntax = VoidImpl
```

I guess that'd be slightly better.

PR #1708, opened 2018-04-10
WIP: Make IO a bifunctor

jdegoes:
Initial WIP for transforming `IO` into a `Bifunctor`, with a precisely-specified error type. This allows a precise, compile-time specification of _if_ and _how_ an `IO` computation may fail.

Many methods have been generalized. The remaining work involves more precisely specifying error types—for example, ensuring that effects which cannot fail are polymorphic in the error type—as well as updating documentation to reflect the new error model.


effect/shared/src/main/scala/scalaz/effect/IORef.scala
```diff
@@ -15,28 +18,104 @@ package scalaz.effect
...
 
   /**
-   * Writes a new value to the `IORef`.
+   * Atomically modifies the `IORef` with the specified function. This is not
+   * implemented in terms of `modifyFold` purely for performance reasons.
    */
-  final def write(a: A): IO[Unit] = IO.sync(this.value = a)
+  final def modify[E](f: A => A): IO[E, A] = IO.sync {
+    var loop = true
+    var next : A = null.asInstanceOf[A]
```
emilypi: 😱

effect/shared/src/main/scala/scalaz/effect/IORef.scala
```diff
@@ -15,28 +18,104 @@ package scalaz.effect
...
+  }
 }
 
 object IORef {
 
   /**
    * Creates a new `IORef` with the specified value.
    */
-  final def apply[A](a: A): IO[IORef[A]] = IO.sync(new IORef[A](a))
+  final def apply[E, A](a: A): IO[E, IORef[A]] = IO.sync(new IORef[A](new AtomicReference(a)))
```
emilypi: @jmcardon 😄 you were spot on!

effect/shared/src/main/scala/scalaz/effect/AsyncReturn.scala
```diff
@@ -9,22 +9,14 @@ package scalaz.effect
...
-}
+sealed abstract class AsyncReturn[E, A]
 object AsyncReturn {
-  type Canceler = Throwable => Unit
+  type Interruptor = Throwable => Unit
 
-  private final case object Later                    extends AsyncReturn[Nothing]
-  final case class Now[A](value: A)                  extends AsyncReturn[A]
-  final case class MaybeLater[A](canceler: Canceler) extends AsyncReturn[A]
+  private final case object Later extends AsyncReturn[Nothing, Nothing]
```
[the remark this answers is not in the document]
jdegoes: There is also @TomasMikula's way of encoding these polymorphic values, which I have meant to deploy ever since I learned about it (see also: `IO.never`, `IO.unit`, and several other values which could benefit from this transformation).

I'll be sure I clean all this stuff up in a subsequent commit!

effect/shared/src/main/scala/scalaz/effect/IORef.scala
jdegoes: Hehe. The other alternative is to `@tailcall` it and inspect the generated bytecode.

Unfortunately any unnecessary boxing in `IORef` will significantly decrease performance given how low-level the construct is...

effect/shared/src/main/scala/scalaz/effect/IORef.scala
```diff
@@ -15,28 +18,104 @@ package scalaz.effect
  * } yield ()
  * }}}
  */
-final class IORef[A] private (@volatile private var value: A) {
-
+final class IORef[A] private (private val value : AtomicReference[A]) extends AnyVal {
```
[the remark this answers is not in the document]
jdegoes: Superb! Great minds and all that. 😄 

The main benefit to `AtomicReference` is that a concurrent-safe version of `modify` can be implemented (and has been, here), which means you can use `IORef` as a single-variable STM.

Surprisingly there's also a precedent for this functionality in Haskell's `IORef`, although I can't recall exactly what it is (some atomic compare-and-swap variant).

jdegoes:
@jmcardon 

> That is, when working with IO[E1, A] and IO[E, B] where E1 <: E, attempting to compose these two won't work out of the box without a cast since IO is invariant, despite it being correct.

This makes sense for performance reasons (of course you can `leftMap` to widen). I wonder if we should use @alexknvl's `Liskov` or actually use subtyping. Probably the latter would make for an easier-to-use API.

effect/shared/src/main/scala/scalaz/effect/IORef.scala
emilypi: I hear you and trust the motivation 😄

fommil:
how about renaming this IO to GIO (for General) and then `IO[a] = GIO[Throwable, a]`

It's going to be a pain to migrate codebases that are using IO already for dealing with legacy java apis.

jdegoes:
@fommil 

screenshot_945

effect/README.md
```diff
@@ -39,7 +39,7 @@ import scalaz.effect.{IO, SafeApp}
 import scalaz.effect.console._
 
 object MyApp extends SafeApp {
-  def run(args: IList[String]): IO[Unit] =
+  def run(args: IList[String]): IO[E, Unit] =
```
Blaisorblade: Where's type variable `E` bound? Or is it a top-level type named `E`? Or a _meta_variable `E` which isn't introduced yet? Your examples seem to use `Error` instead.

effect/README.md
```diff
@@ -53,15 +53,15 @@ object MyApp extends SafeApp {
 You can lift pure values into `IO` with `IO.point`:
 
 ```scala
-val liftedString: IO[String] = IO.point("Hello World")
+val liftedString: IO[E, String] = IO.point("Hello World")
```
Blaisorblade: here too whence `E`

effect/README.md
```diff
@@ -117,12 +117,12 @@ val contacts: IO[IList[Contact]] =
 You can create `IO` actions that describe failure with `IO.fail`:
 
 ```scala
-val io: IO[String] = IO.fail(new Error("Oh noes!"))
+val io: IO[E, String] = IO.fail(new Error("Oh noes!"))
```
Blaisorblade: all introduced `E` until here seem unbound

effect/shared/src/main/scala/scalaz/effect/AsyncReturn.scala
```diff
@@ -17,8 +17,9 @@ object AsyncReturn {
   object Later {
     // Trick thanks to Sam Halliday. We eliminate allocation overhead but also
```
fommil: it's really @xuwei-k's trick... mine was cleaner but broke exhaustivity checking. And @hrhino came up with the fix for the dead code thing.

BTW, can we please make @hrhino a member of the org so he can approve PRs?

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -1,16 +1,20 @@
...
 import scalaz.data.Maybe
+import scalaz.Void
+
+import scalaz.effect.Errors._
 
 /**
- * An `IO[A]` ("Eye-Oh of A") is an immutable data structure that describes an
- * effectful action that may throw an exception (`Throwable`), run forever, or
- * produce a single `A` at some point in the future.
+ * An `IO[E, A]` ("Eye-Oh of A") is an immutable data structure that describes an
```
Blaisorblade: Still read `"Eye-Oh of A"`?

hrhino: "Eye-Oh Ee of Aye", you.

fommil: and bingo was his name o.

effect/shared/src/main/scala/scalaz/effect/AsyncReturn.scala
hrhino: I think I am a member? so I'm confused.

fommil: @hrhino oh, when I see your reviews they are often grayed out... whereas others come through with a green tick.

effect/README.md
```diff
@@ -131,10 +131,10 @@ val result = openFile("data.json").attempt.map {
 }
 ```
 
-You can submerge failures with `IO.absolve`, which turns an `IO[Throwable \/ A]` into an `IO[A]`:
+You can submerge failures with `IO.absolve`, which turns an `IO[Void, E \/ A]` into an `IO[E, A]`:
 
 ```scala
-def sqrt(io: IO[Double]): IO[Double] =
+def sqrt(io: IO[E, Double]): IO[E, Double] =
   IO.absolve(
```
Blaisorblade: If absolve only accepts `IO[Void, E \/ A]`, how does this code typecheck?

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -1,16 +1,20 @@
...
+ * effectful action that may fail with an `E`, run forever, or produce a single
+ * `A` at some point in the future.
```
Blaisorblade: Looking forward to more documentation on if/when this guarantee fails. Quoting what I asked on [Gitter](https://gitter.im/scalaz/scalaz?at=5ace33ff270d7d3708b80e16):

> does anybody plan to state "formally" under what conditions you get other exceptions? Not sure Scalazzi is enough
> it seems E behaves basically like a phantom type parameter at the edges
> 1. so you need to import Java functions with the correct spec
> 2. you have "async. exceptions" from the runtime, you need to consider them
> 3. if you listen to Java checked exception specs, you need to consider unchecked exceptions as "async"

effect/README.md
hepin1989: I do like the `Done` one.

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -161,6 +161,12 @@ sealed abstract class IO[E, A] { self =>
       case \/-(a) => IO.now[E2, A](a)
     }
 
+  /**
+   * Widens the error type to any supertype. While `leftMap` suffices for this
+   * purpose, this method is significantly faster for this purpose.
+   */
+  final def widen[E2 >: E]: IO[E2, A] = self.asInstanceOf[IO[E2, A]]
```
fommil: no Liskov?

jdegoes: If I use `<~<`, then the user will have to construct it explicitly, as these things are not automatically summoned into scope for subtyping relationships. Or am I missing something, @alexknvl?

alexknvl: `implicit` will work in Scalaz8

jdegoes: @alexknvl Unfortunately, I cannot use it, since the method I require (`substCv`) requires variance in the type parameter `E`.

Well, I can require it as an implicit and ignore it, performing the coercion, but then scalac will fail due to an unused implicit. 😄

alexknvl: You can un-ignore something using
```scala
val _ = ev
```
But regardless, I don't feel particularly strongly about this :)

jdegoes: I could go either way, too. I do feel like we need a general solution to this problem since many other classes in Scalaz 8 are invariant.

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -502,28 +603,44 @@ object IO extends IOInstances {
...
-  final def suspend[A](io: => IO[A]): IO[A] = Suspend(() => io)
+  final def suspend[E, A](io: => IO[E, A]): IO[E, A] = Suspend(() => io)
 
   /**
    * Imports a synchronous effect into a pure `IO` value. If the thunk returns
    * a null value, this will throw a `NullPointerException` inside `IO`.
    *
    * {{{
-   * def putStrLn(line: String): IO[Unit] = IO.sync(println(line))
+   * def putStrLn(line: String): IO[Throwable, Unit] = IO.partialSync(println(line))
```
pjrt: This scaladoc is for `sync`, yet it mentions `partialSync`.

effect/shared/src/main/scala/scalaz/effect/IO.scala
alexknvl: Is there any way to create `IsCovariant[IO]` instance?

effect/shared/src/main/scala/scalaz/effect/IO.scala
```diff
@@ -39,31 +43,31 @@ import scalaz.data.Maybe
  * values, see the default interpreter in `RTS` or the safe main function in
  * `SafeApp`.
  */
-sealed abstract class IO[A] { self =>
+sealed abstract class IO[E, A] { self =>
```
pjrt: Not really sure what it is that we expect `E` to be. From my point of view, `E` can't really be anything but some `E <: Exception`. If we are considering `E` to be any error type, like `String` for example, then how is that different than `EitherT[IO, E, A]`? (where `IO` is the current IO). 

Now, one cool thing that can happen from exposing this `E` is that we can now have `IO[Exception with NoStackTrace, A]`. This will allow your code to use exceptions without the stacktrace issue (and even control it). Not sure how impure code will convert from their Exception/Throwable into a NoStrackTrace one, but maybe there's way.

jdegoes:
@Blaisorblade All right, updated the docs (mostly) to match the current state of the code. Previously, they were result of search & replace (`IO[A]` => `IO[E, A]`) which is why they didn't make sense. Let me know if you spot anything!

effect/shared/src/main/scala/scalaz/effect/IO.scala
jdegoes: There is no requirement on `E`. It can be some ADT, unrelated to exceptions, or even just `String` if all you care about is error messages; `Void` / `Nothing` if you don't want any errors; or `Unit` if you want to treat an IO as `F[Option[A]]`. 

You can `leftMap` over `E` to "lift" an IO from one error type to another, which lets you change your error types in different parts of your application.

effect/shared/src/main/scala/scalaz/effect/IO.scala
jdegoes: Fixed, thanks!

effect/shared/src/main/scala/scalaz/effect/AsyncReturn.scala
jdegoes: Yep he's a member already.

effect/shared/src/main/scala/scalaz/effect/IO.scala
pjrt: Then I'm confused: what's the difference between `IO[E, A]` and `EitherT[IO, E, A]`?

effect/shared/src/main/scala/scalaz/effect/IO.scala
TomasMikula: You could have `IsCovariant` for `IO` without coercions if `IO` was an opaque type implemented by a covariant `IOImpl[+E, +A]`. 😃 (That's how `IsCovariant` is done for `Maybe`, because it is just an alias for the covariant `Option`.)

effect/shared/src/main/scala/scalaz/effect/IO.scala
jbgi: > what's the difference between IO[E, A] and EitherT[IO, E, A]

considering an `IO[A]` that never catch exceptions, there are isomorphic (I think), the difference is better performance and ease of use (no/less monad transformers).

> and how does E interface with flat out thrown exceptions?

See [`partialSnyc`](https://github.com/jdegoes/scalaz/blob/a05f9f1f0ecde46159b598d92f19f9de5e56d22c/effect/shared/src/main/scala/scalaz/effect/IO.scala#L638). The strategy here is to deffer constraints on type parameters until you actually need them, to maximize possible usages. It is similar to how one does not need an `Order[A]` to construct a `ISet.empty[A]`.

jdegoes: @pjrt 

The type `EitherT[IO, E, A]` — for Scalaz 7, Monix Task, Cats Effect, Future, etc. — has two separate error channels, and two `MonadError` instances (one is fixed to `Throwable`, and one varies as `E`).

You can and probably should think of `EitherT[IO, E, A]` for these effect monads as being equivalent to `EitherT[EitherT[Unexceptional, Throwable, ?], E, A]` (for some unexceptional effect monad `Unexceptional`), which is a nesting of 2 layers of `Either`. Such values may "fail" — in the functional programming sense — and be recovered from in two distinct ways: the inner `Either` (with `Throwable`), or the outer `Either` (with `E`).

In contrast, `IO[E, A]` has a single `MonadError` instance, and should be thought of as equivalent to `EitherT[Unexceptional, E, A]`. There is only one way it may "fail" — in the functional programming sense — and be recovered from: with `E`.

With the older effect monads, because there are two separate error channels, and because one of them is fixed at compile-time to `Throwable`, it limits the precision with which you can describe effects. You cannot describe effects that cannot fail, nor can you describe effects that can fail in only one of `n` distinct ways, because the bundling of the inner `Either` with its `Left` case fixed to `Throwable` prevents such descriptions.

Now, another way to formulate the Scalaz 8 IO bifunctor is as `EitherT[Unexceptional, E, A]`. Arguably, this is a better design, practical considerations aside. However, monad transformers are a terrible fit for Scala, and can easily destroy performance and limit applications of effect monads. Instead of taking this approach, Scalaz 8 IO bakes in the `EitherT` so it has no runtime overhead.

pjrt: > considering an IO monad that never catch exceptions

Is such an IO even useful? IO is meant for those kinds of things.

> the difference is better performance and ease of use (no/less monad transformers).

Then it is just another form of `EitherT[IO, E, A]`? That's a bit confusing, does EitherT have any meaning at that point? Should it be removed?

> See partialSnyc. The strategy here is to deffer constraints on type parameters until you actually need them, to maximize possible usages. It is similar to how one does not need an Order[A] to construct a ISet.empty[A].

I get that, but the idea of turning `Throwable => E` sounds like "My error ADT `E` will always need a `SomeThrowable(t: Throwable)` case". Which *feel* wrong.

pjrt: @jdegoes 

> However, monad transformers are a terrible fit for Scala, and can easily destroy performance and limit applications of effect monads. Instead of taking this approach, Scalaz 8 IO bakes in the EitherT so it has no runtime overhead.

That's actually fair (better performance), even though it makes it bit confusing. Maybe make a mention about EitherT on the scaladoc? I can see other people becoming confused about its relationship with EitherT.

pjrt:
I'll second @fommil 's concern. Calling this "generic" IO `IO` will not only require somewhat large refactors on projects that are using the current IO, but it will also confuse people who are used to the original IO definition.

jdegoes:
@pjrt It is well known that almost no one uses Scalaz 7 `IO`. Almost all Scalaz projects use `Task` exclusively. I don't think this will be a major issue in practice. 

Also in order to cope with the name and semantic changes for many methods, it will be necessary to deploy a ScalaFix rewrite... or at least come up with a facade. Either of which help address the migration problem.

jmcardon:
@pjrt there are _no_ projects using the current IO because no one has published it yet.

fommil:
I was thinking more about refactoring cats IO to this ;-)

Doing a global rename is trivial with scalafix. So I'm less concerned. I think I'd rather be typing IO than Task, though, to be honest.

pjrt:
Opps, sorry, lost the context there. I was also thinking about cats-effects IO. My main concern is still people looking for IO and being introduced to `IO[E,A]`, not what I would expect. But the scaladoc was updated to explain it, so that should be good.

effect/shared/src/main/scala/scalaz/effect/IO.scala
pjrt: @jbgi Ah, I see what you mean now.

Milyardo:
How about the name`Compute` instead of `Unexceptional`?

fommil:
Hmm, I really hate to keep harping on about the name of things (it's such a waste of time to bikeshed about names), but since we are in general consensus that scalaz8 will use Haskell names where there is precedent (even if we think the haskell name is bad), there is definitely precedent for `IO` to have a single type parameter, so I think this `IO` should be renamed to `GIO` or something and `type IO[a] = GIO[Throwable, a]`, perhaps also `type Task[a] = GIO[Throwable, a]` for people using the scalaz7 `Task`.

I shall yield my objection if there is a very strong desire for `IO` to have a bifunctor, but I reserve my right to complain forever about having to type 4 characters instead of 2 everywhere I am interacting with legacy code :smile:

build.sbt
```diff
@@ -36,8 +36,8 @@ lazy val benchmarks = project.module
         "org.scala-lang" % "scala-reflect"  % scalaVersion.value,
         "org.scala-lang" % "scala-compiler" % scalaVersion.value % "provided",
         "org.scalaz"     %% "scalaz-core"   % "7.2.7",
-        "io.monix"       %% "monix"         % "3.0.0-M1",
-        "org.typelevel"  %% "cats-effect"   % "0.4"
+        "io.monix"       %% "monix"         % "3.0.0-M3",
```
barambani: this is currently at 3.0.0-RC1 I think

pjrt:
I'm with @fommil on this. I understand the desire to make it a bifunctor, but the fact that it is now more akin to EitherT makes me think it should have a different name.

jdegoes:
Scalaz has a strong historical history of (a) improving upon the mistakes of Haskell, (b) incentivizing users to do the correct thing, even when unfamiliar or inconvenient.

Personally, I feel the default in any new code base should always be `IO[E, A]`, never `IO[A]`. The latter only has a place in legacy code in the process (perhaps indefinite) of being migrated. 

The old `IO` and `Task` were already "broken bifunctors" in the sense they bundled the `Either` and fixed the left type to `Throwable`. This defect has been rectified in this design. This is exactly `IO`, but without fixing the left type parameter to `Throwable`.

So my preference would be to keep the name `IO`, because this is _the construct_ that you should use for effects in new code bases, _not_ a type synonym such as `Task`.  Simultaneously, we could introduce a new 2-letter synonym for `IO[Throwable, A]`, such as `Ef[A]` or `OI[A]`, to ease typing.

If it feels strange, it's only because it's new. People complained about fork / join, fibers, and interruption, and now even cats effect has simple versions. This is the shape of all future effect monads in Scala, it's just going to take a little while (and some benchmarking) for the others to catch up.

However, I don't have a terribly strong preference (names are no hill for me to die on!). If we have a vote and most contributors think we should change the name, I'm OK with that.

jdegoes:
@Milyardo The problem with that is `IO[Void, A]` can represent actual effectful code, just not code that fails. Think of things like `new AtomicInteger()` or `System.nanoTime()`, which cannot fail, but are still effectful. It's not "pure compute".

Milyardo:
`Compute` doesn't imply purity to me, but I can see how it does to some. To say something is computable says that a computation eventually terminates much more than than it says it can't fail. So I guess the name is a bit off still.

emilypi:
BIG MERGE NICE JOB EVERYBODY

jdegoes:
👯‍♂️ 🎉

jmcardon:
[IT'S MERGED WOO](https://www.youtube.com/watch?v=pD_imYhNoQ4)

PR #1709, opened 2018-04-10
Eq typeclass.

alexknvl:


example/src/main/tut/typeclass/Eq.md
```diff
@@ -0,0 +1,42 @@
...
+*The `Eq` type class describes a type which can be compared for observational equality.*
+
+A `Eq` must satisfy the following [laws](https://en.wikipedia.org/wiki/Identity_of_indiscernibles):
+
+- Reflexivity:
+  - `x === x` for any `x`.
+- The indiscernibility of identicals:
+  - For any `x` and `y`, if `x` is identical to `y`, then `x` and `y` have all the same properties, or in other words there is no function `f: A => Boolean` that returns `true` for `x` and `false` for `y`.
+- The identity of indiscernibles:
+  - For any `x` and `y`, if `x` and `y` have all the same properties, then `x` is identical to `y`.
```
hrhino: I would simplify this to 

```markdown
- The indiscernibility of identicals:
  - For any `x` and `y`, if `x === y`, then `x` and `y` are indiscernable
- The identity of indiscernibles:
  - For any `x` and `y`, if `x` and `y` are indiscernable, then `x === y`
```

and then mention something like 

```markdown
*Indiscernability* formalizes the intuitive notion of two objects having the exact same properties. Two values `x, y: A` are indiscernable if there exists no function `f: A => Boolean` such that `f(x)` is `true` and `f(y)` is `false`.
```

jdegoes:
YAY!!!!!!!

PR #1711, opened 2018-04-11
Symbolic ops macros.

alexknvl:
https://github.com/scalaz/scalaz/pull/1411 served as inspiration.
See https://github.com/scalaz/scalaz/pull/1711/files#diff-726e672735d618c9d9bc069c449a4fbeL6 for an example.

### DIFF
--- a/base/shared/src/main/scala/scalaz/typeclass/EqSyntax.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/EqSyntax.scala
@@ -5,7 +5,7 @@ import scala.language.experimental.macros
 
 trait EqSyntax {
   implicit final class ToEqOps[A: Eq](a: A) {
-    private[typeclass] type Equal
-    def ===(f: A): Boolean = macro meta.SymOps.fa_1[Equal]
+    private[typeclass] type equal
+    def ===(f: A): Boolean = macro meta.SymOps.fa_1[equal]
   }
 }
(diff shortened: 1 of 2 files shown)


base/shared/src/main/scala/scalaz/typeclass/EqSyntax.scala
```diff
@@ -1,9 +1,11 @@
...
 package typeclass
 
+import scala.language.experimental.macros
+
 trait EqSyntax {
   implicit final class ToEqOps[A: Eq](a: A) {
-    // TODO: macro syntax
-    def ===(b: A): Boolean = implicitly[Eq[A]].equal(a, b)
+    private[typeclass] type Equal
+    def ===(f: A): Boolean = macro meta.SymOps.fa_1[Equal]
```
hrhino: why uncap it? `def ===(f: A): Boolean = macro meta.SymOps.fa_1[equal]` seems fine to me.

I wish we had literal types in scala 2; `macro meta.SymOps.fa_1["equal"]` would be better still.

alexknvl: Fixed.

hrhino: Thanks!

alexknvl:
All credit due to @edmundnoble

jdegoes:
@alexknvl @edmundnoble This looks awesome! However, it would be best to push the constraint to the method for better compiler errors (&, in theory, performance).

alexknvl:
I think it is out of scope for this PR, but otherwise I agree with the compiler error part. Performance should be the same because of the macro.

edmundnoble:
This is so goddamn good.

PR #1714, opened 2018-04-12
Reworked syntax macros, changed most syntax ops to use method evidence.

alexknvl:


jdegoes:
@alexknvl Awesome work. Even better that it's net negative. 😄

TomasMikula:
Deferring the requirement of a typeclass instance will cause ambiguity when there are two unrelated typeclasses (`Foo` and `Bar` below) with an operation of the same name (`go` below). The following works, but would break if the typeclass requirement was deferred to the `go` methods.

```scala
trait Foo[A] {
  def go(a: A): Int
}

object Foo {
  def apply[A](implicit ev: Foo[A]): Foo[A] = ev

  implicit def stringFoo: Foo[String] = new Foo[String] {
    def go(a: String) = a.length
  }

  implicit final class FooOps[A: Foo](self: A) {
    def go: Int = Foo[A].go(self)
  }
}

trait Bar[A] {
  def go(a: A): Int
}

object Bar {
  def apply[A](implicit ev: Bar[A]): Bar[A] = ev

  implicit def intBar: Bar[Int] = new Bar[Int] {
    def go(a: Int) = a
  }

  implicit final class BarOps[A: Bar](self: A) {
    def go: Int = Bar[A].go(self)
  }
}

object Main {
  import Foo._
  import Bar._

  42.go       // Bar[Int].go(42)
  "hello".go  // Foo[String].go("hello")
}
```

We might keep the op names unique within scalaz, but do we want to pollute the global namespace for everyone?

alexknvl:
I noticed this problem when changing `Profunctor` and `Bifunctor`, they both have `lmap` and `rmap`, so syntax stopped resolving.

The original motivation was https://scalafiddle.io/sf/jHlufJl/0 - the first version with `implicit` on the method gives a more informative message.

TomasMikula:
🤔 Yeah, I admit the error message with deferred typeclass instance is much more useful. But maybe the error message in the other case could be improved in scalac.

TomasMikula:
There's an open ticket for it, https://github.com/scala/bug/issues/7906, but it's been open for 4.5 years. 😕

TomasMikula:
I don't want to block this. If others agree that the trade-off (i.e. helpful error messages at the cost of occasional conflicting method names) is worth it, go ahead an merge. We can reconsider if/when the linked scalac issue is resolved.

jdegoes:
I think I am in favor of a merge. Also we can make method names unique to help resolve this problem.

jdegoes:
@hrhino 👍 I badly want the ability to have a namespace with all TLFs, without conflicts.

hrhino:
@alexknvl [AOL voice:] you've got conflicts!

alexknvl:
hear hear

PR #1718, opened 2018-04-15
VoidModule should extend VoidSyntax

### LINKED ISSUE #1707: A minor improvement to the newtype encoding.
https://github.com/scalaz/scalaz/pull/1704#discussion_r180409646

nicflores:
Fixes #1707.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/package.scala
+++ b/base/shared/src/main/scala/scalaz/data/package.scala
@@ -1,7 +1,7 @@
 package scalaz
 
 package object data {
-  val Void: VoidModule with VoidSyntax = VoidImpl
+  val Void: VoidModule = VoidImpl
 
   val Forall: ForallModule with ForallSyntax = ForallImpl
   val ∀ : Forall.type                        = Forall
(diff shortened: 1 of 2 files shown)


base/shared/src/main/scala/scalaz/data/package.scala
```diff
@@ -1,7 +1,7 @@
 package scalaz
 
 package object data {
-  val Void: VoidModule with VoidSyntax = VoidImpl
+  val Void: VoidModule = VoidImpl
 
   val Forall: ForallModule with ForallSyntax = ForallImpl
```
alexknvl: There is also `Forall` and `Forall2`, which use the same encoding.

nicflores: @alexknvl,
Let me know if the latest commit is what you had in mind.

TomasMikula: We don't need ` with ForallSyntax` here anymore.

jdegoes:
@nicflores Nice to see this. 😄

Issue #1662, opened 2018-03-04
fromIso everywhere


fommil:
In https://github.com/scalaz/scalaz/pull/1661 I introduced a method on MonadError called `fromIso` which allowed deriving a `MonadError` from something isomorphic that has one already.

Most things already have an `IsomorphismThing` trait in `Isomorphism.scala` but very few people know about it, and using it is tedious.

We should add `fromIso` methods onto all typeclass companions that has an isomorphism.

The convention should be:

1. `Iso` parameter is not implicit
2. underlying / delegate instance is implicit

since `Iso` is not technically a typeclass.

fommil:
it would also be good if there was Iso syntax. It would need to consider isos that are either way around and flip the opposite one. So maybe two implicit classes, to avoid overloading.

This would remove the need for `.toList` and `.fromList` style methods.

xuwei-k:
> Iso is not technically a typeclass.

create new typeclass ? https://github.com/xuwei-k/scalaz/commit/8076ed999b4d99dcc0ca30f70562f4450d5260d3

fommil:
I guess that works :smile:

fommil:
I'm increasingly convinced that `fromIso` is how to derive typeclasses of shape `F[_[_]]` such as `Monad`. see https://gitlab.com/fommil/scalaz-deriving/issues/74

The `fromIso` convention is easily supported, but obtaining the `IsoFunctor` for the data type could be difficult unless we encourage them to be implicit or we derive them with a macro (I have one already).

I also think I need a refinement of `Iso` to include string names of the fields, but that's another conversation.

xuwei-k:
https://github.com/scalaz/scalaz/pull/1842
https://github.com/scalaz/scalaz/pull/1847

fommil:
:-D

PR #1720, opened 2018-04-16
derivation for isomorphic things (#1662)

valydia:


zainab-ali:
Thanks for opening this @valydia!  

@fommil this PR is part of [syntactic sugar's hack evening on scalaz-deriving related issues](https://github.com/syntactic-sugar-london/project-suggestions/blob/master/2018-04-16/scalaz.md).  Is it what you expected?

core/src/main/scala/scalaz/Enum.scala
```diff
@@ -251,5 +251,13 @@ object Enum {
     }
     z
   }
+
+  import Isomorphism.<=>
+
+  def from[F, G](D: F <=> G)(implicit E: Enum[G]): Enum[F] =
```
zainab-ali: This should probably be called `fromIso` so it's the same as the other ones

zainab-ali:
It would be cool to have tests for these, in line with the `MonadErrorTest`.  The isomorphism between `Decoder` and `Kleisli[String \/ ?, String, A]` in the `MonadErrorTest` was simple, yet relevant.

I'll have a think on the isomorphisms we could use for these typeclasses.

fommil:
Oh wow, thanks! I'll try to take a look over the coming days.

core/src/main/scala/scalaz/Enum.scala
valydia: Sorry, that's a typo from me, well spotted! Thanks

zainab-ali:
Looks like the build is erroring with 

```
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install oracle-java8-installer" failed and exited with 100
```

Which is just odd.  It clearly has no relation to the code change.

core/src/main/scala/scalaz/Isomorphism.scala
```diff
@@ -235,7 +235,7 @@ trait IsomorphismOrder[F, G] extends Order[F] {
   override def order(x: F, y: F): Ordering = G.order(iso.to(x), iso.to(y))
 }
 
-trait IsomorphismEnum[F, G] extends Enum[F] {
+trait IsomorphismEnum[F, G] extends Enum[F] with IsomorphismOrder[F, G] {
```
fommil: this probably breaks bincompat

valydia: I've added because the rules seems to be if  typeclass B is a subtype a typeclass A, then typeclass IsomorphismB would mix in IsomorphismA, for instance `IsomorphismMonoid` mixes in `IsomorphismSemigroup`. I suspected it that it could break the bincompat, I'll remove it and I'll get `IsomorphismEnum` to overrides the `order` function as well.

valydia:
There are much more typeclasses  but that was all I could cover in that hacking session. I can try to finish it off in the next session or at a later date 😄.

zainab-ali:
I think it would be good to merge this in, even if there are some typeclasses we can add later (I want to add some laws!).  Do we need further review by @xuwei-k ?

fommil:
I was waiting for Kenji but I'm fine to merge since it passes CI and is only adding things.

PR #1727, opened 2018-04-17
Change deep map to synchronous deep bind tests

jmcardon:
- The deep map tests currently are simply wrong, and it's supposed to stackoverflow now. Encouraging deep map means you hate all that is good in the world. 
- We have no test for deep synchronous binds.

### DIFF
--- a/effect/jvm/src/test/scala/scalaz/effect/RTSTest.scala
+++ b/effect/jvm/src/test/scala/scalaz/effect/RTSTest.scala
@@ -52,9 +52,9 @@ class RTSSpec(implicit ee: ExecutionEnv) extends Specification with AroundTimeou
     test eval of async fail                 $testEvalOfAsyncAttemptOfFail
 
   RTS synchronous stack safety
-    deep map of point                       $testDeepMapOfPoint
-    deep map of now                         $testDeepMapOfNow
-    deep map of sync effect                 $testDeepMapOfSyncEffectIsStackSafe
+    deep bind of point                      $testDeepBindOfPoint
+    deep bind of now                        $testDeepBindOfNow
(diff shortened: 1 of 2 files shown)


TomasMikula:
What is a deep map, anyway? If you mean `fa.map(f1).map(f2) ... .map(fN)`, why is it bad?

effect/jvm/src/test/scala/scalaz/effect/RTSTest.scala
```diff
@@ -326,14 +326,14 @@ class RTSSpec(implicit ee: ExecutionEnv) extends Specification with AroundTimeou
     if (n <= 0) 0
     else n + sum(n - 1)
 
-  def deepMapPoint(n: Int): IO[Throwable, Int] =
-    if (n <= 0) IO.point(n) else IO.point(n - 1).map(_ + 1)
+  def deepMapPoint(n: Int, ctr: Int): IO[Throwable, Int] =
```
TomasMikula: Should this be renamed to `deepBindPoint`?

jdegoes:
By functor laws, `fa.map(f1).map(f2) ... .map(fN) == fa.map(fa.andThen(f2).andThen...(fN))`. Since the latter stack overflows for some `N`, we can conclude the former should stack overflow, as well. 😆 😈 

I'm in favor of this pull request. Although I would like to know what `N` is and the test could be changed into that. We do need deep bind tests for synchronous code.

effect/jvm/src/test/scala/scalaz/effect/RTSTest.scala
jmcardon: Fair. Will do!

PR #1730, opened 2018-04-19
[scalaz 7.2.x] support Scala 2.13.0-M4-pre-20d3c21

xuwei-k:


core/src/main/scala/scalaz/ImmutableArray.scala
```diff
@@ -31,7 +30,7 @@ sealed abstract class ImmutableArray[+A] {
   def ++[B >: A: ClassTag](other: ImmutableArray[B]): ImmutableArray[B]
 }
 
-sealed abstract class ImmutableArrayInstances {
```
szeiger: I was wondering what the instances for `ImmutableArray` are for until I realized that scalaz has its own `ImmutableArray` type. Ours might still be renamed to it would resolve the ambiguity but may create another one (because that's why we want to remove it in the first place). See https://github.com/scala/bug/issues/10836.

project/build.scala
```diff
@@ -286,7 +311,7 @@ object build {
     resolvers += Resolver.sonatypeRepo("releases"),
     kindProjectorVersion := "0.9.6",
     libraryDependencies += compilerPlugin("org.spire-math" % "kind-projector" % kindProjectorVersion.value cross CrossVersion.binary)
-  ) ++ osgiSettings ++ Seq[Sett](
+  ).flatMap(_.settings) ++ osgiSettings ++ Seq[Sett](
```
dwijnand: jfyi there's a `Def.settings` utility you can use too

tests/jvm/src/test/scala/scalaz/std/java/EnumTest.scala
```diff
@@ -33,11 +33,11 @@ object EnumTest extends SpecLite {
   }
 
   "pred" in {
-    TimeUnit.values.sliding(2) forall { case Array(tu1, tu2) => tu1 === tu2.pred }
+    TimeUnit.values.sliding(2).map(_.toList) forall { case List(tu1, tu2) => tu1 === tu2.pred }
```
dwijnand: does `case Seq(` not work?

xuwei-k: `scala.Array#sliding` return 
- `Iterator[Array[Int]]` in Scala 2.12
- `Iterator[scala.collection.mutable.WrappedArray[Int]]` in Scala 2.13.0-M4-pre-20d3c21

```
Welcome to Scala 2.12.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_172).
Type in expressions for evaluation. Or try :help.

scala> Array(1, 2, 3, 4).sliding(2).map{ case Seq(a, b) => }
<console>:12: error: scrutinee is incompatible with pattern type;
 found   : Seq[A]
 required: Array[Int]
       Array(1, 2, 3, 4).sliding(2).map{ case Seq(a, b) => }
                                                 ^
```

dwijnand: Sad that you can't patmat a WrappedArray with Seq..

core/src/main/scala-2.13+/scalaz/std/Map.scala
```diff
@@ -0,0 +1,190 @@
...
+    if(m1 contains k) m1 + ((k, f(m1(k), v))) else m1 + ((k, v))
+
+  /** Grab a value out of Map if it's present. Otherwise evaluate
+    * a value to be placed at that key in the Map.
+    */
+  final def getOrAdd[F[_], K, A](m: Map[K, A], k: K)(fa: => F[A])(implicit F: Applicative[F]): F[(Map[K, A], A)] =
+    (m get k).fold(F.map(fa)(a => (m + ((k, a)), a)))(a => F.point((m, a)))
+}
+
+object map extends MapInstances with MapFunctions
```
julienrf: Out of curiosity, what was incompatible with the new collections in the original `Map.scala` file?

julienrf: OK so now the second question is: it looks like the new version should work with both 2.12 and 2.13, no? Or maybe some features are missing from it?

tests/jvm/src/test/scala/scalaz/std/java/EnumTest.scala
szeiger: You can still match it as a `scala.collection.Seq` but not as a `scala.Seq` anymore because it's a mutable collection.

dwijnand: Oh yeah, cool 😄

PR #1734, opened 2018-04-25
Replace all usages of "=:=" with "==="

JordanMartinez:
As suggested in fommil/fpmortals#49

I couldn't compile and test this on my local machine due to a "file name too long" error during compilation.

### DIFF
--- a/core/src/main/scala/scalaz/EitherT.scala
+++ b/core/src/main/scala/scalaz/EitherT.scala
@@ -241,7 +241,7 @@ object EitherT extends EitherTInstances {
   def eitherTU[FAB, AB, A0, B0](fab: FAB)(implicit
                                           u1: Unapply[Functor, FAB]{type A = AB},
                                           @deprecated("scala/bug#5075", "") u2: Unapply2[Bifunctor, AB]{type A = A0; type B = B0},
-                                          l: Leibniz.===[AB, A0 \/ B0]
+                                          l: AB === (A0 \/ B0)
   ): EitherT[u1.M, A0, B0] = eitherT(l.subst[u1.M](u1(fab)))
 
   def monadTell[F[_], W, A](implicit MT0: MonadTell[F, W]): EitherTMonadTell[F, W, A] = new EitherTMonadTell[F, W, A]{
--- a/core/src/main/scala/scalaz/Leibniz.scala
+++ b/core/src/main/scala/scalaz/Leibniz.scala
@@ -69,7 +69,8 @@ sealed abstract class LeibnizInstances {
 object Leibniz extends LeibnizInstances {
 
   /** `(A === B)` is a supertype of `Leibniz[L,H,A,B]` */
-  type ===[A,B] = Leibniz[⊥, ⊤, A, B]
+  @deprecated("Use scalaz's package's type alias instead", "7.3.x")
+  type ===[A,B] = scalaz.===[A, B]
 
   /** Equality is reflexive -- we rely on subtyping to expand this type */
   implicit def refl[A]: Leibniz[A, A, A, A] = new Leibniz[A, A, A, A] {
--- a/core/src/main/scala/scalaz/Leibniz.scala
+++ b/core/src/main/scala/scalaz/Leibniz.scala
@@ -35,7 +35,7 @@ sealed abstract class Leibniz[-L, +H >: L, A >: L <: H, B >: L <: H] {
 }
 
 sealed abstract class LeibnizInstances {
-  import Leibniz._
+  import Leibniz.{=== => _, _}
 
   implicit val leibniz: Category[===] = new Category[===] {
     def id[A]: (A === A) = refl[A]
@@ -80,10 +80,10 @@ object Leibniz extends LeibnizInstances {
   /** We can witness equality by using it to convert between types
    * We rely on subtyping to enable this to work for any Leibniz arrow
    */
-  implicit def witness[A, B](f: A === B): A => B =
+  implicit def witness[A, B](f: scalaz.===[A, B]): A => B =
     f.subst[A => ?](identity)
 
-  implicit def subst[A, B](a: A)(implicit f: A === B): B = f.subst[Id](a)
+  implicit def subst[A, B](a: A)(implicit f: scalaz.===[A, B]): B = f.subst[Id](a)
 
   /** Equality is transitive */
   def trans[L, H >: L, A >: L <: H, B >: L <: H, C >: L <: H](
@@ -165,7 +165,7 @@ object Leibniz extends LeibnizInstances {
     T[_ >: LA <: HA] /*: Injective*/,
     A >: LA <: HA, A2 >: LA <: HA
   ](
-    t: T[A] === T[A2]
+    t: scalaz.===[T[A], T[A2]]
   ): Leibniz[LA, HA, A, A2] = force[LA, HA, A, A2]
 
   def lower2[
@@ -175,6 +175,6 @@ object Leibniz extends LeibnizInstances {
     A >: LA <: HA, A2 >: LA <: HA,
     B >: LB <: HB, B2 >: LB <: HB
   ](
-   t: T[A, B] === T[A2, B2]
+   t: scalaz.===[T[A, B], T[A2, B2]]
   ): (Leibniz[LA, HA, A, A2], Leibniz[LB, HB, B, B2]) = (force[LA, HA, A, A2], force[LB, HB, B, B2])
 }
--- a/core/src/main/scala/scalaz/package.scala
+++ b/core/src/main/scala/scalaz/package.scala
@@ -113,6 +113,9 @@ package object scalaz {
   type <~[+F[_], -G[_]] = NaturalTransformation[G, F]
   type ~~>[-F[_,_], +G[_,_]] = BiNaturalTransformation[F, G]
 
+  /** `(A === B)` is a supertype of `Leibniz[L,H,A,B]` */
+  type ===[A,B] = Leibniz[⊥, ⊤, A, B]
+
   type ⊥ = Nothing
   type ⊤ = Any
 
--- a/core/src/main/scala/scalaz/syntax/std/Function1Ops.scala
+++ b/core/src/main/scala/scalaz/syntax/std/Function1Ops.scala
@@ -2,7 +2,6 @@ package scalaz
 package syntax
 package std
 
-import Leibniz.===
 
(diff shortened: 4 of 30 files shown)


core/src/main/scala/scalaz/std/TypeConstraint.scala
```diff
@@ -7,9 +7,9 @@ trait TypeConstraintInstances {
     def compose[A, B, C](f: B <:< C, g: A <:< B) = f.asInstanceOf[A <:< C]
   }
 
-  val =:= = new Category[=:=] {
+  val === = new Category[===] {
     def id[A] = implicitly
-    def compose[A, B, C](f: B =:= C, g: A =:= B) = f.asInstanceOf[A =:= C]
+    def compose[A, B, C](f: B === C, g: A === B) = f.asInstanceOf[A === C]
   }
```
JordanMartinez: Not sure whether this should be left alone or not...

JordanMartinez:
@fommil Since you initially suggested making this PR, do you have any insight as to why the build fails?

core/src/main/scala/scalaz/std/TypeConstraint.scala
fommil: I think we want to leave this one alone

fommil:
hmm, I shall defer to our liskov expert @alexknvl ... any ideas why this breaks the build?

core/src/main/scala/scalaz/package.scala
```diff
@@ -113,6 +113,8 @@ package object scalaz {
   type <~[+F[_], -G[_]] = NaturalTransformation[G, F]
   type ~~>[-F[_,_], +G[_,_]] = BiNaturalTransformation[F, G]
 
+  type ===[A,B] = Leibniz[⊥, ⊤, A, B]
```
fommil: there is already a `Leibniz.===`,  I think you should be able to just import it where it is needed without making an alias here.

core/src/main/scala/scalaz/std/TypeConstraint.scala
fommil: also I don't think the cast looks safe.

core/src/main/scala/scalaz/package.scala
JordanMartinez: That was my original approach. However, I decided not to continue with that approach because the other type aliases used through scalaz are not imported (I assume) because their type aliases are also defined here. I also read [this part in CONTRIBUTING.md](https://github.com/scalaz/scalaz/blob/series/7.3.x/CONTRIBUTING.md#hierarchy) that made me wonder whether I should refrain from doing so.

So, I was following what I thought was the example in the rest of the code. I can import the `Leibniz.===` if that's desired. If so, is there a specific place where the imports should go / not go?

core/src/main/scala/scalaz/syntax/std/Function1Ops.scala
```diff
@@ -16,7 +16,7 @@ final class Function1Ops[T, R](private val self: T => R) extends AnyVal {
 
   def unary_!(implicit m: Memo[T, R]): T => R = m(self)
 
-  def toValidation[E](e: => E)(implicit ev: R =:= Boolean): T => Validation[NonEmptyList[E], T] =
+  def toValidation[E](e: => E)(implicit ev: R === Boolean): T => Validation[NonEmptyList[E], T] =
     (t: T) => (if (self(t): Boolean) Success(t) else Failure(NonEmptyList(e)))
```
JordanMartinez: Once `R` really is understood to be `Boolean` by the copmiler, I'm assuming `(self(t): Boolean)` can be reduced to `self(t)`.

core/src/main/scala/scalaz/package.scala
hrhino: 👍 to putting `===` here. At least, I use it more than (say) `BiNaturalTransformation`, which gets a top-level alias. 

I would say that one should point to the other, though. Maybe deprecate `Liskov.===`  and make it an alias of `scalaz.===`?

hrhino:
Isn't the build breaking just because (e.g.) `G[A] =:= Id[A]` (in `BijectionT#lens`) introduces an implicit view from `G[A]` to `Id[A]`, but `G[A] === Id[A]` doesn't? You'd have to implicitly use `evG.apply` or something.

core/src/main/scala/scalaz/std/TypeConstraint.scala
alexknvl: The cast is safe, but it is possible to do that with `===` (or `=:=` in 2.13) without it. I think you might want to add an instance for `===` if there is not one and keep the one for `=:=`, in case anyone already using it.

core/src/main/scala/scalaz/syntax/std/Function1Ops.scala
alexknvl: Is there an implicit conversion when `R === Boolean` is in scope? Otherwise you need to do `ev(self(t))`.

alexknvl:
Yeah, that looks to be the case.

core/src/main/scala/scalaz/package.scala
JordanMartinez: The downside of deprecating the `Leibniz.===` alias is that all type signatures of `A === B` within `Leibniz.scala` now need to be transformed to `===[A, B]` to prevent an error being thrown when I run `sbt compile`.

JordanMartinez:
All of the builds passed except the `scala 2.13.0-M3 | Java 8 | TEST_PROJECT="rootJS"` one [here](https://travis-ci.org/scalaz/scalaz/jobs/373132596#L7566) with an `OutOfMemoryError`

core/src/main/scala/scalaz/package.scala
hrhino: what if you changed `import Leibniz._` there to `import Leibniz.{=== => _, _}`?

JordanMartinez: Oh, duh! On it!

hrhino:
I think that's a transient failure that happens sometimes with the JS build. I kicked it off again.

core/src/main/scala/scalaz/EitherT.scala
```diff
@@ -241,7 +241,7 @@ object EitherT extends EitherTInstances {
   def eitherTU[FAB, AB, A0, B0](fab: FAB)(implicit
                                           u1: Unapply[Functor, FAB]{type A = AB},
                                           @deprecated("scala/bug#5075", "") u2: Unapply2[Bifunctor, AB]{type A = A0; type B = B0},
-                                          l: Leibniz.===[AB, A0 \/ B0]
+                                          l: ===[AB, A0 \/ B0]
```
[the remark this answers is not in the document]
hrhino: Sadly it will be read as `\/[===[AB, A0], B0]`. `AB === (A0 \/ B0)` is safe though

core/src/main/scala/scalaz/package.scala
JordanMartinez: > most of them are in the companion. I don't think it's a huge downside at all.

Correct. I should have been clearer.

JordanMartinez:
The same test passed on the 2.13.0-M3 scala build but failed on the `scala 2.12.6 | java 8 | rootJS` build. It passed everywhere else. Can that single build be restarted?

JordanMartinez:
The restarted build failed again. Is there a larger issue at play here?

fommil:
@xuwei-k ? :crying_cat_face:

fommil:
I've restarted it. I really hate the scalajs and native builds. They are such a pain.

In any case, you might be able to run the tests locally with this
https://stackoverflow.com/questions/28565837/filename-too-long-sbt

Also, read the instructions I left in the `build.sbt` to avoid scalajs and native locally. Unfortunately, that doesn't help here :cry:

JordanMartinez:
@fommil thanks for the tip! I can build it locally now.

sjrd:
Your Travis build looks like it performs `test:fastOptJS` (as a dependency of `test`) of *all* your JS projects *at the same time* (because of aggregation). That's super bad in terms of memory consumption because each `fastOptJS` is memory-hungry, and is not at all the recommended way to do it. You can significantly improve your batch memory consumption on travis with:

a) Add the following settings to your `scalajsProjectSettings`:
```scala
scalaJSLinkerConfig in testSuiteJS ~= { _.withBatchMode(System.getenv("CONTINUOUS_INTEGRATION") == "true") }
```
b) Run tests for various projects in sequence:
```bash
sbt oneProjectJS/test anotherProjectJS/test ...
```
This will ensure that

* Every project's `test:fastOptJS` is executed on its own, avoiding the cumulative memory consumption
* The linkers are configured in batch mode, which instructs them to release memory early instead of keeping it to speed up incremental recompilation (which is unnecessary in a CI job).

fommil:
@sjrd I wonder if perhaps the scalajs plugin could make use of `Tag`s, like I recently added to scalafix https://github.com/scalacenter/scalafix/pull/697

xuwei-k:
https://github.com/scalaz/scalaz/pull/1740 🤔

sjrd:
Good question. I will have to investigate that.

Note that in general, we've been moving to doing the *least* amount of configuration in the core sbt-scalajs plugin (it's much cleaner now in 1.x), because it's easier to *add* config in users' projects than to remove it. But we could at least strongly encourage in the docs a one-liner configuration like that. Or put it in sbt-scalajs if it is easy to properly reconfigure by users.

xuwei-k:
> all your JS projects at the same time (because of aggregation)

scalaz have [only one test project](https://github.com/scalaz/scalaz/tree/series/7.3.x/tests) (e.g. `core/src/test` does not exits) 😕

sjrd:
Ah!

Then yes maybe the change in your PR will help reduce the dominant factor.

hrhino:
Thanks for the advice, @sjrd! 🙇

TomasMikula:
Thanks! And welcome to scalaz 😉

core/src/main/scala/scalaz/Leibniz.scala
```diff
@@ -69,7 +69,8 @@ sealed abstract class LeibnizInstances {
 object Leibniz extends LeibnizInstances {
 
   /** `(A === B)` is a supertype of `Leibniz[L,H,A,B]` */
-  type ===[A,B] = Leibniz[⊥, ⊤, A, B]
+  @deprecated("Use scalaz's package's type alias instead", "7.3.x")
```
fommil: I didn't even know this was possible

Issue #1706, opened 2018-04-10
Potentially shortcircuiting foldMaps


TomasMikula:
Implement `foldMap` that supports early termination whenever the target `Monoid` does, for sequential data structures:

 - [ ] `List`
 - [ ] `IList`
 - [ ] `NonEmptyList`
 - [ ] `EphemeralStream`

This is analogous to what's already been done for `Stream` in https://github.com/scalaz/scalaz/commit/1fbc0cb10e555667e69e1360b8d4b1ad1d3cfb5e#diff-bf0e1bb9c3e7f9beb88c492281f38ed7

malancas:
Hi, I'm a new contributor. I've started an implementation and would like to pick this issue up if it's okay. Thanks.

PR #1738, opened 2018-04-30
Issue #1706 - Potentially short-circuiting foldMap

malancas:
## [Issue #1706 ](https://github.com/scalaz/scalaz/issues/1706)

I have a question regarding adding tests for the new implementations. I was unable to add a test for the NonEmptyList foldMap implementation to `NonEmptyListTest.scala` but was able to add corresponding tests for the other data structures. Is there something different about that data structure? Also, I was curious if adding more psumMap test cases utilizing the different data structures like I did made sense.

### DIFF
--- a/core/src/main/scala/scalaz/EphemeralStream.scala
+++ b/core/src/main/scala/scalaz/EphemeralStream.scala
@@ -171,7 +171,10 @@ sealed abstract class EphemeralStreamInstances {
     override def foldRight[A, B](fa: EphemeralStream[A], z: => B)(f: (A, => B) => B): B =
       if(fa.isEmpty) z else f(fa.head(), foldRight(fa.tail(), z)(f))
     override def foldMap[A, B](fa: EphemeralStream[A])(f: A => B)(implicit M: Monoid[B]) =
-      this.foldRight(fa, M.zero)((a, b) => M.append(f(a), b))
+      M.unfoldrSum(fa)(as => as.headOption match {
+        case Some(a) => Maybe.just((f(a), as.tailOption.getOrElse(EphemeralStream())))
+        case None => Maybe.empty
+      })
     override def foldMap1Opt[A, B](fa: EphemeralStream[A])(f: A => B)(implicit B: Semigroup[B]) =
       foldMapRight1Opt(fa)(f)((l, r) => B.append(f(l), r))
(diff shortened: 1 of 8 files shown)


TomasMikula:
Thanks for the PR!

What do you mean by "I was unable to add a test"?

psumMap tests for the other data structures definitely make sense. Just
note that you are not really testing stack safety with a list/stream of
size 7.

malancas:
In regards to being unable to add a test to NonEmptyList, I received an error stating that `foldMap is not a member of NonEmptyList`. However, I was able to resolve the issue. I'm currently addressing the list/stream sizes. Thanks for your feedback.

malancas:
Apologies for the delay on this. After working a bit more on it, it seems like the `foldRight from foldMap` test in `FoldableTest.scala` can’t rely on lazy evaluation anymore. Additionally, due to the non-lazy nature of `List`, `IList`, and `NonEmptyList`, I removed the “stack-safe” wording from the corresponding `psumMap` tests since it seemed like using large sized instances of these data structures in the tests caused stack issues. If there’s anything I’ve misunderstood, I’d be happy to update. Thanks!

TomasMikula:
> Additionally, due to the non-lazy nature of `List`, `IList`, and `NonEmptyList`, I removed the “stack-safe” wording from the corresponding psumMap tests since it seems using very large sized instances of these data structures (10000 elements each) in the tests caused stack issues.

The stack overflow is caused by their overridden implementations of `psumMap`. We can now just remove those implementations and rely on the [default ones](https://github.com/scalaz/scalaz/blob/1fbc0cb10e555667e69e1360b8d4b1ad1d3cfb5e/core/src/main/scala/scalaz/Foldable.scala#L286-L292) in terms of `foldMap` and the tests should pass.

> After working a bit more on it, it seems like the `foldRight from foldMap` test in `FoldableTest.scala` can’t rely on lazy evaluation anymore.

Yeah, I don't think we should expect that `foldRight` from `foldMap` should be lazy, or that `foldRight` from `foldMap` from `foldRight` should preserve the laziness of the original `foldRight`.

@jbgi, what do you think, since you added that test recently?

 Also, how would it come up in practice that we redefine the `Foldable[EphemeralStream]` instance from the original instance's `foldMap`?

TomasMikula:
Thanks for the contribution, @malancas! 🎉

PR #1740, opened 2018-05-01
execute "test:fastOptJS" explicitly before test. avoid OutOfMemoryError

xuwei-k:
https://github.com/scalaz/scalaz/pull/1734#issuecomment-385627061

> each fastOptJS is memory-hungry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ jdk:
 script:
   - git config core.whitespace tab-in-indent,trailing-space,blank-at-eol
   - git show --oneline --check
-  - sbt "++$TRAVIS_SCALA_VERSION!" -J-Xmx3784m checkGenTypeClasses "project $TEST_PROJECT" test:compile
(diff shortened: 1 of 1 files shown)


sjrd:
This is likely to cause the same OutOfMemoryError but now during the `rootJS/test:fastOptJS`. You have only moved the problem one sbt command earlier.

*Executing* tests consumes virtually no memory in the JVM in Scala.js, since everything is offloaded to Node.js instances, which have their own memory space. This change might help a bit in that it reduces the amount of *system* memory that you use overall, but the bigger win is in running the various `test:fastOptJS` separately from each other.

sjrd:
Given https://github.com/scalaz/scalaz/pull/1734#issuecomment-385629550, this change might actually be taking care of a bottleneck, indeed.

Although if there really is only one test project, why run `rootJS/test` instead of `theTestProjectJS/test`? It would avoid spawning spurious linkers trying to link codebases without any entry point for no reason. (It would also slightly help the JVM side, note!)

fommil:
I wonder if [class-monkey](https://gitlab.com/fommil/class-monkey) would help with the OOM. Although it doesn't work for Java 9+

PR #1750, opened 2018-05-08
Add string interpolator for Debug

NeQuissimus:
Inspired by #1496

### DIFF
--- a/base/shared/src/main/scala/scalaz/typeclass/DebugSyntax.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/DebugSyntax.scala
@@ -2,9 +2,35 @@ package scalaz
 package typeclass
 
 import language.experimental.macros
+import language.implicitConversions
 
 trait DebugSyntax {
   implicit final class ToDebugOps[A: Debug](self: A) {
     def debug: String = macro meta.Ops.f_0
+  }
+
+  implicit final def debugInterpolator(sc: StringContext): DebugInterpolator.Interpolator =
+    new DebugInterpolator.Interpolator(sc)
+}
+
+object DebugInterpolator {
+  final class HasDebug private (override val toString: String) extends AnyVal
+
+  object HasDebug extends HasDebug0 {
+    implicit def mat[A](x: A)(implicit D: Debug[A]): HasDebug = new HasDebug(D.debug(x))
+  }
+
+  sealed abstract class HasDebug0 {
+    @annotation.implicitAmbiguous(
+      "Cannot use the `z` interpolator to interpolate a value of type ${A}, as no implicit Show[${A}] instance is in scope."
+    )
+    implicit def ambiguousDebug1[A](a: A): HasDebug =
+      sys.error(s"Cannot use the `z` interpolator to interpolate ${a}, as no implicit Show instance is in scope.")
+    implicit def ambiguousDebug2[A](a: A): HasDebug =
+      sys.error(s"Cannot use the `z` interpolator to interpolate ${a}, as no implicit Show instance is in scope.")
+  }
+
+  final class Interpolator(val sc: StringContext) extends AnyVal {
+    def z(args: HasDebug*): String = sc.s(args: _*)
   }
 }
--- a/example/src/main/tut/typeclass/Debug.md
+++ b/example/src/main/tut/typeclass/Debug.md
@@ -40,3 +40,23 @@ Maybe.just(Foo(1)).debug
 ```
 
 *Note that the `Debug` instance for `Maybe` resolves against the `Debug` instance for `Foo`.*
+
+# String interpolator
+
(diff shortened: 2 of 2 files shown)


base/shared/src/main/scala/scalaz/typeclass/DebugSyntax.scala
```diff
@@ -2,9 +2,25 @@ package scalaz
...
   implicit final class ToDebugOps[A: Debug](self: A) {
     def debug: String = macro meta.Ops.f_0
   }
+
+  implicit final def debugInterpolator(sc: StringContext): DebugInterpolator.DebugInterpolator =
+    DebugInterpolator.DebugInterpolator(sc)
+}
+
+object DebugInterpolator {
+  final case class Debugs(override val toString: String) extends AnyVal
```
hrhino: arguably this could be (private constructor; not a case class)

```scala
final class Debugged private (override val toString: String) extends AnyVal
object Debugged {
  implicit def mat[A](x: A)(implicit A: Debug[A]): Debugged = 
    new Debugged(A.debug(a))
}
```

but I don't think it really matters

base/shared/src/main/scala/scalaz/typeclass/DebugSyntax.scala
```diff
@@ -2,9 +2,25 @@ package scalaz
...
+
+  object Debugs {
+    implicit def mat[A](x: A)(implicit D: Debug[A]): Debugs = Debugs(D.debug(x))
+  }
+
+  final case class DebugInterpolator(sc: StringContext) extends AnyVal {
```
hrhino: having `class DebugInterpolator` _inside_ `object DebugInterpolator` feels kinda weird.

also, and this is _really_ minor, this being a case class means that `sc"fnord"` will give the error

```scala
scala> sc"fnord"
<console>:13: error: StringContext does not take parameters
       sc"fnord"
       ^
```

rather than

```scala
scala> cs"fnord"
<console>:13: error: value cs is not a member of StringContext
       cs"fnord"
       ^
```

... not like anyone would do that, though, right?

Bonus:

```scala
scala> equals"${StringContext("", "")}"
res16: Boolean = true
```

example/src/main/tut/typeclass/Debug.md
```diff
@@ -40,3 +40,14 @@ Maybe.just(Foo(1)).debug
...
+# String interpolator
+
+Scalaz contains a custom string interpolator that takes advantage of `Debug`.
+Inside the `z` String interpolation, only objects that have `Debug` instances can be used.
+
+```tut
+val foo = Foo(123)
+
+val interpolated = z"${foo}"
+```
```
hrhino: possibly a block demonstrating that without the `Debug` instance, compilation fails?

base/shared/src/main/scala/scalaz/typeclass/DebugSyntax.scala
NeQuissimus: I renamed this class to "HasDebug" because it makes the type error much more useful:

```scala
scala> z"${bar}"
<console>:20: error: type mismatch;
 found   : Bar
 required: scalaz.typeclass.DebugInterpolator.HasDebug
       z"${bar}"
           ^
```

Seems more intuitive to require `HasDebug` than `Debugs`. (Although I wish I could customize this somehow and say "Need Debug". But I don't think this is possible here.)

hrhino: You can, [actually](https://scalafiddle.io/sf/Uycxpb8/0):

```scala
// Start writing your ScalaFiddle code here

trait Show[A] {
  def shows(a: A): String
}

final class Shewn(override val toString: String)
object Shewn extends Shewn0 {
  implicit def materialize[A](a: A)(implicit A: Show[A]): Shewn =
    new Shewn(A shows a)
}
sealed abstract class Shewn0 {
  @annotation.implicitAmbiguous("Cannot use the `z` interpolator to interpolate a value of type ${A}, as no implicit Show[${A}] instance is in scope.")
  implicit def fail1[A](a: A): Shewn = ???
  implicit def fail2[A](a: A): Shewn = ???
}

final class Z(sc: StringContext) {
  def z(shewns: Shewn*): String = sc.s(shewns : _*)
}

implicit def conv(sc: StringContext): Z = new Z(sc)

implicit val showStr: Show[String] = _.toString

println(z"asdf ${"zyfg"} xjrl") // compiles
println(z"this is an int ${1}")
// ScalaFiddle.scala:27: error: Cannot use the `z` interpolator to interpolate a value of type scala.this.Int, as no implicit Show[scala.this.Int] instance is in scope.
//  println(z"this is an int ${1}")
//                             ^
```

(I think I'll backport this to 7.3.x)

NeQuissimus: Oh, I like this a lot, let me update mine.

PR #1752, opened 2018-05-09
rename show"" -> z"", and use @implicitAmbiguous for more explanatory error messages

hrhino:
I don't know how to make this work with the test suite without setting up all the macro-based `shouldNotCompile` stuff, but it works.

Additionally, rename `show""` to `z""`, because I like it more. I'll be glad to take it out if other people don't like it, though.

cc @ChristopherDavenport

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -112,7 +112,7 @@ object build {
   )
 
   val lintOptions = Seq(
-    "-Xlint:_,-type-parameter-shadow",
+    "-Xlint:_,-type-parameter-shadow,-missing-interpolator",
     "-Ywarn-dead-code",
     "-Ywarn-unused-import",
     "-Ywarn-numeric-widen",
@@ -164,6 +164,8 @@ object build {
     },
     scalacOptions in (Compile, compile) ++= "-Yno-adapted-args" +: lintOptions,
     scalacOptions in (Test, compile) ++= lintOptions,
+
+    scala213_pre_cross_setting,
 
     scalacOptions in (Compile, doc) ++= {
       val base = (baseDirectory in LocalRootProject).value.getAbsolutePath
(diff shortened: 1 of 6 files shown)


xuwei-k:
`@implicitAmbiguous` does not exists in Scala 2.11.x

hrhino:
oh wow, it's newer than I thought

NeQuissimus:
Code looks fine otherwise :D

project/build.scala
```diff
@@ -284,6 +286,19 @@ object build {
...
+    // @fommil tells me we could use sbt-sensible for this
+    unmanagedSourceDirectories in Compile ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((2L, minor)) =>
+          Some((baseDirectory in Compile).value.getParentFile / s"src/main/scala-2.$minor")
+        case _               =>
+          None
+      }
+    }
+  }
```
hrhino: @fommil wdyt

fommil: I love the way it's a `Long`.

Implying that the madness will not stop with Scala 3.

project/build.scala
```diff
@@ -284,6 +286,19 @@ object build {
     }
   )
 
+  private[this] val scala213_pre_cross_setting = {
+    // sbt wants `scala-2.13.0-M1`, `scala-2.13.0-M2`, ... (sbt/sbt#2819)
+    // @fommil tells me we could use sbt-sensible for this
+    unmanagedSourceDirectories in Compile ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((2L, minor)) =>
+          Some((baseDirectory in Compile).value.getParentFile / s"src/main/scala-2.$minor")
```
fommil: there is a `sourceDirectory` you could use instead of hardcoding the `src/main` but it doesn't matter much.

hrhino: I am but a mortal.

project/build.scala
hrhino: I expect Scala to be the perfect language by version 34359738368.103079215116.2

ChristopherDavenport:
I'm a little skeptical about the switch to `z` I could be convinced of a different interpolator name, but that doesn't seem to convey a lot of semantic information about what you are attempting to express.

hrhino:
I'm torn: I dislike the length of `show""` (that's more than 5% of a line!) and I made `z` a bit of a pun on `s` from the stdlib; on the other hand, I can see how it would be potentially initially confusing.

The intention is to be as terse as `s` while providing the same level of safety, I guess.

JordanMartinez:
> I made `z` a bit of a pun on `s` from the stdlib; on the other hand, I can see how it would be potentially initially confusing.

Initially, yes, but isn't that partly what documentation is for? Couldn't one provide a simple gist / blog post that explains it quickly and which can be easily found via a Google search? One letter is about as terse as one can get. The pun makes sense: `s` for **S**cala's interpolator => `z` for Scala**Z**'s interpolator

project/build.scala
dwijnand: Who wants to live forever?

ChristopherDavenport:
I'm by no means against `z` just want to make sure something like the above explanation is discoverable.

hrhino:
> something like the above explanation is discoverable.

I'll throw a doc comment on `z` explaining the name, and possibly something on `Show`'s doc comment, too.

hrhino:
Fake?

PR #1754, opened 2018-05-10
Remove trailing whitespace.

TomasMikula:
Any ideas why this hasn't been caught by Travis in the [PR build](https://travis-ci.org/scalaz/scalaz/builds/376666222), but [was caught](https://travis-ci.org/scalaz/scalaz/builds/377230870) after merge?


hrhino:
shouldn't the whitespace be noticed in the last commit, though? or do we only check the diff?

PR #1758, opened 2018-05-10
no array copy IO Stack

jbgi:
Implementation of `Stack` with always O(1) push/pop by avoiding array copy, using array nesting instead. Also has the nice property to "shrink" as the stack unroll.

Benchmarks show a 5% percent perf improvement, but I'd like someone else confirm this on a desktop because I do not trust by laptop cpu to yield continuous/consistent performance.

And also some other minor commits.


jdegoes:
Excellent work so far.

PR #1768, opened 2018-05-12
Initial std module

NeQuissimus:
I know, I know... Nobody is excited about the standard library at all... We still need to do this, so I am gauging interest in HOW we will want to do it.

I have now created a separate module but refer to it in the same `examples` documentation as the base stuff. (mostly because it would probably be weird to publish multiple websites)

Is my basic idea here how we will want to tackle the stdlib?

- `scalaz.std._` import
- `FooInstances` trait per stdlib type
- "smooshMap" all the traits into the package object


build.sbt
```diff
@@ -52,8 +52,20 @@ lazy val metaJVM = meta.jvm
 
 lazy val metaJS = meta.js
 
+lazy val std = crossProject.module
+  .in(file("std"))
+  .settings(
+    libraryDependencies ++=
+      Seq("org.scala-lang" % "scala-reflect"  % scalaVersion.value,
+          "org.scala-lang" % "scala-compiler" % scalaVersion.value % "provided")
```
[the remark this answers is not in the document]
NeQuissimus: copy/pasta :D I don't think so...

NeQuissimus:
Ah, I see the point about primitives. I will remove that again. Those should definitely be in base

NeQuissimus:
Alright, so I did a thing...

I added a `scalaz.std` package to the base project that contains all the instances traits for the primitive types.
The rest of the `scalaz.std` package is in a separate module, with instances for `Either`, `List` and friends.

Naturally, everything has some degree of documentation.

I took some of the code from 7.3.x and let it "inspire" me (to copy and paste its implementation). I don't guarantee correctness. I do guarantee inefficient runtime.
I think all that is beyond this initial creation of the module. I just want to get the basics up and running, so people can contribute if they so please.

I am done for the moment and would ask for feedback before we merge.

base/shared/src/main/scala/scalaz/std/utils.scala
```diff
@@ -0,0 +1,10 @@
+package scalaz
+package std
+
+import typeclass.{ DebugClass, EqClass }
+
+object utils {
```
hrhino: make this `private[scalaz]` or `private[std]` rather than its members?

std/shared/src/main/scala/Either.scala
```diff
@@ -0,0 +1,24 @@
...
+    })
+
+  implicit def eitherEq[L, R](implicit X: Eq[L], Y: Eq[R]): Eq[Either[L, R]] =
+    instanceOf(new EqClass[Either[L, R]] {
+      def equal(first: Either[L, R], second: Either[L, R]): Boolean = (first, second) match {
+        case (Left(x), Left(y))   => X.equal(x, y)
+        case (Right(x), Right(y)) => Y.equal(x, y)
+        case _                    => false
+      }
+    })
```
hrhino: it's also a `Bifunctor` and `Debug`, maybe more

std/shared/src/main/scala/Tuple.scala
```diff
@@ -0,0 +1,32 @@
...
+        A.equal(a1, a2) && B.equal(b1, b2) && C.equal(c1, c2)
+      case _ => false
+    }
+
+  implicit final def tuple4Eq[A, B, C, D](implicit A: Eq[A], B: Eq[B], C: Eq[C], D: Eq[D]): Eq[(A, B, C, D)] =
+    instanceOf[EqClass[(A, B, C, D)]] {
+      case ((a1, b1, c1, d1), (a2, b2, c2, d2)) =>
+        A.equal(a1, a2) && B.equal(b1, b2) && C.equal(c1, c2) && D.equal(d1, d2)
+      case _ => false
+    }
```
hrhino: maybe we should get some sort of tupley codegen stuff happening, but not now

hrhino: also `Debug`, maybe the various `*->*`-kinded instances in the rightmost type parameter?

std/shared/src/main/scala/collection/List.scala
```diff
@@ -0,0 +1,35 @@
...
+    override def flatMap[A, B](oa: List[A])(f: A => List[B]): List[B] = oa.flatMap(f)
+    override def flatten[A](ma: List[List[A]]): List[A]               = ma.flatten
+  })
+
+  implicit def listEq[A: Eq]: Eq[List[A]] =
+    instanceOf(new EqClass[List[A]] {
+      def equal(first: List[A], second: List[A]): Boolean = (first.corresponds(second)(Eq[A].equal))
+    })
+
+  /* https://github.com/scalaz/scalaz/pull/1633
```
hrhino: 👍

std/shared/src/main/scala/collection/Set.scala
```diff
@@ -0,0 +1,19 @@
+package scalaz
+package std
+
+import typeclass.{ EqClass, MonadClass }
+
+trait SetInstances {
+  implicit val setMonad: Monad[Set] = instanceOf(new MonadClass[Set] {
```
hrhino: eww. how principled are we going to be about the stdlib instances?

NeQuissimus: So from what I read, a set has a monad if its `A` has an `Ord`... I don't know that I can represent that. So may be better to just remove Set?

hrhino: Yeah. The generalized form is

```scala 
trait RFunctor[F[_]] {
  type Constr[T]
  def rfmap[A, B: Constr](fa: F[A])(f: A => B): F[B]
}
```

and so on. (You get `trait Functor[F[_]] extends RFunctor[F] { type Constr[T] = DummyImplicit }`)

So either we include the generalizations like this (functors into subcategories IIRC) or forego this instance.

NeQuissimus: I removed it :D

Issue #1760, opened 2018-05-10
Change SafeApp to support MainResult


jdegoes:
A `MainResult` type can be defined that determines the behavior of the runtime system with respect to threads.

```scala
sealed trait MainResult
object MainResult {
  final case class Wait(code: Int) extends MainResult
  final case class Immediate(code: Int) extends MainResult
  final object Daemon extends MainResult
}
```

The design is up for discussion but the main feature should be the ability to exit without waiting for threads to finish, exit waiting for threads to finish, or not exit regardless of threads.

hrhino:
#1774

(just trying to keep the issue tracker clean, sorry John if this isn't actually done somehow)

jdegoes:
@hrhino Thanks!

Issue #1759, opened 2018-05-10
Change SafeApp to expect an error free main function


jdegoes:
`SafeApp` lets the user choose the type of error of the main function, which is a holdover from pre-bifunctor days.

It makes more sense to force the main fiber to handle all its errors. 

So this:

```scala
def run(args: List[String]): IO[Error, Unit]
```

can change to this:

```
def run(args: List[String]): IO[Void, Unit]
```

and the `Error` type member of `SafeApp` can be deleted, with documentation updated to match.

hrhino:
#1774

PR #1774, opened 2018-05-14
Modify `SafeApp#run` to return `IO[Void, ExitStatus]`

ktonga:
Solves #1759 and #1760

### DIFF
--- a/effect/jvm/src/main/scala/scalaz/effect/SafeApp.scala
+++ b/effect/jvm/src/main/scala/scalaz/effect/SafeApp.scala
@@ -1,15 +1,22 @@
 // Copyright (C) 2017 John A. De Goes. All rights reserved.
 package scalaz.effect
 
+import scala.concurrent.duration.Duration
+
 /**
  * The entry point for a purely-functional application on the JVM.
  *
  * {{{
+ * import java.io.IOException
  * import scalaz.effect.{IO, SafeApp}
  * import scalaz.effect.console._
  *
  * object MyApp extends SafeApp {
- *   def run(args: List[String]): IO[E, Unit] =
+ *
+ *   def run(args: List[String]): IO[Void, ExitStatus] =
+ *     myAppLogic.attempt.map(_.fold(_ => 1)(_ => 0)).map(ExitStatus.ExitNow(_))
+ *
+ *   def myAppLogic: IO[IOException, Unit] =
  *     for {
  *       _ <- putStrLn("Hello! What is your name?")
  *       n <- getStrLn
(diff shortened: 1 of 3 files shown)


effect/jvm/src/main/scala/scalaz/effect/SafeApp.scala
```diff
@@ -19,16 +24,23 @@ package scalaz.effect
  * }}}
  */
 trait SafeApp extends RTS {
-  type Error
+
+  final class ExitStatus(val code: Int)
```
jdegoes: This is good although it doesn't address the fundamental question:

*What should happen when the main fiber terminates?*

The answers to this question can be:

1. Exit immediately.
2. Exit when no thread in the RTS has any work.
3. Do not exit under any circumstance.

This may be tricky to implement, so I'd suggest to make the API now, and I can help implement (2). The other two can be implemented now and behavior of (2) can simply replicate behavior of (3).

jdegoes: As for API, perhaps `ExitStatus` can include `ResidentStatus` which can be one of `ExitNow`, `ExitWhenDone`, `DoNotExit`. 

Then you'd have: `ExitStatus(23, ExitNow)`, for example.

Or perhaps others have suggestions for a more attractive API.

Actually, it occurs to me that `DoNotExit` is incompatible with an exit code, so likely `ResidentStatus` should be folded into `ExitStatus` to give something like:

```scala
sealed trait ExitStatus
object ExitStatus {
  case object ExitNow(code: Int) extends ExitStatus
  case object ExitWhenDone(code: Int) extends ExitStatus
  case object DoNotExit extends ExitStatus
}
```

effect/jvm/src/main/scala/scalaz/effect/SafeApp.scala
```diff
@@ -19,16 +24,23 @@ package scalaz.effect
...
+    val Error: ExitStatus            = apply(1)
+  }
 
   /**
    * The main function of the application, which will be passed the command-line
-   * arguments to the program.
+   * arguments to the program and has to return an `IO` with the errors fully handled.
    */
-  def run(args: List[String]): IO[Error, Unit]
+  def run(args: List[String]): IO[Void, ExitStatus]
```
jdegoes: I like this, it's much cleaner than before.

effect/jvm/src/main/scala/scalaz/effect/SafeApp.scala
ktonga: Hey @jdegoes thanks for the quick response.

This PR was meant to solve just #1759 (Change SafeApp to expect an error free main function) and I think you are talking more about #1760 (Change SafeApp to support MainResult), I guess my "+ introducing ExitStatus" was confusing, it was just for carrying the status the jvm should exit with.

But don't worry! I'll include this too and solve both issues in this PR.

jdegoes: Two for one, I'll take it. 😉

ktonga: @jdegoes Done. Now with `ExitStatus` you can define when to exit.

effect/jvm/src/main/scala/scalaz/effect/SafeApp.scala
```diff
@@ -19,16 +26,32 @@ package scalaz.effect
...
-  final def main(args0: Array[String]): Unit = unsafePerformIO(run(args0.toList))
+  final def main(args0: Array[String]): Unit =
+    unsafePerformIO(run(args0.toList)) match {
+      case ExitStatus.ExitNow(code) =>
+        sys.exit(code)
+      case ExitStatus.ExitWhenDone(code, timeout) =>
+        unsafeShutdownAndWait(timeout)
+        sys.exit(code)
+      case ExitStatus.DoNotExit =>
+        unsafePerformIO(IO.never.toUnit)
```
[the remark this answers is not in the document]
ktonga: But what daemon threads do we have? The thread pool is creating non-daemon threads, that's why with the previous implementation the program exited after performing the IO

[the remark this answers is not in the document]
ktonga: I checked, and without that line it just terminates. I found that to be the most elegant way to keep it running. Do you have any other way in mind?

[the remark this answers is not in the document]
jdegoes: I think the meaning of `DoNotExit` should be: it will not call `System.exit(_)`, not that it will fire up other threads or sleep forever. Maybe it could be changed to `ExitNormally` or something similar. After all, if the user wants this behavior, they can achieve it by sleeping forever or sequencing `never` (which will result in a permanently blocked and wasted thread).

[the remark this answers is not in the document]
ktonga: Yeah yeah, totally on the same page now, it got it wrong at first. But check my latest changes, I removed the line.

jdegoes:
@ktonga Awesome work. Thank you for your contribution! 👊

PR #1775, opened 2018-05-14
Expose Inject typeclass

mrdziuban:
It looks like the `Inject` typeclass has always been sealed (initial commit: https://github.com/scalaz/scalaz/commit/1d79ad50981db1a6931172d58baf5b341b223c87), but I've come across a use case where I'd like to define custom instances.

### DIFF
--- a/core/src/main/scala/scalaz/Inject.scala
+++ b/core/src/main/scala/scalaz/Inject.scala
@@ -7,7 +7,7 @@ import std.option.{none, some}
  *
  * @see [[http://www.staff.science.uu.nl/~swier004/Publications/DataTypesALaCarte.pdf]]
  */
-sealed abstract class Inject[F[_], G[_]] extends (F ~> G) {
+abstract class Inject[F[_], G[_]] extends (F ~> G) {
   def apply[A](fa: F[A]): G[A] = inj(fa)
   def unapply[A](ga: G[A]): Option[F[A]] = prj(ga)
   def inj[A](fa: F[A]): G[A]


fommil:
interesting. Can you share any details about the use-case?

@ethul `sealed` it, but it's not clear if unsealing it is a good or bad idea.

mrdziuban:
Sure, I've found a use case for providing an instance to inject a coproduct into another coproduct if it can be proved that the first coproduct is used to build the second coproduct, e.g.

```scala
import scalaz.{Coproduct, Inject}
import scala.language.higherKinds

implicit def cop2CopInj[C[_], F[_], G[_]](implicit injF: Inject[F, C], injG: Inject[G, C]): Inject[Coproduct[F, G, ?], C] =
  new Inject[Coproduct[F, G, ?], C] {
    def inj[A](in: Coproduct[F, G, A]): C[A] = in.fold(injF, injG)
    def prj[A](c: C[A]): Option[Coproduct[F, G, A]] =
      injF.unapply(c).map(Coproduct.left[G](_))
        .orElse(injG.unapply(c).map(Coproduct.right[F](_)))
  }

type ListOpt[A] = Coproduct[List, Option, A]
type SetListOpt[A] = Coproduct[Set, ListOpt, A]

implicitly[Inject[ListOpt, SetListOpt]]
```

TomasMikula:
There's a problem with treating `Inject` as a (non-sealed) typeclass, because in general there isn't just one unique way to inject `F[_]`  into `G[_]`, which may cause typeclass incoherence.

TomasMikula:
For the same reason the `MonadPartialOrder` typeclass was removed. (See #1432 and specifically the commit message in https://github.com/scalaz/scalaz/pull/1432/commits/e528b357591b2f40998b29d812df715110c06f2e.)

fommil:
should we then consider resealing and adding this new implicit for derivations of that sort?

TomasMikula:
I don't find the mentioned use case very compelling without further context. Such intermediate coproducts will be costly at runtime, so requiring the adaptation to be explicit might be a good way to discourage them.

fommil:
(feel free to revert... I have no strong feelings)

TomasMikula:
@mrdziuban Wouldn't it actually be better in your use case to work with the target coproduct (possibly appropriately abstracted à la tagless final) from the get-go?

fommil:
umm... actually I've changed my mind. I think we should unseal `Inject` and that might close #1839 

Looking more closely at `Inject` it doesn't reference `Coproduct` directly (just the evidence) so we could provide iota evidence for this and not have to change anything.

mrdziuban:
I'm using coproducts to combine DSLs with the free monad pattern and have a few different coproducts defined to keep concerns separated. There are certain parts of the code that use the larger coproduct and certain parts that use a smaller one because I don't want those parts to have access to certain functionality. This way the compiler enforces that the code can't access anything it shouldn't.

With the implicit instance from my comment above, I'm able to interpret a program of the smaller coproduct into a program of the larger coproduct. Regarding your comment, @TomasMikula, I don't want to always work with the target coproduct because then I lose the compile-time guarantee.

fommil:
@mrdziuban with talk like that... you've earned yourself an appointment as reviewer of #1837 :stuck_out_tongue:

TomasMikula:
@mrdziuban So you have something like

```scala
def doStuff: Free[Coproduct[FooOp, BarOp, ?], Int]
def doMoreStuff: Free[Coproduct[FooOp, Coproduct[BarOp, BazOp, ?], ?], String]
```

and you want to be able to inject `Coproduct[FooOp, BarOp, A]` into `Coproduct[FooOp, Coproduct[BarOp, BazOp, ?], A]`.

What I'm saying is that you could refactor to

```scala
def doStuff[F[_]](implicit I: Inject[FooOp, F], J: Inject[BarOp, F]): Free[F, Int]
def doMoreStuff[F[_]](implicit I: Inject[FooOp, F], J: Inject[BarOp, F], K: Inject[BazOp, F]): Free[F, String]
```

and work with some abstract `F[_]`, which will ultimately be instantiated to the target type `Coproduct[FooOp, Coproduct[BarOp, BazOp, ?], ?]` without any intermediate representations. We have eliminated the need for the injections added by this PR and at the same time saved some allocations.

Refactoring further you can go fully tagless final

```scala
def doStuff[F[_] : Monad : Foo : Bar]: F[Int]
def doMoreStuff[F[_] : Monad : Foo : Bar : Baz]: F[String]
```

PR #1776, opened 2018-05-14
add Monoid instance for Unit; add Monoid and Applicative instances fo…

jdegoes:
…r Fiber; refactor test to use fiber append

### DIFF
--- a/base/shared/src/main/scala/scalaz/typeclass/SemigroupSyntax.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/SemigroupSyntax.scala
@@ -6,5 +6,7 @@ import scala.language.experimental.macros
 trait SemigroupSyntax {
   implicit final class ToSemigroupOps[A: Semigroup](a: A) {
     def append(f: => A): A = macro meta.Ops.fa_1
+
+    def <>(f: => A): A = macro meta.Ops.fa_1
   }
 }
--- a/base/shared/src/main/scala/scalaz/typeclass/SemigroupSyntax.scala
+++ b/base/shared/src/main/scala/scalaz/typeclass/SemigroupSyntax.scala
@@ -6,7 +6,5 @@ import scala.language.experimental.macros
 trait SemigroupSyntax {
   implicit final class ToSemigroupOps[A: Semigroup](a: A) {
     def append(f: => A): A = macro meta.Ops.fa_1
-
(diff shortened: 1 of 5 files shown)


base/shared/src/main/scala/scalaz/typeclass/SemigroupSyntax.scala
```diff
@@ -6,5 +6,7 @@ import scala.language.experimental.macros
 trait SemigroupSyntax {
   implicit final class ToSemigroupOps[A: Semigroup](a: A) {
     def append(f: => A): A = macro meta.Ops.fa_1
+
+    def <>(f: => A): A = macro meta.Ops.fa_1
```
hrhino: Does this resolve? I don't see how...

After @alexknvl's #1714 it should be easy to get this to work, though; 👍 to the name.

effect/shared/src/main/scala/scalaz/effect/Fiber.scala
```diff
@@ -37,4 +39,29 @@ trait Fiber[E, A] {
    * successfully interrupted or has produced its result.
    */
   def interrupt[E2](t: Throwable): IO[E2, Unit]
+
+  /**
+   * Monoidal composition of fibers.
+   */
+  final def <>(that: => Fiber[E, A])(implicit M: Monoid[A]): Fiber[E, A] =
+    zipWith(that)(M.append(_, _))
```
hrhino: since this is the monoid instance's `<>`, would it make sense to have one delegate to the other? I think the consensus was for the typeclass instance method to delegate to the class's method, but I'm not sure of that.

base/shared/src/main/scala/scalaz/typeclass/SemigroupSyntax.scala
jdegoes: It definitely doesn't resolve. I'll delete it.

effect/shared/src/main/scala/scalaz/effect/Fiber.scala
jdegoes: 👍 Class method is faster so makes sense to delegate to that one.

hrhino:
I would like the `<>` (_á la_ Haskell) back once we have explicitly-named syntax methods from Alex's PR.

Issue #1763, opened 2018-05-11
module and package object initialization causes deadlocks


hrhino:
I'm not sure what to do about this (or how to ensure it'll never be a problem).

see, e.g. comments on scalaz/scalaz#1756, and [related Gitter discussion](https://gitter.im/scalaz/scalaz?at=5af56d491cfca775e1fe99c8).

jbgi:
@mielientiev what was the workaround that you applied for #1756 to solve the issue?

In any case in seems wise that modules should not have any `val`s, ie. either use lazy val or object holder.
eg. for the case of `IListImpl.isCovariantInstance` the solution could be 
```scala

private[data] object IListImpl extends IListModule {
  type IList[A] = Fix[Maybe2[A, ?]]

  def uncons[A](as: IList[A]): Maybe2[A, IList[A]] =
    Fix.unfix[Maybe2[A, ?]](as)

  implicit def isCovariantInstance: IsCovariant[IList] = IListImplInstances.isCovariantInstance

}

private[data] object IListImplInstances {
  import IListImpl.IList

  implicit val isCovariantInstance: IsCovariant[IList] = instanceOf(new IsCovariantClass.LiftLiskov[IList] {

    override def liftLiskov[A, B](implicit ev: A <~< B): IList[A] <~< IList[B] = {
      type <~~<[F[_], G[_]] = ∀.Prototype[λ[α => F[α] <~< G[α]]]
      val ev1 = Λ[α](Maybe2.isCovariant_1[α].liftLiskov[A, B]): Maybe2[A, ?] <~~< Maybe2[B, ?]
      Fix.liftLiskov[Maybe2[A, ?], Maybe2[B, ?]](ev1.make)(Maybe2.isCovariant_2[A])
    }
  })
}
```

mielientiev:
the workaround was just to swap tests. (add a test to the beginning)

PR #1779, opened 2018-05-14
Add semigroup instance for Void  #1777

### LINKED ISSUE #1777: Add Semigroup instance for Void

wi101:
Fixes #1777

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/Void.scala
+++ b/base/shared/src/main/scala/scalaz/data/Void.scala
@@ -2,7 +2,7 @@ package scalaz
 package data
 
 import com.github.ghik.silencer.silent
-import scalaz.typeclass.{ EqClass, SemigroupClass }
+import scalaz.typeclass.SemigroupClass
 
 trait VoidModule {
   type Void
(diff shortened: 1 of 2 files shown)


base/shared/src/main/scala/scalaz/data/Void.scala
```diff
@@ -31,8 +32,17 @@ object VoidModule extends VoidSyntax {
   implicit def void_<~<[A]: Void <~< A = Void.conforms[A]
 }
 
+trait VoidInstance {
```
jdegoes: For consistency with other examples, this should be called `VoidInstances`.

wi101:
There is a compile error because the implicit val voidEq is moved in `Void`, there is a val Iso in `Prelude.scala` which uses this implicit Eq[Void] (package scalaz extends BaseTypeclasses ... and `Iso` is in BaseTypeclasses trait). I change Iso to a lazy val, the test works again.
@jdegoes What do you think? Do I keep this change or move Eq[Void] to EqInstances?

jdegoes:
@wi101 I am good with either way (either keep the lazy val, or do not move `Eq[Void]` into `VoidInstances`). I think the question of how to solve these types of problems in general is undecided (#1763).

wi101:
Ok, I'll do the second ;)

jdegoes:
Thanks for the pull request! 🙏 😄

Issue #1601, opened 2017-12-26
missing LICENSE in jar file


fommil:
technically we're violating our own LICENSE by not including it in the jar file.

I have a simple fix, from `sbt-sensible`, I'll send a PR... something like this

```
    resources in Compile ++= {
      val orig = (resources in Compile).value
      val base = baseDirectory.value
      val root = (baseDirectory in ThisBuild).value

      def fileWithFallback(name: String): File =
        if ((base / name).exists) base / name
        else if ((root / name).exists) root / name
        else throw new IllegalArgumentException(s"legal file $name must exist")

      Seq(fileWithFallback("LICENSE"), fileWithFallback("NOTICE"))
    },
```

fommil:
(yeah... my son was born shortly after writing this ticket. You're not getting a PR anytime soon, sorry)

hrhino:
#1783

hrhino:
actually, I'm going to leave this open to remind myself to forward port to sz8.

PR #1783, opened 2018-05-16
add LICENSE.txt to all packaged jars.

hrhino:
update license year

Fixes scalaz/scalaz#1601

(I tried using `resources`, but that doesn't get included with the docs; this seems the easiest way to include some random file with no other meaning.)

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2009-2014 Tony Morris, Runar Bjarnason, Tom Adams, Kristian Domagala, Brad Clow, Ricky Clarkson, Paul Chiusano, Trygve Laugstøl, Nick Partridge, Jason Zaugg
+Copyright (c) 2009-2018 Tony Morris, Runar Bjarnason, Tom Adams, Kristian Domagala, Brad Clow, Ricky Clarkson, Paul Chiusano, Trygve Laugstøl, Nick Partridge, Jason Zaugg
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/project/build.scala
+++ b/project/build.scala
@@ -302,6 +302,12 @@ object build {
         }
         </developers>
       ),
+
+    licenseFile := {
+      val LICENSE_txt = (baseDirectory in ThisBuild).value / "LICENSE.txt"
+      if (!LICENSE_txt.exists()) sys.error(s"cannot find license file at $LICENSE_txt")
+      LICENSE_txt
+    },
     // kind-projector plugin
     libraryDependencies ++= (scalaBinaryVersion.value match {
       case "2.10" =>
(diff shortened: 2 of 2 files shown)


project/build.scala
```diff
@@ -396,6 +405,8 @@ object build {
     }
   }
 
+  lazy val licenseFile = settingKey[File]("The license file to include in packaged artifacts")
```
fommil: no need for `lazy`, I think it's cargo culting at this point :smiley:

hrhino: It's 100% cargo culting I guess. Sorry :P

fommil: it's not worth pushing an update to this PR, the CI uses a small diesel power plant ... think of the carbon.

hrhino: and thus are these myths perpetuated.

LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2009-2014 Tony Morris, Runar Bjarnason, Tom Adams, Kristian Domagala, Brad Clow, Ricky Clarkson, Paul Chiusano, Trygve Laugstøl, Nick Partridge, Jason Zaugg
+Copyright (c) 2009-2018 Tony Morris, Runar Bjarnason, Tom Adams, Kristian Domagala, Brad Clow, Ricky Clarkson, Paul Chiusano, Trygve Laugstøl, Nick Partridge, Jason Zaugg
```
fommil: I really think we should add @xuwei-k to this list. Is there a reason you have not added yourself?

hrhino: Honestly I'm confused by this whole names-in-licensing situation. Obviously the copyright is held by more people than this, and this list doesn't match what's in the `pomExtra`. 

Would it be the best to just say "the ScalaZ contributors"? That's not false. I could go through the git history and make a `CONTRIBUTORS` file that we could pinky-swear to keep up-to-date.

Also, is there a reason that _you_ haven't added _yourself?

hrhino: The super-high-tech solution is to keep a list of people lying around, then have the package step prepend `Copyright © 2009-$currentYear ${folks mkString ", "}` to this file. Then we could add the same to the `pomExtra`.

fommil: Thing with BSD-3 is you can't change it. Technically you'd need everybody to agree to even change this, but I don't think anybody will mind a year bump and extra names from real contributors.

I don't feel I've added enough to 7.2 to justify adding my name. But I've added it in 7.3 because I added the `Deriving` base typeclasses and it took me a lot of time to figure that out.

Yeah, it should be everybody's name. Apache 2.0 encourages a `NOTICE` file, which can even be generated from the git history.

hrhino: Is this line technically part of the license, then? That seems strange, given that under BSD everyone retains their copyright whether or not they're here. Would "Copyright 2009-2018, the Contributors (see included CONTRIBUTORS file)" pass muster, then?

fommil: yes this line is part of the license, and no it's not conventional for a BSD license to refer to a contributors file because it isn't included in the license. You'd have the words "copyright the contributors, see in the included file" with no obligation for anybody to actually share that file. So you'd end up creating a 4th clause. etc etc etc.

fommil: @hrhino I like that solution for scalaz8. We'd need to have a version of the file itself for our own source distribution.

hrhino: Right, so, I was thinking "copyright the contributors\n${bsd license}" in `LICENSE`, and when it's packaged we insert the names into that and the pom.

But that presumably wouldn't be legit, because it would imply that we're distributing it under a (textually) different LICENSE than what people see looking at the repo, and a different one than the previous version whenever a new contributor joins the cabal.

> ... with no obligation for anybody to actually share that file. So you'd end up creating a 4th clause.

BSD4, you saw it here first

hrhino:
cool, thanks!

Issue #1540, opened 2017-11-28
Scalaz 8 package organization


jdegoes:
A place to discuss a more fine-grained package organization for Scalaz 8.

 * **scalaz.algebra**
 * **scalaz.ct**
 * **scalaz.data**
 * **scalaz.effect**
 * **scalaz.misc**
 * **scalaz.mtl**
 * **scalaz.mutable**
 * **scalaz.test**
 * **scalaz.type**

NeQuissimus:
Could we get a quick description for each of these? Are you planning to build a testing framework? type == typeclass?
Will every package have some sort of Prelude?

jdegoes:
@NeQuissimus Loosely based on [this ticket](https://github.com/scalaz/scalaz/issues/1526#issuecomment-347383066). 

I don't imagine each package would have its own prelude, just one prelude for the whole project.

As for testing framework, @xuwei-k has donated [Scalaprops](https://github.com/scalaprops/scalaprops) to Scalaz 8, and we could make some improvements on that similar to Hedgehog.

NeQuissimus:
Ah, makes sense. And "thank you" to @xuwei-k !

TomasMikula:
"package" as a binary artifact (separate sbt project), or `package` in the sense of Scala?

jdegoes:
@TomasMikula `package` in the sense of Scala.

TomasMikula:
I would like if the dependencies between packages formed an _acyclic_ graph. Not sure if that's feasible. We could use [lihaoyi/acyclic](https://github.com/lihaoyi/acyclic#package-cycles) to enforce that.

jdegoes:
👍

Jacoby6000:
I think having a `misc` package is dangerous unless there are very clearly defined things which are miscellaneous.  Usually it winds up just being a bucket of unrelated things which people were too lazy to find a good place for.

vmarquez:
@TomasMikula I'm not sure if that's possible or worth compromising for... what comes to mind is the usefulness of having typeclass/syntax for data types, but also the fact that we want some type classes to return Maybe for example...

jdegoes:
@Jacoby6000 Fixed

@vmarquez the type classes can depend on the data package, and implement instances for them there (no orphans).

TomasMikula:
> I'm not sure if that's possible

[...]

PR #1784, opened 2018-05-16
Split typeclass definitions to algebra and ct packaged

saczawap:
https://github.com/scalaz/scalaz/issues/1540

I didn't want to make the PR to big so I moved only typeclasses as the first step.
The question is what to do with:
- DebugClass
- EqClass - algebra?
- IsContravariantClass
- IsCovariantClass


base/shared/src/main/scala/scalaz/Prelude.scala
```diff
@@ -112,9 +114,9 @@ trait AllFunctions
     with data.DisjunctionFunctions
     with data.MaybeFunctions
     with data.KleisliFunctions
-    with typeclass.InvariantFunctorFunctions
```
NeQuissimus: As redundant as it may be, I quite like the package prefixes on the traits here, so one can find them more easily...

saczawap: yeah probably makes sense :+1:

jdegoes:
@saczawap Awesome work! I'm really excited about this change because it will make it vastly easier to find what you're looking for, especially if you're using a text editor and not an IDE. The old type class / data distinction that has existed for some time is not fine-grained (like Haskell or PureScript) and results in too many unrelated classes in each directory.

For `IsCovariant`, etc., I would suggest `scalaz.types`.

For `Eq`, I think `scalaz.relations` may make sense, but it's possible there won't be much in there (equality, ordering, etc.). If so, maybe a `scalaz.core` makes more sense (although I worry about it becoming a dumping ground for poorly organized classes).

Anyone else have feedback on this?

edmundnoble:
This change doesn't make sense to me. All of the type classes under CT are algebraic; all of the type classes under algebra have CT representations.

emilypi:
I think deserves some deliberation, considering the size of the algebra additions we're looking at, in addition to their use-case. I agree with @edmundnoble to some degree that any separation of algebras  , typeclasses, and category libraries will have some cognitive dissonance associated with them. After all, every algebra can be viewed as a Lawvere Doctrine in `ct`, or as categories themselves (e.g. `Monoid` + `Group`s), but at the same time, people are probably only going to be using them from the naive definitions. 

On the other hand, I've been working on coming up with a coherent algebraic typeclass hierarchy, and the sheer number of typeclasses could potentially cause some difficulty using Scalaz 8. Here is the hierarchy I've been playing around with:

```scala
// Common Algebraic Hierarchy
  type Quasigroup[T]           = InstanceOf[QuasigroupClass[T]]
  type Loop[T]                 = InstanceOf[LoopClass[T]]
  type Semigroup[T]            = InstanceOf[SemigroupClass[T]]
  type Monoid[T]               = InstanceOf[MonoidClass[T]]
  type MonoidAction[T]         = InstanceOf[MonoidActionClass[T]]
  type Group[T]                = InstanceOf[GroupClass[T]]
  type GroupAction[T]          = InstanceOf[GroupActionClass[T]]
  type Torsor[T]               = InstanceOf[TorsorClass[T]]
  type CommutativeGroup[T]     = InstanceOf[CommutativeGroupClass[T]]
  type NearSemiring[T]         = InstanceOf[NearSemiringClass[T]]
  type Semiring[T]             = InstanceOf[SemiringClass[T]]
  type KleeneAlgebra[T]        = InstanceOf[KleeneAlgebraClass[T]]
  type StarSemiring[T]         = InstanceOf[StarSemiringClass[T]]
  type NearRing[T]             = InstanceOf[NearRingClass[T]]
  type NonUnitalRing[T]        = InstanceOf[NonUnitalRingClass[T]]
  type Ring[T]                 = InstanceOf[RingClass[T]]
  type CommutativeRing[T]      = InstanceOf[CommutativeRingClass[T]]
  type QuotientRing[T]         = InstanceOf[QuotientRingClass[T]]
  type Module[T]               = InstanceOf[ModuleClass[T]]
  type ProjectiveModule[T]     = InstanceOf[ProjectiveModuleClass[T]]
  type InjectiveModule[T]      = InstanceOf[InjectiveModuleClass[T]]
  type PrincipalIdealDomain[T] = InstanceOf[PrincipalIdealDomainClass[T]]
  type DedekindDomain[T]       = InstanceOf[DedekindDomainClass[T]]
  type RingAlgebra[T]          = InstanceOf[RingAlgebraClass[T]]
  type AssociativeAlgebra[T]   = InstanceOf[AssociativeAlgebraClass[T]]
  type VectorSpace[T]          = InstanceOf[VectorSpaceClass[T]]
  type FieldAlgebra[T]         = InstanceOf[FieldAlgebraClass[T]]
  type LieGroup[T]             = InstanceOf[LieGroupClass[T]]
  type LieAlgebra[T]           = InstanceOf[LieAlgebraClass[T]]
  type GradedAlgebra[T]        = InstanceOf[GradedAlgebraClass[T]]
  type Bialgebra[T]            = InstanceOf[BialgebraClass[T]]

  // Order Hierarchy
  type Ord[T]                  = InstanceOf[OrdClass[T]]
  type PreOrder[T]             = InstanceOf[PreOrderClass[T]]
  type PartialOrder[T]         = InstanceOf[PosetClass[T]]
  type TotalOrder[T]           = InstanceOf[TotalOrderClass[T]]

  // Order Algebra Hierarchy
  type Semilattice[T]          = InstanceOf[SemiLatticeClass[T]]
  type JoinSemilattice[T]      = InstanceOf[JoinSemilatticeClass[T]]
  type MeetSemilattice[T]      = InstanceOf[MeetSemilatticeClass[T]]
  type Lattice[T]              = InstanceOf[LatticeClass[T]]
  type BoundedLattice[T]       = InstanceOf[BoundedLatticeClass[T]]
  type DistributiveLattice[T]  = InstanceOf[DistributiveLatticeClass[T]]
  type Band[T]                 = InstanceOf[BandClass[T]]

  // Logical Algebra Hierarchy
  type BooleanAlgebra[T]       = InstanceOf[BooleanAlgebraClass[T]]
  type HeytingAlgebra[T]       = InstanceOf[HeytingAlgebraClass[T]]

  // Oid Rage Hierarchy
  type Semigroupoid[T]         = InstanceOf[SemigroupoidClass[T]]
  type Groupoid[T]             = InstanceOf[GroupoidClass[T]]

  // Algebraic Graph Heirarchy (help wanted)
  type Graph[T]                = InstanceOf[GraphClass[T]]
  type DirectedGraph[T]        = InstanceOf[DigraphClass[T]]
  type DirectedAcyclicGraph[T] = InstanceOf[DirectedAcyclicGraphClass[T]]
  type CayleyGraph[T]          = InstanceOf[CayleyGraphClass[T]]

```

A sane grouping of each of these concepts would be very appreciated from a user's perspective. It's simply too much clutter to not separate from the "standard" typeclass-opedia typeclasses.

alexknvl:
I think `Eq` belongs in `core`. It is the most fundamental typeclass there can possibly be (due to its laws; observational equivalence, the definition of purity and referential transparency depends on it). `Equiv` or something of that nature would belong in `algebra`.

jdegoes:
Technically everything in `data` has a representation in category theory. In fact, everything does, one way or another. That's not a reason, IMO, to group everything into a `ct` folder. 

Folders are indexes that allow humans to more quickly find what they're looking for. We should use these indexes to enable faster lookup of what would otherwise be a massive lake of unindexed files.

jdegoes:
@emilypi 

> On the other hand, I've been working on coming up with a coherent algebraic typeclass hierarchy, 

Can I just say that I'm really, really, really, really looking forward to see this PR? 

🎉 🕺🎉 🕺 🎉🕺 🎉🕺 🎉🕺 🎉🕺 🎉🕺

saczawap:
All type classes have been moved.

jdegoes:
👍 Looks great! There's some work to do moving more stuff into, e.g. `types`, but I realize this should be done incrementally so I'm good to merge now to prevent conflict hell.

Thanks @saczawap!

PR #1787, opened 2018-05-18
More stds

NeQuissimus:
Overprotective banana

Kids, be safe when dealing with the std.

This is a WIP for me to throw stdlib instances into.

So far:
- Bifunctor for Either
- Debug for Either
- Monad for Vector
- Eq for Vector
- WIP version of Debug for Vector
- WIP Ord for Boolean
- WIP Ord for Byte
- WIP Ord for Int
- WIP Ord for Long
- WIP Ord for Double
- Updated Eq for Double

Notes:
- Eq for Double considers `0.0` and `-0.0` to be equal, can differentiate `NaN`s that have different values of `Double.doubleToRawLongBits`

### DIFF
--- a/base/shared/src/main/scala/scalaz/std/Boolean.scala
+++ b/base/shared/src/main/scala/scalaz/std/Boolean.scala
@@ -1,18 +1,18 @@
 package scalaz
 package std
 
+import algebra.OrdClass
 import utils._
 
 trait BooleanInstances {
   implicit val booleanDebug: Debug[Boolean] = toStringDebug[Boolean]
   implicit val booleanEq: Eq[Boolean]       = universalEq[Boolean]
 
-  /* https://github.com/scalaz/scalaz/pull/1792
   implicit val booleanOrd: Ord[Boolean] = instanceOf(new OrdClass[Boolean] {
     def comp(a: Boolean, b: Boolean) = (a, b) match {
       case (true, false) => GT
       case (false, true) => LT
-      case _ => EQ
+      case _             => EQ
     }
-  })*/
+  })
 }
--- a/base/shared/src/main/scala/scalaz/std/Boolean.scala
+++ b/base/shared/src/main/scala/scalaz/std/Boolean.scala
@@ -6,4 +6,13 @@ import utils._
 trait BooleanInstances {
   implicit val booleanDebug: Debug[Boolean] = toStringDebug[Boolean]
   implicit val booleanEq: Eq[Boolean]       = universalEq[Boolean]
+
+  /* https://github.com/scalaz/scalaz/pull/1792
+  implicit val booleanOrd: Ord[Boolean] = instanceOf(new OrdClass[Boolean] {
+    def comp(a: Boolean, b: Boolean) = (a, b) match {
+      case (true, false) => GT
+      case (false, true) => LT
+      case _ => EQ
+    }
+  })*/
 }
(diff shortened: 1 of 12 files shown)


std/shared/src/main/scala/collection/Vector.scala
```diff
@@ -0,0 +1,35 @@
+package scalaz
+package std
+
+import typeclass.{ EqClass, MonadClass }
+
+trait VectorInstances {
+  implicit val vectorMonad: Monad[Vector] = instanceOf(new MonadClass[Vector] {
+    override def map[A, B](fa: Vector[A])(f: A => B): Vector[B]             = fa.map(f)
+    override def ap[A, B](fa: Vector[A])(f: Vector[A => B]): Vector[B]      = fa.zip(f).map { case (e, f) => f(e) }
```
hrhino: for stdlib stuff I wonder if it makes sense to manually fuse operations like this:

```scala
(fa.iterator zip f.iterator).map({ case (e, f) => f(e) })(breakOut)
```

/me ducks

std/shared/src/main/scala/collection/Set.scala
```diff
@@ -1,11 +1,15 @@
...
 
 trait SetInstances {
   implicit def setEq[A: Eq]: Eq[Set[A]] =
     instanceOf(new EqClass[Set[A]] {
       def equal(first: Set[A], second: Set[A]): Boolean = (first.toStream.corresponds(second.toStream)(Eq[A].equal))
     })
+
+  implicit val setFunctor: Functor[Set] = instanceOf(new FunctorClass[Set] {
+    def map[A, B](ma: Set[A])(f: A => B): Set[B] = ma.map(f)
+  })
```
hrhino: this has similar problems to `Monad[Set]`

NeQuissimus: Ah, you are right!

base/shared/src/main/scala/scalaz/std/Boolean.scala
```diff
@@ -6,4 +6,13 @@ import utils._
...
   implicit val booleanEq: Eq[Boolean]       = universalEq[Boolean]
+
+  /* https://github.com/scalaz/scalaz/pull/1792
+  implicit val booleanOrd: Ord[Boolean] = instanceOf(new OrdClass[Boolean] {
+    def comp(a: Boolean, b: Boolean) = (a, b) match {
+      case (true, false) => GT
+      case (false, true) => LT
+      case _ => EQ
+    }
+  })*/
```
hrhino: How useful is `Ord[Boolean]`? It seems like either choice is arbitrary.

base/shared/src/main/scala/scalaz/std/Double.scala
```diff
@@ -7,7 +7,25 @@ import typeclass.EqClass
 import utils._
 
 trait DoubleInstances {
+  private[this] val NegZero: Double = -0.0
+  private[this] val PosZero: Double = 0.0
```
hrhino: `final val` and use `0.0d` rather than the type ascription?

base/shared/src/main/scala/scalaz/std/Boolean.scala
NeQuissimus: Looked at Haskell's instance and figured "Why not?" :D

jdegoes:
@NeQuissimus Mainly some merge conflicts.

base/shared/src/main/scala/scalaz/std/Boolean.scala
jdegoes: It's useful. Remember some data structures require a total ordering even though it's arbitrary.

jdegoes:
This ready for merging?

PR #1791, opened 2018-05-18
move voidEq in VoidInstances

wi101:


base/shared/src/main/scala/scalaz/data/Void.scala
```diff
@@ -33,6 +33,8 @@ object VoidModule extends VoidSyntax {
 }
 
 trait VoidInstances {
+  implicit final val voidEq: Eq[Void] = instanceOf[EqClass[Void]]((a, b) => a.absurd)
```
hrhino: not sure if this being a `val` opens us up to #1763-style deadlocks...

wi101: oh :( it's the same problem. I thought this change would be possible, after moving all instances under scalaz/data and then Prelude extend them.

PR #1792, opened 2018-05-18
Add simple Ord

NeQuissimus:


base/shared/src/main/scala/scalaz/typeclass/Ord.scala
```diff
@@ -0,0 +1,23 @@
...
+final case object GT extends Ordering
+final case object EQ extends Ordering
+
+trait OrdClass[A] extends EqClass[A] {
+  def comp(a: A, b: A): Ordering
+
+  def <(a: A, b: A): Boolean = comp(a, b) eq LT
+  def <=(a: A, b: A): Boolean = {
+    val r = comp(a, b)
+    (r eq LT) || (r eq EQ)
```
hrhino: or `comp(a, b) ne GT`

base/shared/src/main/scala/scalaz/typeclass/Ord.scala
```diff
@@ -0,0 +1,23 @@
+package scalaz
+package typeclass
+
+sealed abstract class Ordering
```
hrhino: probably should `extends Product with Serializable` to avoid the LUB sadness.

```scala
scala> List(LT, GT)
res0: List[Product with Serializable with Ordering] = List(LT, GT)
```

although I'm not sure if that's really going to be a problem with this specific one.

example/src/main/tut/typeclass/Ord.md
```diff
@@ -0,0 +1,44 @@
+---
+layout: docs
+title:  "Ord"
+---
+
+# Ord
+
+*Ord is used for totally ordered datatypes.*
```
hrhino: I wonder if we should (eventually) have a `PartialOrder` typeclass, or even a `PreOrder`.

hrhino: Maybe explain that "totally ordered" means that any two values must have a defined ordering between them? The type of `comp` does imply that, though.

NeQuissimus: Is the explanation right below not sufficient?

jdegoes:
@NeQuissimus Just needs some merge conflicts resolved.

Also question where to put this: probably in `core` alongside `Eq`? /cc @alexknvl

alexknvl:
`algebra`?

PR #1794, opened 2018-05-18
WIP sufficient definition checker based on macros

hrhino:
This one is based on macros.

[This one](https://github.com/hrhino/plugin-poc) is more extensible and based on a plugin. (I named it more broadly because @alexknvl had mentioned wanting a place to put some plugin stuff under the scalaz banner.)

I'm putting this up here to invite discussion on whether this is the right way to go about this. I have similar functionality as a compiler plugin. There are pros and cons to each approach:

The main reason to use macros is that they will be available to 3rd party implementors, so we can run the minimality checking without them needing to do anything. If this is implemented as a compiler plugin, typeclass implementors would need to remember to compile with the plugin enabled. (I know sbt models compiler plugins as dependencies; there's no way to "transitively" enable a plugin, is there?)

The upshot to a compiler plugin is that it's a compiler plugin and therefore not based on this pretty meh API, and doesn't need the `protected val minimal` thing to make it work.

This uses only names, and therefore prevents overloading. I think that's a good thing, so I'll consider it on purpose.

TODO, assuming we go the macro way:

- [ ] tests
- [ ] apply everywhere, remove `Alt` traits
- [ ] more sanity checks (methods must exist, no overloading ...)

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/DisjunctionInstances.scala
+++ b/base/shared/src/main/scala/scalaz/data/DisjunctionInstances.scala
@@ -27,11 +27,13 @@ trait DisjunctionInstances {
     }
 
   implicit val disjunctionBifunctor: Bifunctor[Disjunction] =
-    instanceOf(new BifunctorClass[Disjunction] with BifunctorClass.DeriveLmapRmap[Disjunction] {
-      def bimap[A, B, S, T](fab: A \/ B)(as: A => S, bt: B => T): S \/ T = fab match {
-        case -\/(a) => -\/(as(a))
-        case \/-(b) => \/-(bt(b))
-      }
+    instanceOf(new BifunctorClass[Disjunction] {
+      val minimal = meta.IsMinimal()
+      override def bimap[A, B, S, T](fab: A \/ B)(as: A => S, bt: B => T): S \/ T =
+        fab match {
+          case -\/(a) => -\/(as(a))
+          case \/-(b) => \/-(bt(b))
+        }
     })
 
   implicit def disjunctionEq[A, B](implicit A: Eq[A], B: Eq[B]): Eq[A \/ B] =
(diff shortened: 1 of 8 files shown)


TomasMikula:
Cool stuff!

I was confused by the name "minimality checker" when in fact we don't care whether the implementation is minimal, but rather whether we have implemented enough methods. So "sufficiency checker"?

hrhino:
Yeah, that's better. Words are hard.

base/shared/src/main/scala/scalaz/data/DisjunctionInstances.scala
```diff
@@ -27,11 +27,13 @@ trait DisjunctionInstances {
     }
 
   implicit val disjunctionBifunctor: Bifunctor[Disjunction] =
-    instanceOf(new BifunctorClass[Disjunction] with BifunctorClass.DeriveLmapRmap[Disjunction] {
-      def bimap[A, B, S, T](fab: A \/ B)(as: A => S, bt: B => T): S \/ T = fab match {
-        case -\/(a) => -\/(as(a))
-        case \/-(b) => \/-(bt(b))
-      }
+    instanceOf(new BifunctorClass[Disjunction] {
+      val minimal = meta.IsMinimal()
```
alexknvl: Is it possible to find all `new X` such that `X` is annotated with at least one `@minimal`?

base/shared/src/main/scala/scalaz/typeclass/BifunctorClass.scala
```diff
@@ -15,32 +15,17 @@ package typeclass
  *       lmap(rmap(fab)(bt))(as) === rmap(lmap(fab)(as))(bt) === bimap(fab)(as, bt)
  *   ```
  */
-trait BifunctorClass[F[_, _]] {
+@meta.minimal("bimap", ("lmap", "rmap"))
```
alexknvl: :100:

alexknvl:
From Gitter:

Maybe `@instance` on a `val` or `def` instead of `val minimal = ...`. With a compiler plugin it would be possible to just find all instance declarations AFAIU.

`@meta.typeclass(minimal=("bimap", ("lmap", "rmap")))` and make it automatically insert `extends Typeclass`?

hrhino:
https://github.com/hrhino/plugin-poc is a version of this as a compiler plugin.

edmundnoble:
@hrhino Closing this for now in favor of the plugin.

Issue #1786, opened 2018-05-17
Improve Async in Scalaz 8 IO


jdegoes:
Async can benefit from 2 improvements:

 * Delete `Async.Later`, and implement `Async.later` by passing a no-op canceler to `Async.MaybeLater`. This will simplify the ADT and handling the different cases in the code.
 * Add a new constructor, `Async.MaybeLaterIO`, whose canceler has signature `Throwable => IO[Void, Unit]`. This represents a pure canceler. Add a `Async.maybeLaterIO` helper function, and support the new canceler inside `RTS` by submitting the run of the canceller to the thread pool.

After these two improvements, the ADT for `Async` will be less redundant, and will also have gained the ability to represent pure cancellations, which is a necessary property for "closure".

See `asyncIO` for similar idea applied to the `async` constructor.

barambani:
I would give this a try if that's ok

jdegoes:
@barambani That would be awesome! I am here to help with anything at all. 💪

PR #1795, opened 2018-05-19
Issue 1786 - Improve Async in Scalaz 8 IO

barambani:
Addresses this https://github.com/scalaz/scalaz/issues/1786

### DIFF
--- a/effect/jvm/src/main/scala/scalaz/effect/RTS.scala
+++ b/effect/jvm/src/main/scala/scalaz/effect/RTS.scala
@@ -5,8 +5,8 @@ import scala.annotation.switch
 import scala.annotation.tailrec
 import scala.concurrent.duration.Duration
 import java.util.concurrent.atomic.AtomicReference
-import java.util.concurrent.{Executors, TimeUnit}
-import java.lang.{Runnable, Runtime}
+import java.util.concurrent.{ Executors, TimeUnit }
+import java.lang.{ Runnable, Runtime }
 
 import scalaz.data.Disjunction._
 import scalaz.data.Maybe
@@ -128,7 +128,7 @@ trait RTS {
       }
     }
 
-  final def performed(canceler: PureCanceler): Canceler =
+  final def impureCanceler(canceler: PureCanceler): Canceler =
     th => unsafePerformIO(canceler(th))
 }
 
@@ -526,7 +526,7 @@ private object RTS {
                         case Async.MaybeLaterIO(pureCancel) =>
                           // As for the case above this stores an impure canceler
                           // obtained performing the pure canceler on the same thread
-                          awaitAsync(id, rts.performed(pureCancel))
+                          awaitAsync(id, rts.impureCanceler(pureCancel))
 
                           eval = false
                       }
@@ -815,7 +815,7 @@ private object RTS {
           case Async.MaybeLater(cancel) =>
             c1 = cancel
           case Async.MaybeLaterIO(pureCancel) =>
-            c1 = rts.performed(pureCancel)
+            c1 = rts.impureCanceler(pureCancel)
         }
 
         right.register(raceCallback[B, C](k, state, rightWins)) match {
@@ -824,7 +824,7 @@ private object RTS {
           case Async.MaybeLater(cancel) =>
             c2 = cancel
           case Async.MaybeLaterIO(pureCancel) =>
-            c2 = rts.performed(pureCancel)
+            c2 = rts.impureCanceler(pureCancel)
         }
 
         val canceler = combineCancelers(c1, c2)
--- a/effect/jvm/src/main/scala/scalaz/effect/RTS.scala
+++ b/effect/jvm/src/main/scala/scalaz/effect/RTS.scala
@@ -521,7 +521,9 @@ private object RTS {
 
                           eval = false
 
-                        case Async.Later() =>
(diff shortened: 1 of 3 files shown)


effect/shared/src/main/scala/scalaz/effect/Async.scala
```diff
@@ -1,29 +1,27 @@
...
- * or `maybeLater`, which represents an interruptible asynchronous action.
+ * `maybeLater`, which represents an interruptible asynchronous action or `maybeLaterIO`
+ * which represents an interruptible asynchronous action where the canceler has the
+ * form `Throwable => IO[Void, Unit]`
  */
 sealed abstract class Async[E, A]
 object Async {
-  type Interruptor = Throwable => Unit
+  type Canceler     = Throwable => Unit
+  type PureCanceler = Throwable => IO[Void, Unit]
```
jdegoes: These are good. Now create some `val` for these:

```scala
val Canceler: Canceler = _ => ()
val PureCanceler: PureCanceler = _ => IO.unit[Void]
```

effect/shared/src/main/scala/scalaz/effect/Async.scala
```diff
@@ -32,7 +30,8 @@ object Async {
    *
    * See `IO.async0` for more information.
    */
-  final def later[E, A]: Async[E, A] = Later()
+  final def later[E, A]: Async[E, A] =
+    MaybeLater(_ => ())
```
jdegoes: This will allocate a closure every single time, which will negatively affect performance in a measurable way.

To avoid this, replace with:

```scala
... = MaybeLater(Canceler)
```

effect/jvm/src/main/scala/scalaz/effect/RTS.scala
```diff
@@ -128,6 +128,9 @@ trait RTS {
         future.cancel(true); ()
       }
     }
+
+  final def submitCanceler[E, A](canceler: Throwable => IO[E, A]): Throwable => Unit =
+    th => submit(unsafePerformIO(canceler(th)))
```
jdegoes: After thinking about this more, I think we want `unsafePerformIO(canceler(th))`.

The reason is that: for an impure canceler, it's run in the same thread. I don't want that semantic to change merely because the canceler has been made pure. Ideally it should run in the same thread.

Also note that a canceler should always have `Void` as the error type so we know it cannot fail. So you should make that change here because we want a guarantee it won't fail on us (except for non-recoverable error / defect).

effect/shared/src/main/scala/scalaz/effect/Async.scala
barambani: added. Thanks 👍

effect/shared/src/main/scala/scalaz/effect/Async.scala
barambani: 👍 , done

effect/jvm/src/main/scala/scalaz/effect/RTS.scala
barambani: Got it, I changed it. I also changed the name as is not really submitting anymore. What do you think ? Thanks

effect/jvm/src/main/scala/scalaz/effect/RTS.scala
```diff
@@ -128,6 +127,9 @@ trait RTS {
         future.cancel(true); ()
       }
     }
+
+  final def performed(canceler: PureCanceler): Canceler =
```
jdegoes: I'd suggest calling this `impureCanceler` because it takes a `PureCanceler` and returns an impure one.

effect/shared/src/main/scala/scalaz/effect/Async.scala
```diff
@@ -32,7 +31,8 @@ object Async {
    *
    * See `IO.async0` for more information.
    */
-  final def later[E, A]: Async[E, A] = Later()
+  final def later[E, A]: Async[E, A] =
+    MaybeLater(NoOpCanceler)
```
jdegoes: I am sorry, we have to make one more change here, to eliminate the overhead of this common case:

```scala
final def later[E, A]: Async[E, A] = _Later.asInstanceOf[Async[E, A]]

private val _Later: Async[Nothing, Nothing] = MaybeLater(NoOpCanceler)
```

Otherwise, we have to incur the overhead of allocating `MaybeLater` every time `later` is called, which will noticeably affect performance.

effect/jvm/src/main/scala/scalaz/effect/RTS.scala
barambani: brilliant, changing it

effect/shared/src/main/scala/scalaz/effect/Async.scala
barambani: 👍  I'm doing it. Thanks

barambani: I was wondering if it shouldn't be done also for `maybeLater`, `maybeLaterIO` and `now` ?

jdegoes:
@barambani Beautiful work! I hope it's clear why this improves the expressiveness, does not decrease performance, and maintains the same semantics as the older version (across both impure and pure versions). Thanks for your great contribution! 😄

barambani:
Thank you for the help 👍

Issue #1798, opened 2018-05-19
Invert dependency between Scalaz 8 IO and Scalaz 8


jdegoes:
Scalaz 8 IO is integrated thoroughly into Scalaz 8, and the effect package depends on the core data types and type classes. However, it's possible and relatively straightforward to flip this dependency around, allowing Scalaz 8 IO to live in its own module, with Scalaz 8 and even Scalaz 7.3 depending on this module. This ticket outlines the steps involved in such a process.

 - [ ] Create a `scalaz-io` repository, or co-opt `ioeffect`.
 - [ ] Refactor Scalaz 8 IO to not use any core data types, instead relying on "church encoding" for things like `\/` and `Maybe`.
 - [ ] Move Scalaz 8 IO instances from `scalaz.effect` into core, and flip the dependency.
 - [ ] Put a code freeze on Scalaz 8 IO, and copy the Scalaz 8 IO code into `scalaz.io` (as per proposed package organization), and into the separate repository for `scalaz-io`.
 - [ ] Setup package publishing for Scalaz 8 IO.
 - [ ] Publish a first version of Scalaz 8 IO.
 - [ ] Delete the Scalaz 8 IO from Scalaz 8, and instead depend on the published `scalaz-io` package.
 - [ ] Delete the `IO` implementation and associated data types from Scalaz 7.3, and instead, depend directly on `scalaz-io`, adding the necessary instances in core.

fommil:
if we co-opt ioeffect we'll need to provide a 7.2 backcompat module with the orphan instances. Seems a good reason to make a 3.0 release.

What about the package name?

I like the thought of having access to IO in core. No excuses for impure code!

ktonga:
I could start with this one. Can someone create the new repo?

fommil:
@ktonga let's just use ioeffect. We can cut a 3.0.0 release.

ktonga:
I liked more `scalaz-io` it aligns better with the naming convention in the rest of the repos :(

jdegoes:
@fommil @ktonga I do prefer to use a new repo.

Ioeffect can stay for 7.2. We should rename it, of course.

ktonga:
Also in 7.2 we have to keep the package with a name other than `effect`
What you think about using `scalaz.io` as the base package in the new repo? it can be used across all versions. `effect` is too generic :)

fommil:
[...]

PR #1799, opened 2018-05-19
Add `IsCovariant` instance for `IO[E, ?]`. Fixes #1764

### LINKED ISSUE #1764: `IsCovariant` instances for `IO`
https://github.com/scalaz/scalaz/pull/1756#discussion_r187693960
It's okay to assert them.

kamilkloch:
Hello,

this a WIP attempt to resolve #1764. 
(Also, my first attempt to contribute to scalaz).
Is using `As.unsafeForce` the correct approach?

### DIFF
--- a/effect/shared/src/main/scala/scalaz/effect/IOInstances.scala
+++ b/effect/shared/src/main/scala/scalaz/effect/IOInstances.scala
@@ -2,7 +2,8 @@
 package scalaz
 package effect
 
-import ct._
+import scalaz.ct._
+import scalaz.types.IsCovariantClass
 
 trait IOInstances {
   implicit def monad[E]: Monad[IO[E, ?]] =
@@ -24,5 +25,7 @@ trait IOInstances {
       override def lmap[A, B, S](fab: IO[A, B])(as: A => S): IO[S, B] = fab.leftMap(as)
       override def rmap[A, B, T](fab: IO[A, B])(bt: B => T): IO[A, T] = fab.map(bt)
     })
+
+  implicit def isCovariant[E]: IsCovariant[IO[E, ?]] = IsCovariantClass.unsafeForce[IO[E, ?]]
 
 }
--- a/base/shared/src/main/scala/scalaz/types/IsCovariantClass.scala
+++ b/base/shared/src/main/scala/scalaz/types/IsCovariantClass.scala
@@ -22,6 +22,15 @@ trait IsCovariantClass[F[_]] {
 
 object IsCovariantClass {
 
+  /** Unsafe witness that the type constructor `F[_]` is covariant. */
+  def unsafeForce[F[_]]: IsCovariant[F] =
+    instanceOf(new IsCovariantClass.LiftLiskov[F] {
+      override def liftLiskov[A, B](implicit ev: A <~< B): F[A] <~< F[B] = {
+        val _ = ev
+        As.unsafeForce[F[A], F[B]]
+      }
+    })
+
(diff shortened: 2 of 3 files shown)


alexknvl:
I think it makes sense to add `IsCovariant.unsafeForce[F[_]]`. In this case it is the correct approach since `IO[E, A]` is covariant, but there is no way to *prove* it in Scala, so the only way is to assert it.

alexknvl:
In view of https://github.com/scalaz/scalaz/issues/1798, I am not 100% sure we want `IO` to depend on `IsCovariant`.

jdegoes:
@alexknvl It's true. We will turn this around so `scalaz` depends on `scalaz-io`. So the instances will move into core.

jdegoes:
@kamilkloch Nice work, and very ambitious first ticket. 😄 Thanks for your work on this.

jdegoes:
@alexknvl Good to merge?

kamilkloch:
@jdegoes Not yet, `IsCovariant` instance for `IO[?, A]` is missing. Also, one has to decide what happens with `IsCovariant` for `IO` after #1798 happens.

kamilkloch:
@alexknvl Pls take a look.

effect/shared/src/main/scala/scalaz/effect/IOInstances.scala
```diff
@@ -25,4 +26,8 @@ trait IOInstances {
       override def rmap[A, B, T](fab: IO[A, B])(bt: B => T): IO[A, T] = fab.map(bt)
     })
 
+  implicit def isCovariant[E]: IsCovariant[IO[E, ?]] = IsCovariantClass.unsafeForce[IO[E, ?]]
+
+  implicit def isLeftCovariant[A]: IsCovariant[IO[?, A]] = IsCovariantClass.unsafeForce[IO[?, A]]
```
alexknvl: Could you add a test `implicitly[IsCovariant[IO[?, Int]]]` to see if it actually resolves?

jdegoes: I think we can only have `IsCovariant` for `A`. 

For `E`, we need a newtype. Otherwise, the implicits will be ambiguous.

Also, a newtype can capture `Functor` instances for `E`, as well.

kamilkloch: I do not quite understand. The code below compiles. Should I add it to the PR?
```scala
class IOCovariance {
  implicitly[IsCovariant[IO[Int, ?]]]
  implicitly[IsCovariant[IO[?, Int]]]
}
```

alexknvl: Surprisingly, they are not ambiguous with partial unification (AFAIU it's enabled by default on Scalafiddle): https://scalafiddle.io/sf/KNhiag1/0

kamilkloch:
@alexknvl Pls take a look.

effect/shared/src/main/scala/scalaz/effect/IOInstances.scala
hrhino: It picks the one with the rightmost hole, it appears: https://scalafiddle.io/sf/14zY5XS/1

hrhino: Which is to say, I'm not sure how likely you are to be able to _use_ the instance for the left hole: https://scalafiddle.io/sf/3FNT07Y/0

jdegoes:
@kamilkloch Excellent work! Congratulations, you're a Scalaz 8 contributor. 😉

Issue #1305, opened 2016-12-13
Change order of type parameters to monad transformers?


TomasMikula:
I suppose there was a reason why type parameters of some monad transformers in scalaz are in different order than in Haskell. For example

```scala
StateT[F[_], S, A]
```

instead of

```scala
StateT[S, F[_], A]
```

The latter seems to be a more common order of partial application.

Whatever the reason was, is it still valid?

Now that there's partial unification of type arguments, should we change the type parameter order to the one that's more common?

At least the following types could benefit from different parameter order:
 - [ ] `StateT`
 - [ ] `WriterT`
 - [ ] `ReaderT`
 - [ ] `ReaderWriterStateT`
 - [ ] `EitherT`
 - [ ] `ContT`

xuwei-k:
@retronym wrote 5 years ago https://github.com/scalaz/scalaz/commit/4be176876edfde66b4018ed598c887eb1011f658#diff-f8a29528927f2b43e168ec1b8fede81fR239

> Declare type parameters requiring type constructors first in type parameter declarations. This is and arbitrary choice,
> but the consistency is worth it.
>
>     class Foo[F[_], A, B] // good
>     class Foo[A, F[_], B] // bad

https://github.com/scalaz/scalaz/blob/v7.3.0-M7/CONTRIBUTING.md#type-parameters

TomasMikula:
I see.  Back then (before `-Ypartial-unification`), I guess optimizing for the most common use case would have made no difference. Might be time to reconsider?

xuwei-k:
> Might be time to reconsider?

👍

S11001001:
We can't really reconsider until we no longer support versions of Scala that *don't* have `-Ypartial-unification` or an equivalent substitute (like the plugin) available.  For the 7.3.x series, we still support Scala 2.10.

I've tagged this for 8 because there are good design cleanup reasons to require `-Ypartial-unification` be turned on for all users in that release.

jbgi:
>  For the 7.3.x series, we still support Scala 2.10.

not anymore: https://github.com/scalaz/scalaz/pull/1367.

TomasMikula:
See #1800 and #1838.

jozi-k:
@TomasMikula I went through list of monad transformers and type parameters are now in correct order. I believe you can close this issue.

TomasMikula:
Thanks for checking, @jozi-k! Closing.

PR #1800, opened 2018-05-20
Monad transformers switch type

JordanMartinez:
Addresses #1305 

Note:
- Did not do ReaderWriterStateT due to not knowing where the `F[_]` should go in the type signature (e.g. `RWST[R, F, W, S, A]` or `RWST[R, W, S, F, A]`? I'm assuming the latter...?)
- Not sure whether the swap for StateT and ContT are supposed to go deeper (e.g. also do swap in `IndexedStateT`)
- Some cases of using `StateT[S, F, A]` would not compile unless type was changed to that which it aliases: `IndexedStateT[F, S, S, A]`
- Not sure whether methods types or other classes should also be changed (e.g. `method[F[_], A]` stays the same? or becomes `method[A, F[_]]`? / `MonadPlus[F[_], A]` or `MonadPlus[A, F[_]]`)
- "Fix Later" commit: code wasn't compiling and I'm not sure how to fix it. Figured it'd be better to get more of this PR done and then figure that out with everyone else's help.

I'm going to sleep after this, so won't respond until Monday.

### DIFF
--- a/core/src/main/scala/scalaz/EitherT.scala
+++ b/core/src/main/scala/scalaz/EitherT.scala
@@ -13,7 +13,7 @@ import Liskov.<~<
  * EitherT(x).map(1+).run // Some(\/-(2))
  * }}}
  * */
-final case class EitherT[F[_], A, B](run: F[A \/ B]) {
+final case class EitherT[A, F[_], B](run: F[A \/ B]) {
   import OptionT._
 
   final class Switching_\/[X](r: => X) {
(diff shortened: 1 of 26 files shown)


TomasMikula:
> - Did not do ReaderWriterStateT due to not knowing where the `F[_]` should go in the type signature (e.g. `RWST[R, F, W, S, A]` or `RWST[R, W, S, F, A]`? I'm assuming the latter...?)

Yes, the latter. (I wouldn't follow Haskell religiously, but it cannot hurt to check what has worked there: [The RWST monad transformer](https://hackage.haskell.org/package/transformers-0.5.5.0/docs/Control-Monad-Trans-RWS-Lazy.html#g:2).)

> - Not sure whether the swap for StateT and ContT are supposed to go deeper (e.g. also do swap in `IndexedStateT`)

I would say yes, if only for consistency.

> - Some cases of using `StateT[S, F, A]` would not compile unless type was changed to that which it aliases: `IndexedStateT[F, S, S, A]`

This should go away after it is changed to `IndexedStateT[S, S, F, A]`.

> - Not sure whether methods types or other classes should also be changed (e.g. `method[F[_], A]` stays the same? or becomes `method[A, F[_]]`? / `MonadPlus[F[_], A]` or `MonadPlus[A, F[_]]`)

I would change method type parameter order accordingly, but I don't think it would apply to these examples.

---

We should also remove the following from [CONTRIBUTING.md](https://github.com/scalaz/scalaz/blob/series/7.3.x/CONTRIBUTING.md):

> Declare type parameters requiring type constructors first in type parameter declarations. This is an arbitrary choice,
but the consistency is worth it.
>
>     class Foo[F[_], A, B] // good
>     class Foo[A, F[_], B] // bad

JordanMartinez:
> We should also remove the following
> > Declare type parameters requiring type constructors first in type parameter declarations. This is an arbitrary choice,
> > but the consistency is worth it.
> >     class Foo[F[_], A, B] // good
> >     class Foo[A, F[_], B] // bad

Should this not necessarily be removed entirely but just simply changed to say something else. For example, we could change it to read "The higher-kinded type (i.e. `F[_]` / `G[_]`) should always go before the thing it enapsulates (i.e. `A` / `B`), assuming such a thing is included in the method/class"? If so, these examples would be correct:
````scala
def method0[F](fa: F[_]): F[_] = ???
def method1[F, A](fa: F[A]): F[A] = ???
def method2[F, A, G, B](fa: F[A], f: A => G[B]): G[B] = ???

case class Clazz0[F[_]]
case class Clazz1[F[_], A]
case class Clazz2[F[_], A, G[_]]
````
I'm asking partly because I do not fully understand yet what is gained by switching these type parameters, nor what problems might arise with the above examples

I'll address your other comments later. Hopefully by the end of the day.

**Comment by TomasMikula** [comment]
> I do not fully understand yet what is gained by switching these type parameters

It is just optimizing for the most common use case in conjunction with partial unification. It's just more common in practice that one needs to infer `StateT[S, ?[_], ?]` than `StateT[?, F, ?]`, so we choose the order of type parameters such that the more common use case can be inferred using partial unification, which works from the right.

```scala
val foo: StateT[S, F, A] = ???
def bar[MT[_[_], _], G[_], B](mtgb: MT[G, B]): Unit = ???

bar(foo) // trying to unify MT[G, B] with StateT[S, F, A],
         // B is unified with A, G is unified with F,
         // and MT is unified with StateT[S, ?[_], ?]
```

JordanMartinez:
Thanks. I also read through scala/scala#5102 and that explained a bit more context, too.

JordanMartinez:
For ContT's type alias hierarchy...
````scala
// changes to ContT[R, Id, A]
type Cont[R, A] = ContT[Id, R, A] 

// does this change to ContsT[R, Id, M, A] ???
type ContT[M[_], R, A] = ContsT[Id, M, R, A]

// Does this change to IndexedContsT[R, W, M, R, A] ???
type ContsT[W[_], M[_], R, A] = IndexedContsT[W, M, R, R, A]
````

**Comment by JordanMartinez** [comment]
Note: the current code does not make the corresponding swap in the test package. I'll need to update it for that, too.

**Comment by TomasMikula** [comment]
As in other monad transformers, make the second to last type parameter be the monad `M[_]`.

As for the comonad type parameter `W[_]`, I wouldn't change its position in this PR, so keep it as the first type parameter for now. (Unless someone has a compelling use case for it, we might even remove the `W[_]` parameter altogether in the future; see #1737.)

**Comment by JordanMartinez** [comment]
> As in other monad transformers, make the second to last type parameter be the monad M[_].

Looks like I'll need to fix `RWST` since it puts `F[_] in-between `S1` and `S2`

**Comment by JordanMartinez** [comment]
Oh, nvm. I did it correctly the first time.

**Comment by JordanMartinez** [comment]
In `EndomorphicTest`
````scala
implicit def endoEqual[F[_], G[_[_], _, _], A](
    implicit F: Equal[G[F, A, A]]
  ): Equal[Endomorphic[G[F, ?, ?], A]] =
    Equal.equalBy(_.run)

checkAll(monoid.laws[Endomorphic[Kleisli[?, Option, ?], Int]])
checkAll(semigroup.laws[Endomorphic[Cokleisli[Option, ?, ?], Int]])
````
It fails at `monoidLaws` because it expected `G[F, ?, ?]`, not `G[?, F, ?]`. I guess this means I have to swap `Cokleisli`'s types as well...?

**Comment by JordanMartinez** [comment]
`Kleisli`'s tests are failing because it can't figure out that it needs to wrap it in a class to get some methods that seem related to `Arrow`:
````
>>>
|||
-*-
mapsnd
first
proleft
***
````

**Comment by TomasMikula** [comment]
I suggest to leave `Kleisli` unchanged in this PR, and define `ReaderT` as

```scala
type ReaderT[A, F[_], B] = Kleisli[F, A, B]
```

If that causes problems, leave out `ReaderT` for now as well. (In the future we might want to hide the fact that `ReaderT` is an alias for `Kleisli`.)

It is in fact the case that when we talk about `Kleisli` arrows, we have a monad `F[_]` in mind before talking about the source and target objects `A`, `B`.

On the other hand, when we use the `ReaderT` monad transformer, we typically know the input type `A` and allow the additional effect monad `F[_]` to be specified later.

This is also how things are in Haskell ([Kleisli](https://hackage.haskell.org/package/base-4.11.1.0/docs/Control-Arrow.html#t:Kleisli), [ReaderT](http://hackage.haskell.org/package/transformers-0.5.5.0/docs/Control-Monad-Trans-Reader.html#t:ReaderT)):

```haskell
Kleisli m a b
ReaderT r m a
```

JordanMartinez:
Alright, I'll try that out and see how it goes.

Also, I'm still not sure why `EitherT` is failing in the example usage:
````
[error] /scalaz/example/src/main/scala/scalaz/example/UnapplyInference.scala:13:22: 
          value :-> is not a member of scalaz.EitherT[Int,Option,Int]
[error]     println((eitherT :-> (_ - 1)).run) // Some(Right(0))
[error]                      ^
[error] /scalaz/example/src/main/scala/scalaz/example/UnapplyInference.scala:24:63: 
          value bisequence is not a member of scalaz.EitherT[List[Int],Option,List[Int]]
[error]     val bisequence: List[EitherT[Int, Option, Int]] = eitherT.bisequence[List, Int, Int]
[error]

**Comment by JordanMartinez** [comment]
I'll push the EitherT type swap commit (cherrypick) once I see how this does since it is known that the EitherT commit has compilation issues.

**Comment by JordanMartinez** [comment]
Aye. The code compiles, but some test is looping infinitely and causing the build to time out.

**Comment by JordanMartinez** [comment]
I've done my part. Now I need help from the community: 
- [ ] Figure out why compiler can't find EitherT's implicit Bitraverse/Bifunctor instances
- [ ] Figure out why the tests stop running after `Coproduct.traverse1`'s second property test
- [ ] Feedback on PR (e.g. whether type parameters are in correct places for methods and whatnot)

**Comment by virdis** [comment]
@JordanMartinez I think this PR should be open against scalaz8 branch, that could explain the failures you have listed above.

**Comment by JordanMartinez** [comment]
So... The [last comment](https://github.com/scalaz/scalaz/issues/1305) that implies this should be done for the `7.3.x` series in #1305 was a mistake? If so... that really sucks, haha!

**Comment by jdegoes** [comment]
@JordanMartinez Thanks for making it this far! Folks will chime in as they have time, and I'll do a review of the type parameter ordering when I get a chance.

**Comment by jdegoes** [comment]
@JordanMartinez The order looks correct to me! I did not check that all type classes were so modified, but the ones in this pull request look good. 👍

**Comment by JordanMartinez** [comment]
@jdegoes Thanks for the feedback!

I looked into the tests not responding and have narrowed the problem down to the StateT commit. When I tested that commit, the tests stopped responding just like the build here. So, I cherrypicked the WriterT commit on top of the HEAD of the `7.3.x` series when I initially started this PR and ran the tests then. They passed without issue. I did the same cherrypick-then-test method for the remaining commits in the following order and they all passed: ReaderWriterStateT, ContT, ReaderT, EitherT

I'm looking into why the EitherT example isn't compiling next.

**Comment by TomasMikula** [comment]
The problem with `EitherT` comes down to whether we will primarily use it as a monad transformer or as a bifunctor. This PR chooses the monad transformer use case, i.e. parameter order `EitherT[A, M[_], B]`. Such `EitherT[A, M[_], B]` will not unify with `F[X, Y]` (where we would like to infer `F` := `EitherT[?, M, ?]`, `X` := `A`, `Y` := `B`) out of the box.

However, we can instruct the compiler to consider this possibility by adding a new case of `Unapply2` inference to `object Unapply2` to adapt types `Foo[A, M[_], B]` to shape `Bar[A, B]`:
```scala
implicit def unapplyMAFB[TC[_[_, _]], M0[_, _[_], _], A0, F0[_], B0](implicit TC0: TC[M0[?, F0, ?]]): Unapply2[TC, M0[A0, F0, B0]] {
  type M[X, Y] = M0[X, F0, Y]
  type A = A0
  type B = B0
} =
  new Unapply2[TC, M0[A0, F0, B0]] {
    type M[X, Y] = M0[X, F0, Y]
    type A = A0
    type B = B0
    def TC = TC0
    def leibniz = refl
  }
```

JordanMartinez:
I can't get to this now, but hopefully will before the end of today. Thanks for the explanation @TomasMikula!

JordanMartinez:
If this PR is merged now, the `StateT` commit can be merged in a different PR. However, the issue of what CONTRIBUTING.MD says hasn't been resolved. I wonder if that should be further discussed in the original issue that this PR addresses.

TomasMikula:
> However, the issue of what CONTRIBUTING.MD says hasn't been resolved. I wonder if that should be further discussed in the original issue that this PR addresses.

Or just open a PR removing it and discuss there.

JordanMartinez:
> Or just open a PR removing it and discuss there.

See #1841.

JordanMartinez:
Don't merge this yet. Similar to the `StateT` Arbitrary comment in #1838, the type parameters are not swapped for the Arbitrary's of EitherT, etc.

JordanMartinez:
After checking this again, only the Abritraries of `EitherT`, `IndexedContsT`, `WriterT`, and `UnwriterT`'s type params weren't swapped yet. The others were swapped already in their original commit. (`IndexedConstT` was swapped in the `WriterT`/`UnwriterT` Arbitrary commit, but I forgot to uncheck that line when doing a partial commit in IntelliJ)

JordanMartinez:
Not sure whether this is waiting for other members' approval, or the core contributors are just busy, or it is believed that this PR isn't ready for merging due to my prior comment. So, just to clarify, the final issue has been resolved and I believe this is ready for merging/review from others (if that's what's needed).

jdegoes:
@jonm Looks good! Thank you Jordan!

PR #1803, opened 2018-05-20
Delete version

saczawap:
At the end I wanted to set default version to 8.0.0 which overrides sbt-dynber. I assume that the only way to set it is creating a tag like "8.0.0-SNAPSHOT"
@jdegoes @NeQuissimus

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,8 +2,6 @@ import Scalaz._
 
 organization in ThisBuild := "org.scalaz"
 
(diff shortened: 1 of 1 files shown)


jdegoes:
@NeQuissimus Good to merge?

saczawap:
actually tag `8.0.0-SNAPSHOT` will produce versions like `8.0.0-SNAPSHOT-8.0.0-SNAPSHOT%2B2-9e540007-SNAPSHOT`

PR #1812, opened 2018-05-23
Remove redundant implicits and unnecessary prioritization

aaronvargo:


jbgi:
Could you (shortly) explain why the removed implicits def were redundant?

aaronvargo:
Transitivity works. E.g. Bind => Apply and Apply => Functor is enough to get Bind => Functor

jbgi:
Right, I slightly worry of the performance impact of the additional indirections, but its definitely cleaner.
Let me a small benchmark on your branch to try measure the impact.

jbgi:
@jdegoes You will have to update a slide once this is merged! :smile:

jdegoes:
@jbgi Lol. 😆 

@aaronvargo Nice work, and thank you for the contribution!

edmundnoble:
@jbgi can you post the benchmarks? @aaronvargo can you resolve conflicts?

aaronvargo:
I've resolved the conflicts and also added some implicits that were missing:

```scala
Functor[F] => InvariantFunctor[F]
Bifunctor[F] => Functor[F[A, ?]]
Profunctor[F] => Functor[F[A, ?]]
```

Yes, the latter two work. This is one of the advantages of the new typeclass encoding. I also deleted the `.rmap` syntax, since you can now just use `.map`.

base/shared/src/main/scala/scalaz/ct/profunctor.scala
```diff
@@ -7,6 +7,10 @@ trait ProfunctorClass[F[_, _]] {
   def lmap[A, B, C](fab: F[A, B])(ca: C => A): F[C, B]
   def rmap[A, B, C](fab: F[A, B])(bc: B => C): F[A, C]
   def dimap[A, B, C, D](fab: F[A, B])(ca: C => A)(bd: B => D): F[C, D]
+  def functor[A]: Functor[F[A, ?]] =
+    instanceOf(new FunctorClass[F[A, ?]] {
+      def map[B, C](fab: F[A, B])(f: B => C): F[A, C] = rmap(fab)(f)
+    })
```
hrhino: not sure exactly how much I like having this as an instance method.

base/shared/src/main/scala/scalaz/ct/profunctor.scala
```diff
@@ -7,6 +7,10 @@ trait ProfunctorClass[F[_, _]] {
   def lmap[A, B, C](fab: F[A, B])(ca: C => A): F[C, B]
```
hrhino: can you rename this to `lcontramap` or something while you're at it?

aaronvargo: Sure

base/shared/src/main/scala/scalaz/ct/profunctor.scala
aaronvargo: It should probably be moved into a RightFunctor class, which will be a superclass of both Profunctor and Bifunctor

hrhino:
@edmundnoble/@emilypi/@jbgi do you have opinions on the `{Bi,Pro}Functor => Functor` change? otherwise I think this is good to go

jbgi:
@edmundnoble the benchmarks I run was:
```scala
  @Benchmark
  def scalazMap(): BigInt = {
    @tailrec
    def sumTo(t: IO[Void, BigInt], n: Int): IO[Void, BigInt] =
      if (n <= 1) t
      else sumTo(t.map(_ + n), n - 1)

    unsafePerformIO(sumTo(IO.point(0), depth))
  }
```
```scala
  @Benchmark
  def scalazMapMonad(): BigInt = {
    @tailrec
    def sumTo(t: IO[Void, BigInt], n: Int): IO[Void, BigInt] =
      if (n <= 1) t
      else sumTo(Monad[IO[Void, ?]].map(t)(_ + n), n - 1)

    unsafePerformIO(sumTo(IO.point(0), depth))
  }
```
```scala
  @Benchmark
  def scalazMapFunctor(): BigInt = {
    @tailrec
    def sumTo(t: IO[Void, BigInt], n: Int): IO[Void, BigInt] =
      if (n <= 1) t
      else sumTo(Functor[IO[Void, ?]].map(t)(_ + n), n - 1)

    unsafePerformIO(sumTo(IO.point(0), depth))
  }
```
```
Benchmark                        (depth)   Mode  Cnt      Score      Error  Units
IOMapBenchmark.scalazMap             500  thrpt   30  69336.037 ± 4351.798  ops/s
IOMapBenchmark.scalazMapFunctor      500  thrpt   30  65247.316 ±  324.710  ops/s
IOMapBenchmark.scalazMapMonad        500  thrpt   30  64652.068 ±  625.719  ops/s
```

base/shared/src/main/scala/scalaz/ct/profunctor.scala
jbgi: for this PR, I think I'd prefer have this implementation directly in `BaseHierarchy` and this method removed from `[Pro/Bi]functor`.

aaronvargo: done

jbgi:
lgtm!

PR #1814, opened 2018-05-24
Add `*Functions` traits for all typeclasses, mixed into `Prelude`.

hrhino:
No exceptions here. I know at least a few people said they would be interested in this.

Methods like `pure`, which do not allow for `F` to be inferred from the argument types, also get a `*0` variant that uses the "partially applied" hack. A few methods are also exported as natural transformations, named `*NT`, as I saw fit.

`*Functions` traits have been set up for all of the subpackages that contain typeclasses, and *those* now get mixed into `Prelude`, to keep it shorter.

I have used `F` and `A` as type parameters throughout, irrespective of the type parameters on the actual typeclass. We should sort that out.


base/shared/src/main/scala/scalaz/Prelude.scala
```diff
@@ -119,13 +119,12 @@ trait BaseDataAliases { self: BaseData =>
...
 
 trait AllFunctions
-    extends data.VoidFunctions
-    with data.DisjunctionFunctions
-    with data.MaybeFunctions
-    with data.KleisliFunctions
-    with ct.InvariantFunctorFunctions
-    with ct.PhantomFunctions
-    with ct.TraversableFunctions
+    extends AnyRef
```
NeQuissimus: What's the benefit of extending `AnyRef` here?

base/shared/src/main/scala/scalaz/algebra/MonoidFunctions.scala
```diff
@@ -0,0 +1,7 @@
+package scalaz
+package algebra
+
+trait MonoidFunctions {
+  @inline final def empty[A](implicit A: Monoid[A]): A =
```
NeQuissimus: See the build failure :)

base/shared/src/main/scala/scalaz/Prelude.scala
hrhino: If we add a new package that is alphabetically before `algebra`, we don't need to change the line `algebra.AlgebraFunctions` to add it here.

base/shared/src/main/scala/scalaz/algebra/MonoidFunctions.scala
hrhino: yep. I'll propose `mempty` and `mappend`.

base/shared/src/main/scala/scalaz/ct/ProfunctorFunctions.scala
```diff
@@ -0,0 +1,17 @@
...
+package ct
+
+trait ProfunctorFunctions {
+  @inline final def dimap[F[_, _], A, B, C, D](fab: F[A, B])(
+    ca: C => A,
+    bd: B => D,
+  )(implicit F: Profunctor[F]): F[C, D] =
+    F.dimap(fab)(ca)(bd)
+
+  /* TODO: these clash with the bifunctor versions. Decide on a name for them.
```
jdegoes: contramap / comap
inMap / outMap
mapIn / mapOut

base/shared/src/main/scala/scalaz/ct/ChoiceFunctions.scala
```diff
@@ -0,0 +1,9 @@
+package scalaz
+package ct
+
+trait ChoiceFunctions {
+  @inline final def leftchoice[F[_, _], A, B, C](fab: F[A, B])(implicit F: Choice[F]): F[A \/ C, B \/ C] =
```
jdegoes: Capital `c` in `choice`.

base/shared/src/main/scala/scalaz/ct/BifunctorFunctions.scala
```diff
@@ -0,0 +1,14 @@
+package scalaz
+package ct
+
+trait BifunctorFunctions {
+  @inline final def bimap[F[_, _], A, B, S, T](fab: F[A, B])(
+    as: A => S,
+    bt: B => T
+  )(implicit F: Bifunctor[F]): F[S, T] =
+    F.bimap(fab)(as, bt)
+  @inline final def lmap[F[_, _], A, B, S](fab: F[A, B])(as: A => S)(implicit F: Bifunctor[F]): F[S, B] =
```
jdegoes: Another option is to call these `firstMap`, `secondMap` after `Strong` (first/second).

base/shared/src/main/scala/scalaz/ct/ApplicativeFunctions.scala
```diff
@@ -0,0 +1,20 @@
+package scalaz
+package ct
+
+import data.~>
+
+trait ApplicativeFunctions {
+  @inline final def pure[F[_], A](a: A)(implicit F: Applicative[F]): F[A] =
```
jdegoes: `pure` vs `point`. Decisions, decisions.

CONTRIBUTING.md
```diff
@@ -18,4 +18,16 @@ Type classes
...
+     - Each typeclass `Foo` should have the following traits defined:
+        - `FooClass`: The definition of the typeclass.
+        - `FooSyntax`: Syntax macros for the typeclass methods.
+        - `FooFunctions`: A trait containing the typeclass's methods, but with
+                          the typeclass as a type parameter.
+        - `FooInstances`: Instances which are associated with the typeclass and
+                          have nowhere better to go. This should be made a
+                          superclass of `FooClass`'s companion object, to ensure
+                          that the instances are in the correct implicit scope.
+     - Note that this means that all typeclass methods must have unique names!
```
jdegoes: Very important point.

jdegoes:
@hrhino Some much needed changes here! 🙏

hrhino:
Sounds good. I'll make those changes and rebase once #1625 lands.

CONTRIBUTING.md
```diff
@@ -18,4 +18,16 @@ Type classes
  - Hierarchy
 	 - For each trait of the hierarchy, a particular typeclass must appear only once as return type in one hierarchy level (`BH1`, `BH2`, ...).
 	 - Companion objects should only declare implicit instances for the type classes at the top of a hierarchy. 
-
+ - Classes
+     - Each typeclass `Foo` should have the following traits defined:
+        - `FooClass`: The definition of the typeclass.
+        - `FooSyntax`: Syntax macros for the typeclass methods.
+        - `FooFunctions`: A trait containing the typeclass's methods, but with
+                          the typeclass as a type parameter.
```
alexknvl: as an implicit parameter?

hrhino: Thanks. I originally had a confusing attempt to say "with the typeclass instance as a type parameter" or something but I trashed it as insufficiently clear.

jdegoes:
@hrhino Landed!

PR #1817, opened 2018-05-25
A quick and dirty Tree.

Milyardo:


example/src/main/tut/data/Free.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: docs
+title: "Free"
+---
+
+# Free
+
+Free is a recursive data structure that either returns a vale `A` or returns an effect computation in `F[_]` that yields
```
hrhino: "value", "effectful".

That said, I think this makes `Free` sound like it's limited to representing computations. It's also just an "`F`-branching tree", which this doesn't make clear.

example/src/main/tut/data/Free.md
```diff
@@ -0,0 +1,25 @@
...
+another `Free`. Free can be summarized with following signature:
+
+`type Free[F[_], A] = A \/ F[Free[F,A]]`
```
hrhino: I'd say "`Free` is defined as:" and use a code block rather than inline quote.

example/src/main/tut/data/Free.md
```diff
@@ -0,0 +1,25 @@
...
+
+Free gets it's namesake from it's Monad instance which can be used to derive a monad from any arbitrary `Functor`.
```
hrhino: "gets its name from its"...

"any" and "arbitrary" are redundant together.

Maybe add a note that this _doesn't_ have the built-in `Coyoneda` that 7.x and cats have.

example/src/main/tut/data/Free.md
```diff
@@ -0,0 +1,25 @@
...
+Free can also be used as fix point defining recursive data structures.
+
+## Basics
+
+```tut:silent
+import scalaz._, Scalaz._
+import data._
+
+val pureFree = Free.pure[Maybe, Int](1) //Create a Free from a pure value using Maybe.
```
hrhino: These shouldn't be in a `tut:silent` block IMO.

base/shared/src/main/scala/scalaz/data/Free.scala
```diff
@@ -0,0 +1,36 @@
...
+trait FreeInstances {
+  import Free.{ unwrap, wrap }
+
+  implicit def freeMonad[F[_]](implicit F: Functor[F]): Monad[Free[F, ?]] =
+    instanceOf(
+      new MonadClass[Free[F, ?]] with MonadClass.DeriveMap[Free[F, ?]] with BindClass.DeriveFlatten[Free[F, ?]]
+      with BindClass.DeriveAp[Free[F, ?]] {
+        def pure[A](a: A): Free[F, A] = Free.pure(a)
+        def flatMap[A, B](fa: Free[F, A])(f: A => Free[F, B]): Free[F, B] =
+          unwrap(fa).fold(a => f(a))(ffa => wrap(\/-(F.map(ffa)(fa2 => flatMap(fa2)(f)))))
```
hrhino: deriving `map` using this is horridly inefficient; I'd just write it out.

emilypi:
@Milyardo I closed a PR recently which details some of the progression regarding where we want to head in terms of Free and Cofree - see here: https://github.com/scalaz/scalaz/pull/1613

The encoding you used was almost exactly the same code I first proposed in the history of that PR, but due to its inefficiency, isn't really viable as the `Free` we'd like to use. Another encoding was proposed by @TomasMikula which was essentially `data Free f a = Pure a | LiftF f a | Impure (f e) (e -> f a)`, which was good, but we also proposed an alternative encoding which effectively stored the flatmaps in a type-aligned stack-ended queue for additional speedup (i.e. making use of `ACatenable`). You can free free to rip out any portions of code you like. the `Freer` implementation (i.e. the `ACatenable`-based solution) wasn't stack-safe yet, but the `Free` implementation was basically finished.

I have not had the time, nor the desire to finish this PR as I make ramp down the amount of Scala i'm doing in favor of Haskell, but we're free anytime in the Gitter to discuss potential implementations.

jdegoes:
`ACatenable` sounds like the perfect structure for this.

hrhino:
What about just `Tree` or `LeafyTree` or `GTree` (for "generalized tree")? If `F` is a functor, `Tree[F, A]` is a freely-generated monad, but it's also just a tree.

Milyardo:
I'm all for renaming this, it's great for trees and also great as a fix point in recursion scheme. I'm leaning to calling it just `Tree` as well unless someone has any objections.

edmundnoble:
The problem with that is that actually *any* fixpoint structure without functions inside can be considered a tree. Specifically, trees with Free are sorts of "leafy trees", i.e. trees with data at leaves (of type `A`, anyway; the branching `F[_]` context can contain data as well).

edmundnoble:
@alexknvl and I talked about this in the scalaz/scalaz gitter; what do you think about `FixFree` @Milyardo?

Milyardo:
Sounds good to me.

emilypi:
@Milyardo you still working on this? I'm good to merge if you make the changes suggested

emilypi:
@Milyardo looks like you need to run fmt and we're good.

PR #1818, opened 2018-05-25
use scalaz.Void instead of java.lang.Void

KadekM:


jdegoes:
@KadekM Thanks, & congrats on your first contribution. 🙏 😄

PR #1820, opened 2018-05-26
Move data and other things

saczawap:
https://github.com/scalaz/scalaz/issues/1540
I wonder where these thing should go:
- forall
- identity
- const
- these
- upstar
- downstar
- void
I assume that there should be dependence hierachy like `core` -> `types` -> `ct` -> `data`
What about std?

Based on the issue everything that goes into `effect` now should be in `io`.


jdegoes:
I'd move `Const` and `These` to data, perhaps. Not sure about the others. /cc @alexknvl

jdegoes:
We should place things in such a location as dependencies are _one-way only_.

alexknvl:
By instances I meant `instance : Void`. I guess a better term is *the number of inhabitants*. So `Void` is 0, `Unit` is one, `(A, B)` = A * B and `A \/ B` = A + B.

jdegoes:
@saczawap Thanks for this ongoing work! It looks like there's one conflict and it seems we're all agreed `Void` should go into `core` (though it's not clear to me if you were asking these questions for your next pull request, or for this one).

saczawap:
So it looks like a pattern emerges: `core` depends on `types` and `algebra`, `data` depends on all three and `ct` depends also on `data` which, in my mind, is a little bit surprising, I was expecting that `data` would depend on `ct`.

jdegoes:
> So it looks like a pattern emerges: core depends on types and algebra, data depends on all three and ct depends also on data which, in my mind, is a little bit surprising, I was expecting that data would depend on ct.

Ok, we need to make sure that `ct` only depends on `data`, and not the other way around, which means that instances for `data` must be defined in `ct`, not `data`.

Then we can document this dependency graph and be sure to maintain it. Or maybe @fommil knows some way to enforce acyclic package dependencies.

fommil:
https://github.com/lihaoyi/acyclic

PR #1824, opened 2018-05-27
removes no longer needed @silents due to scala#10623

KadekM:


jdegoes:
@KadekM Thank you Marek!

PR #1829, opened 2018-05-29
Include scheduler in app shutdown

ktonga:


effect/jvm/src/main/scala/scalaz/effect/RTS.scala
```diff
@@ -72,6 +72,8 @@ trait RTS {
   }
 
   final def unsafeShutdownAndWait(timeout: Duration): Unit = {
+    scheduledExecutor.shutdown()
+    scheduledExecutor.awaitTermination(timeout.toMillis, TimeUnit.MILLISECONDS)
```
ktonga: @jdegoes do you prefer to have more control over this?
Maybe something like:

```..., scheduledTimeout: Option[Duration]...```

And we `shutdowndNow` for `None`

jdegoes: I think just 1 timeout is fine, we can use it for both (ideally in parallel but then we need threads for that 😆 ). `None` may be sensible.

ktonga: The semantics of this second TO would be `None`: shot it in the head, and `Some(to)`: wait for scheduled jobs to start, then wait for them to finish (they run in `threadPool`).

But if you think that's too much and the PR is ok as it is now I'm ok with that too :smiley:

jdegoes: Yeah, I think it's too much and also non-obvious, since I would expect `None` to wait forever. 😄

PR #1830, opened 2018-05-30
Validation as newtype over disjunction for scalaz 8

Jacoby6000:
There may be things wrong with this, but I'm not sure... I don't see where tests are written for other things, but I'd be happy to write some for this if people could point me in the right direction.

To write this, I basically just followed the patterns that the `Maybe` newtype is using.  While doing that, I noticed that the unapply methods on `Maybe` (#1832) will not get proper exhaustivity warnings.  There may have been discussion over this already, but I'm not aware.  Just want to make sure people are aware.   Also, because I copied `Maybe`, the `Validation` implementation has the same issue.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/validation.scala
+++ b/base/shared/src/main/scala/scalaz/data/validation.scala
@@ -17,13 +17,8 @@ trait ValidationModule {
     def unapply[A, B](vab: Validation[A, B]): Option[B] = fold(vab)(_ => None, Some(_))
   }
 
-  final def fold[A, B, C](vab: Validation[A, B])(ia: A => C, vb: B => C): C = vab match {
-    case Invalid(a) => ia(a)
-    case Valid(b)   => vb(b)
-  }
-
   /* functions */
-
+  def fold[A, B, C](vab: Validation[A, B])(ia: A => C, vb: B => C): C
   def fromDisjunction[A, B](oa: A \/ B): Validation[A, B]
   def invalid[L, R](value: L): Validation[L, R]
   def valid[L, R](value: R): Validation[L, R]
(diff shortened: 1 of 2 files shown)


base/shared/src/main/scala/scalaz/data/validation.scala
```diff
@@ -0,0 +1,113 @@
...
+    case Invalid(a) => ia(a)
+    case Valid(b)   => vb(b)
+  }
+
+  @inline def fromDisjunction[L, R](disj: L \/ R): Validation[L, R] = disj
+
+  @inline def invalid[L, R](value: L): Validation[L, R] = -\/(value)
+  @inline def valid[L, R](value: R): Validation[L, R]   = \/-(value)
+
+  implicit def validationApplicative[L](implicit L: Semigroup[L]): Applicative[Validation[L, ?]] =
```
hrhino: I think these instances aren't going to get resolved where they are. You need a (`ValidationInstances` mixed into a) `ValidationModule` object, is what I think we're doing.

Jacoby6000: Ah right. Forgot that part.  Will add it

Jacoby6000: Looks like we're doing it a little differently for newts, but you were write about the discoverability problem

edmundnoble:
Now that we have #1910, the instances should go in a different place @Jacoby6000, otherwise I'm :+1: for merge.

Jacoby6000:
@edmundnoble I actually just rethought this. I did this wrong... Instead, `Validation` should just be a tag that goes on `Either`, which prompts the compiler to use a different `Applicative` instance.. Maybe add syntactic sugar for something which isn't called `flatMap`, but has the same signature... `cats` has `andThen` on validation, for example.

PR #1833, opened 2018-05-30
Fix type inference for disjunction fold

### LINKED ISSUE #1831: Disjunction#fold bad type inference
`fold` on scalaz 8 disjunction is subject to the [same problem](https://github.com/scala/bug/issues/10642) as scala stdlib `Option#fold`.

Jacoby6000:
Resolved #1831


jdegoes:
Do you have an example showing better type inference?

PR #1838, opened 2018-05-31
Swap StateT type parameters

JordanMartinez:
Continuation of #1800 so that that PR can be merged while the problems with StateT get figured out.

### DIFF
--- a/core/src/main/scala/scalaz/StateT.scala
+++ b/core/src/main/scala/scalaz/StateT.scala
@@ -3,7 +3,7 @@ package scalaz
 import scala.annotation.tailrec
 import Id._
 
-sealed abstract class IndexedStateT[F[_], -S1, S2, A] { self =>
+sealed abstract class IndexedStateT[-S1, F[_], S2, A] { self =>
   import IndexedStateT._
 
   /** Run and return the final value and state in the context of `F` */
@@ -24,7 +24,7 @@ sealed abstract class IndexedStateT[F[_], -S1, S2, A] { self =>
     abstract class Eval {
       type S0
       val s0: S0
-      val st: IndexedStateT[F, S0, S2, A]
+      val st: IndexedStateT[S0, F, S2, A]
 
       @tailrec
       final def step: F[Eval \/ (S2, A)] = st match {
@@ -35,7 +35,7 @@ sealed abstract class IndexedStateT[F[_], -S1, S2, A] { self =>
     }
 
     object Eval {
-      def apply[S](f: IndexedStateT[F, S, S2, A], s: S): Eval = new Eval {
+      def apply[S](f: IndexedStateT[S, F, S2, A], s: S): Eval = new Eval {
         type S0 = S
         val s0 = s
         val st = f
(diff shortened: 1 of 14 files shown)


core/src/main/scala/scalaz/StateT.scala
```diff
@@ -3,7 +3,7 @@ package scalaz
 import scala.annotation.tailrec
 import Id._
 
-sealed abstract class IndexedStateT[F[_], -S1, S2, A] { self =>
+sealed abstract class IndexedStateT[-S1, F[_], S2, A] { self =>
```
TomasMikula: `F[_]` should be the second to last parameter, i.e. `IndexedStateT[-S1, S2, F[_], A]`.

JordanMartinez: Just pushed that but the build still fails. I'm assuming that the issue here is similar to EitherT in that type inference is not working correctly? `StateT` is ultimately a wrapper around `IndexedStateT`, so would there be a test that could prove/disprove this?

scalacheck-binding/src/main/scala/scalaz/scalacheck/ScalazArbitrary.scala
```diff
@@ -549,8 +549,8 @@ object ScalazArbitrary extends ScalazArbitraryPlatform {
   implicit def indexedContsTArb[W[_], M[_], R, O, A](implicit A: Arbitrary[W[A => M[O]] => M[R]]): Arbitrary[IndexedContsT[W, M, R, O, A]] =
     Functor[Arbitrary].map(A)(IndexedContsT(_))
 
-  implicit def indexedStateTArb[F[_], S1, S2, A](implicit A: Arbitrary[S1 => F[(S2, A)]]): Arbitrary[IndexedStateT[F, S1, S2, A]] =
-    Functor[Arbitrary].map(A)(IndexedStateT[F, S1, S2, A](_))
+  implicit def indexedStateTArb[S1, F[_], S2, A](implicit A: Arbitrary[S1 => F[(S2, A)]]): Arbitrary[IndexedStateT[S1, S2, F, A]] =
```
TomasMikula: For consistency, let's make it `indexedStateTArb[S1, S2, F[_], A]`.

core/src/main/scala/scalaz/StateT.scala
TomasMikula: If the problem is that the tests hang, it has nothing to do with type inference, since the code already compiled.

scalacheck-binding/src/main/scala/scalaz/scalacheck/ScalazArbitrary.scala
```diff
@@ -537,8 +537,8 @@ object ScalazArbitrary extends ScalazArbitraryPlatform {
     Functor[Arbitrary].map(A)(LazyEitherT[F, A, B](_))
 
   // backwards compatibility
-  def stateTArb[F[+_], S, A](implicit A: Arbitrary[S => F[(S, A)]]): Arbitrary[StateT[F, S, A]] =
-    indexedStateTArb[F, S, S, A](A)
+  def stateTArb[F[+_], S, A](implicit A: Arbitrary[S => F[(S, A)]]): Arbitrary[StateT[S, F, A]] =
```
JordanMartinez: Due to your other comment, should this also be changed to `def stateTArb[S, F[+_], A]`?

TomasMikula: yes!

core/src/main/scala/scalaz/package.scala
```diff
@@ -147,28 +147,28 @@ package object scalaz {
...
   /** A state transition, representing a function `S => (S, A)`.
     *
     * @template
     */
-  type State[S, A] = StateT[Id, S, A]
+  type State[S, A] = StateT[S, Id, A]
 
   object StateT extends StateTInstances with StateTFunctions {
-    def apply[F[_], S, A](f: S => F[(S, A)]): StateT[F, S, A] = IndexedStateT[F, S, S, A](f)
+    def apply[S, F[_], A](f: S => F[(S, A)]): StateT[S, F, A] = StateT[S, F, A](f)
```
TomasMikula: This line introduces an infinite loop.

scalacheck-binding/src/main/scala/scalaz/scalacheck/ScalazArbitrary.scala
TomasMikula: Other than this, looks good to me.

JordanMartinez:
Do you guys care about whether this gets squashed down into one commit?

TomasMikula:
Nah, we can squash and merge.

PR #1840, opened 2018-05-31
rename monoid empty -> mempty

hrhino:
#1814 showed a potential clash between `Monoid`'s `empty` and `Maybe`'s `empty`.

I don't think we should rename `append` -> `mappend`, since I've seen that name confuse new users in two different ways:
- `mappend` is a `Semigroup` method (in scalaz), so the reason for the leading `m` is unclear
- it unfortunately also looks like it begins with `map`

also remove -Xlint/-Ywarn options from tut, because they were leaking into the generated markdown


jbgi:
What about `idElement`?

hrhino:
I have no very strong feelings other than that `empty` is not the best.

`id` is the identity for `Category`; `idElement` is similar but potentially confusing.

jdegoes:
It's called identity element typically. I'm OK with `mempty` though the asymmetry with `append` is noticeable.

hrhino:
Yes, the asymmetry is somewhat concerning.

hrhino:
`mappend` is only weird because it's not (just) a monoid operation. I'll change it.

PR #1745, opened 2018-05-03
derivation for isomorphic things (#1662)

valydia:
Following up on #1720

PR #1842, opened 2018-06-01
auto generate "fromIso" by project/GenTypeClass.scala

### LINKED ISSUE #1662: fromIso everywhere

xuwei-k:
fix #1662 
previous pull request https://github.com/scalaz/scalaz/pull/1745

### DIFF
--- a/core/src/main/scala/scalaz/Align.scala
+++ b/core/src/main/scala/scalaz/Align.scala
@@ -65,6 +65,14 @@ trait Align[F[_]] extends Functor[F] { self =>
 object Align {
   @inline def apply[F[_]](implicit F: Align[F]): Align[F] = F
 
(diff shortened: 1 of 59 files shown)


valydia:
This solution is indeed much more elegant 😃

PR #1845, opened 2018-06-02
Make the docs great again

NeQuissimus:
#1820 reverted a bunch of stuff I had done in `example`, causing a few dead links. Fixed this.
Also, I added cross-linking of
- Internal docs
- References to Scala Standard Library classes point to [Scaladoc](https://www.scala-lang.org/api/current/index.html)
- References to JDK classes point to [Javadoc](https://docs.oracle.com/javase/8/docs/api/)
- Every page links to the source code of the file it describes (GitHub logo at the top)

I created empty pages for those types that had been linked but not created.
Lastly, I changes all the file names in `std` to lower case spelling.

GitHub logo comes from https://github.com/logos, we are allowed to use it as per that page.


NeQuissimus:
Deployed to http://nequissimus.com/scalaz/

example/src/main/tut/ct/Kleisli.md
```diff
@@ -3,12 +3,11 @@ layout: docs
 title:  "Kleisli"
 ---
 
-# Kleisli
+# Kleisli [![GitHub](../img/github.png)](https://github.com/scalaz/scalaz/blob/series/8.0.x/base/shared/src/main/scala/scalaz/ct/kleisli.scala)
```
jdegoes: All the hard coded links look hard to maintain. Is there a better way?

NeQuissimus: Unfortunately, no. The branch name should never change? So only larger refactors should have any impact on this.

jdegoes:
@NeQuissimus A few conflicts then we can get this merged in.

NeQuissimus:
@jdegoes Should be good to go

PR #1849, opened 2018-06-02
Add README, expand welcome page

NeQuissimus:
`sbt-buildinfo` provides an object containing the versioning information included in `build.sbt`.
By making this object available to the `examples` project, we can generate a welcome page that includes the current version of all artifacts, without having to ever maintain the versions by hand.

I also added community/chat links as well as a build status.


NeQuissimus:
It's the same info that will be used to publish. It will require the website is only published off tags.

PR #1851, opened 2018-06-03
Cord v2

hrhino:
Continues #1633 at request of @edmundnoble. `Cord` work is all his.

### DIFF
--- a/base/shared/src/main/scala/scalaz/debug/debug.scala
+++ b/base/shared/src/main/scala/scalaz/debug/debug.scala
@@ -3,42 +3,63 @@ package debug
 
 import scala.{ sys, AnyVal, StringContext }
 
-import scalaz.ct.ContravariantClass
-
 import scala.annotation.implicitAmbiguous
 import scala.language.experimental.macros
 import scala.language.implicitConversions
+
+import ct.ContravariantClass
+import data.Cord
 
 /** A typeclass describing types which can be meaningfully represented as a `String`.
  */
 trait DebugClass[A] {
 
-  /** Produce a `String` representation of `a`.
+  /** Produce a [[Cord]] representation of `a`.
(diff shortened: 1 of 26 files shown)


jdegoes:
@hrhino Is this good to merge?

hrhino:
I think so, but that's why I put it up 😉. The `Cord` implementation is verbatim from #1633.

hrhino:
So it didn't actually work. Fixed now.

hrhino:
Now it is, IMHO, g2g.

base/shared/src/main/scala/scalaz/debug/debug.scala
```diff
@@ -3,66 +3,97 @@ package debug
...
   }
 
   implicit final def debugInterpolator(sc: StringContext): DebugInterpolator.Interpolator =
     new DebugInterpolator.Interpolator(sc)
 }
 
-object DebugInterpolator {
-  final class HasDebug private (override val toString: String) extends AnyVal
+object DebugInterpolator extends DebugInterpolator {
+  implicit def mkHasDebug[A](x: A)(implicit D: Debug[A]): HasDebug =
```
alexknvl: This is a cool trick. I would normally do `Instance[Debug]`, which unnecessarily delays `shows`.

hrhino: This is @edmundnoble's suggestion. And yes, it is.

hrhino:
If we want to merge this soon I'll make a follow-up ticket, otherwise just hold tight

hrhino:
@edmundnoble get off your ass and give us ztest

hrhino:
Thanks Ed.

@jdegoes Is this good to merge?

PR #1855, opened 2018-06-05
IdOps.applyIf: apply f° if predicate satisfied.

jbgi:


core/src/main/scala/scalaz/syntax/IdOps.scala
```diff
@@ -16,9 +16,15 @@ final class IdOps[A](private val self: A) extends AnyVal {
   final def ▹[B](f: A => B): B =
     f(self)
 
+  /** Applies `self` to the provided function if the predicate is satisfied, otherwise return self. */
+  @inline def ?|>[B](p: A => Boolean, f: A => A): A = if (p(self)) f(self) else self
```
xuwei-k: unused type parameter?

```diff
- @inline def ?|>[B](p: A => Boolean, f: A => A): A
+ @inline def ?|>(p: A => Boolean, f: A => A): A
```

jbgi: good catch, thanks! fixed it.

PR #1856, opened 2018-06-05
Don't provide the ZipList applicative instance for List in the docs.

hrhino:
It misled someone on Gitter and is a bad bad example for the children.


uglyfigurine:
_Someone_ thinks its best to not introduce kids to `ZipList` when they are learning about `Applicative` :P

edmundnoble:
@uglyfigurine Agreed, but this is an improvement for now.

PR #1857, opened 2018-06-05
Add Kleisli arrow with operation fusion for imported (impure) functions

jdegoes:
Based on the work of John Hughes in [Programming with Arrows](http://www.cse.chalmers.se/~rjmh/Papers/arrows.pdf) and similar in spirit to TraneIO's [Arrows](https://github.com/traneio/arrows), this pull request introduces an arrow abstraction for importing and composing effectful code in a purely-functional way.

In this work, Kleisli arrows are specialized to `IO`. While the behavior and interface of `KleisliIO` is identical to `Kleisli` on the `IO` type, a separate constructor allows impure functions to be imported as `A => B`, without any `IO` overhead. This constructor, in turn, allows fusion of core arrow operations when both left and right side are impure functions. The result is a dramatic reduction in overhead for performance-critical sections of code composed entirely with impure functions.

Although designed and optimized for fusion of impure functions, `KleisliIO` works seamlessly with pure functions lifted into the structure. Sections that can be optimized _are_ optimized, while still allowing incorporation of the greater flexibility and expressiveness that `IO` provides.

Current major limitations include:

 - Lack of support for recursive combinator
 - Lack of support for infinite / lazy composition 
 - Lack of support for asynchronous
 - There are virtually no error combinators

Because of the support for `IO`, these limitations do not actually hinder expressibility. Anything that cannot currently be done in `KleisliIO` directly can be done in `IO` and lifted into `KleisliIO`.

While almost universally faster for every class of problem it can be used to solve, there is definitely overhead associated with arrow-based programming—mainly all the tupling/eithering that occurs as information and conditions are propagated forward. Future work should be able to eliminate this overhead and show greater performance increases on very complex code written in the arrow style.

This pull request does not include a test suite but @wi101, who assisted with implementing and optimizing this code, has developed a test suite and will contribute it separately.


effect/shared/src/main/scala/scalaz/effect/KleisliIO.scala
```diff
@@ -0,0 +1,441 @@
...
+   * Lifts an impure function into `KleisliIO`, converting throwables into the
+   * specified error type `E`.
+   */
+  final def impure[E, A, B](catcher: PartialFunction[Throwable, E])(f: A => B): KleisliIO[E, A, B] =
+    new Impure(
+      (a: A) =>
+        try f(a)
+        catch {
+          case t: Throwable if (catcher.isDefinedAt(t)) =>
+            throw new KleisliIOError(catcher(t))
```
hrhino: this can be

```scala
try f(a) catch catcher.andThen(new KleisliIOError(_))
```

to avoid executing `catcher` twice.

benchmarks/src/main/scala/scalaz/effect/ArrayFillBenchmark.scala
```diff
@@ -0,0 +1,70 @@
...
+    import IOBenchmarks.monixScheduler
+
+    def arrayFill(array: Array[Int])(i: Int): Task[Unit] =
+      if (i >= array.length) Task.unit
+      else Task.eval(array.update(i, i)).flatMap(_ => arrayFill(array)(i + 1))
+
+    (for {
+      array <- Task.eval(createTestArray)
+      _     <- arrayFill(array)(0)
+    } yield ()).toIO.unsafeRunSync()
```
gvolpe: Calling `toIO` on `Task` doesn't seem fair. You could use `runSyncUnsafe` instead.

effect/shared/src/main/scala/scalaz/effect/KleisliIO.scala
jdegoes: This doesn't type check, and I'm pretty sure even if it did, `andThen` would do the same thing behind the scenes that is being done here.

benchmarks/src/main/scala/scalaz/effect/ArrayFillBenchmark.scala
jdegoes: Fixed, although it has zero impact on performance because it's dominated by the individual operations themselves.

effect/shared/src/main/scala/scalaz/effect/KleisliIO.scala
```diff
@@ -0,0 +1,441 @@
...
+ *
+ * // Program 2
+ * val program2: IO[Void, Unit] =
+ *   (readLine.zipWith(readLine)("You wrote: " + _ + ", " + _) >>> printLine)(())
+ * }}}
+ *
+ * In both of these examples, the `KleisliIO` program is faster because it is
+ * able to perform fusion of effectful functions.
+ */
+sealed trait KleisliIO[E, A, B] extends (A => IO[E, B]) { self =>
```
puffnfresh: Extending functions makes me worried

puffnfresh:
What's the benchmarks look like?

effect/shared/src/main/scala/scalaz/effect/KleisliIO.scala
jdegoes: Why?

jdegoes:
@puffnfresh It's an average of 2x as fast as the fastest monadic approaches across the example benchmarks: https://www.slideshare.net/jdegoes/blazing-fast-pure-effects-without-monads-lambdaconf-2018

edmundnoble:
This needs to have review before it goes in, @jdegoes.

effect/shared/src/main/scala/scalaz/effect/KleisliIO.scala
puffnfresh: Seems like containing a value of this type is just as useful. What's the benefit of extension here?

jdegoes: I think the (tiny) benefit is that you can use it wherever `A => IO[E, B]` is required.

Why prefer `(run: A => IO[E, B])`?

jdegoes:
@edmundnoble It had 2 reviews and two thumbs up. Or do you mean Github review?

effect/shared/src/main/scala/scalaz/effect/KleisliIO.scala
aaronvargo: At the very least, subtyping can interact with partial unification in surprising ways. Consider:

```scala
val foo: KleisliIO[Foo, Bar, Baz] = ...

def bar[F[_], G[_], A](x: F[G[A]]) = x

val baz = foo.map(f)
val qux = bar(foo).map(f)
```

Quiz: Are `baz` and `qux` the same?

Answer: It depends on the definition of `Baz`!

Revised Quiz: What are the semantics of `baz` and `qux` for each of the following definitions of `Baz`?:

```scala
trait Baz
type Baz = Qux[Int]
```

Now seemingly innocuous changes like refactoring `trait Baz` to `type Baz = Qux[Int]` can break code, and with sufficient polymorphism can even cause silent semantic changes rather than type errors.

There may be other problems as well, but I'd have to think about it, and I'd rather just write `.run`

jdegoes: @aaronvargo Sounds good. I'll submit a PR. 👍

edmundnoble:
@jdegoes Github approval, specifically.

PR #1858, opened 2018-06-05
Endo/Const/Identity noots.

hrhino:


base/shared/src/main/scala/scalaz/data/identity.scala
```diff
@@ -1,32 +1,48 @@
...
-          case _                            => false
-      }
-    )
+sealed abstract class IdentityModule {
+  type Identity[A]
+
+  def apply[A](a: A): Identity[A]
+  def run[A](id: Identity[A]): A
+
+  def identityMonad: Monad[Identity]
```
edmundnoble: Is this pattern for avoiding the extra method call for `apply` and `run`?

base/shared/src/main/scala/scalaz/prelude.scala
```diff
@@ -149,11 +149,9 @@ trait AllFunctions
 
 trait AllInstances
     extends data.AMaybeInstances
-    with data.ConstInstances
```
edmundnoble: :+1:

base/shared/src/main/scala/scalaz/data/identity.scala
hrhino: Yeah, I guess. It was easier to write the instance with the transparent type alias. I suppose it's a bit unfair.

PR #1860, opened 2018-06-06
Make Contravariant a subclass of InvariantFunctor

### LINKED ISSUE #1859: Contravariant should be a subclass of InvariantFunctor
`Functor` already is. Just do the same thing.

earldouglas:
Fixes #1859

### DIFF
--- a/base/shared/src/main/scala/scalaz/ct/phantom.scala
+++ b/base/shared/src/main/scala/scalaz/ct/phantom.scala
@@ -6,7 +6,10 @@ import scalaz.data.Const
 import scala.language.experimental.macros
 
 trait PhantomClass[F[_]] extends FunctorClass[F] with ContravariantClass[F] {
+
   def pmap[A, B](ma: F[A]): F[B]
+
+  override def imap[A, B](ma: F[A])(f: A => B)(g: B => A): F[B] = pmap(ma)
 }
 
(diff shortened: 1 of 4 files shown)


base/shared/src/main/scala/scalaz/ct/phantom.scala
```diff
@@ -6,7 +6,10 @@ import scalaz.data.Const
 import scala.language.experimental.macros
 
 trait PhantomClass[F[_]] extends FunctorClass[F] with ContravariantClass[F] {
+
   def pmap[A, B](ma: F[A]): F[B]
+
+  override def imap[A, B](ma: F[A])(f: A => B)(g: B => A): F[B] = pmap(ma)
 }
```
hrhino: arguably you can also add overrides for `map` and `contramap` but that doesn't need to be done now.

earldouglas: Would that make `DeriveMapContramap` redundant?

hrhino: ah, yeah, we're doing it there. I'm working on removing the `Derive*` traits, so don't worry.

hrhino:
Thanks, @earldouglas!!

PR #1861, opened 2018-06-06
[wip?] scalaz compiler plugin to guarantee sufficient typeclass definitions

hrhino:


base/shared/src/main/scala/scalaz/ct/applicative.scala
```diff
@@ -5,13 +5,8 @@ import scala.language.experimental.macros
...
 trait ApplicativeClass[F[_]] extends ApplyClass[F] {
   def pure[A](a: A): F[A]
-}
-
-object ApplicativeClass {
 
-  trait DeriveMap[F[_]] extends ApplicativeClass[F] with MonadClass.Alt[DeriveMap[F]] {
-    final override def map[A, B](ma: F[A])(f: (A) => B): F[B] = ap(ma)(pure(f))
-  }
+  override def map[A, B](ma: F[A])(f: A => B): F[B] = ap(ma)(pure(f))
```
alexknvl: Epic.

base/shared/src/main/scala/scalaz/ct/phantom.scala
```diff
@@ -3,22 +3,12 @@ package ct
...
-
-  trait DeriveMapContramap[F[_]] extends PhantomClass[F] with Alt[DeriveMapContramap[F]] {
-    final override def map[A, B](ma: F[A])(f: (A) => B): F[B]       = pmap(ma)
-    final override def contramap[A, B](ma: F[A])(f: (B) => A): F[B] = pmap(ma)
-  }
-
-  trait DerivePmap[F[_]] extends PhantomClass[F] with Alt[DerivePmap[F]] {
-    final override def pmap[A, B](ma: F[A]): F[B] = contramap(map(ma)(_ => ()))(_ => ())
-  }
+  def pmap[A, B](ma: F[A]): F[B] = contramap(map(ma)(_ => ()))(_ => ())
```
alexknvl: :+1:

effect/shared/src/main/scala/scalaz/effect/FiberInstances.scala
```diff
@@ -17,18 +17,18 @@ trait FiberInstances {
 
   implicit def fiberApplicative[E]: Applicative[Fiber[E, ?]] =
     instanceOf(new ApplicativeClass[Fiber[E, ?]] {
-      final def pure[A](a: A): Fiber[E, A] = new Fiber[E, A] {
+      override def pure[A](a: A): Fiber[E, A] = new Fiber[E, A] {
```
alexknvl: `final override def`?

base/shared/src/main/scala/scalaz/types/iscovariant.scala
```diff
@@ -9,73 +11,32 @@ package types
  * https://typelevel.org/blog/2014/03/09/liskov_lifting.html
  */
 trait IsCovariantClass[F[_]] {
```
alexknvl: no `@minimal`?

base/shared/src/main/scala/scalaz/types/iscontravariant.scala
```diff
@@ -9,63 +11,32 @@ package types
  * https://typelevel.org/blog/2014/03/09/liskov_lifting.html
  */
 trait IsContravariantClass[F[_]] {
```
alexknvl: `@minimal`

base/shared/src/main/scala/scalaz/ct/bind.scala
```diff
@@ -3,27 +3,12 @@ package ct
...
-  trait DeriveFlatten[M[_]] extends BindClass[M] with Alt[DeriveFlatten[M]] {
-    final override def flatten[A](ma: M[M[A]]): M[A] = flatMap(ma)(identity)
-  }
-
-  trait DeriveFlatMap[M[_]] extends BindClass[M] with Alt[DeriveFlatMap[M]] {
-    final override def flatMap[A, B](ma: M[A])(f: (A) => M[B]): M[B] = flatten(map(ma)(f))
-  }
-
-  trait Alt[D <: Alt[D]]
+  def flatMap[A, B](ma: M[A])(f: A => M[B]): M[B] = flatten(map(ma)(f))
```
jbgi: Could this cause an issue with default impl of `map ` in `Monad`? ie if one override only `flatten` then we can have a circular definition between `map` and `flatMap`:
```scala
  def flatMap[A, B](ma: M[A])(f: A => M[B]): M[B] = flatten(map(ma)(f))
  def map[A, B](ma: M[A])(f: A => B): M[B] = flatMap(ma)(a => pure(f(a)))
```
this case was checked by the "`Alt`"-encoding, would be nice to preserve this check.
I think there is some other similar cases as well.

PR #1862, opened 2018-06-07
Fix type params in Kleisli.md

ktonga:


ktonga:
Hey @hrhino I do prefer `F` better too. Also it is used in the implementation so I tried to change all the `M`s with `F`s. Please double check.

Cheers.

ktonga:
Yeah, you right. Lemme fix it.

ktonga:
Did I miss any? otherwise ready to go.

emilypi:
I'm on board with this. Thanks @ktonga

PR #1869, opened 2018-06-11
enable tests for Scala 2.13.0-M4

xuwei-k:


scalacheck-binding/src/main/scala/scalaz/scalacheck/ScalazArbitrary.scala
```diff
@@ -459,8 +458,6 @@ object ScalazArbitrary extends ScalazArbitraryPlatform {
 
   implicit def MaxMaybeArbitrary[A: Arbitrary]: Arbitrary[MaxMaybe[A]] = Tag.subst(arb[Maybe[A]])
 
-  implicit def ArraySeqArbitrary[A: Arbitrary]: Arbitrary[ArraySeq[A]] = Functor[Arbitrary].map(arb[List[A]])(x => ArraySeq(x: _*))
```
xuwei-k: https://github.com/scalaz/scalaz/commit/104419ce0c7b6343f605d9b8fbe4ac5a8ebbddc3

PR #1872, opened 2018-06-12
Don't define instances alongside class definitions

aaronvargo:
Instances should be organized by data type, otherwise they're hard to find since they may come from any subclass.


base/shared/src/main/scala/scalaz/std/function.scala
```diff
@@ -0,0 +1,79 @@
+package scalaz
+package std
+
+import scala.{ Function0, Function1 }
+import scalaz.ct._
+import scalaz.data.Disjunction
+
+trait Function0Instances {
+
+  implicit val functionMonad: Monad[Function0] = instanceOf(
```
hrhino: `function0Monad`

base/shared/src/main/scala/scalaz/std/function.scala
```diff
@@ -0,0 +1,79 @@
...
+  implicit def function1Monad[C]: Monad[Function1[C, ?]] =
+    instanceOf(new MonadClass[Function1[C, ?]] with BindClass.DeriveFlatten[Function1[C, ?]] {
+      override def ap[A, B](fab: Function1[C, A])(f: Function1[C, A => B]): Function1[C, B] = (c: C) => f(c)(fab(c))
+      override def map[A, B](fab: Function1[C, A])(f: A => B): Function1[C, B]              = fab andThen f
+      override def flatMap[A, B](fab: Function1[C, A])(f: A => Function1[C, B]): Function1[C, B] =
+        (c: C) => f(fab(c))(c)
+      override def pure[A](a: A): Function1[C, A] = (c: C) => a
+    })
+
+  implicit val functionChoice: Choice[? => ?] = instanceOf(
```
hrhino: `function1Choice`

base/shared/src/main/scala/scalaz/std/function.scala
```diff
@@ -0,0 +1,79 @@
...
+      override def leftchoice[A, B, C](ab: A => B): A \/ C => B \/ C =
+        _.fold(a => Disjunction.left(ab(a)), Disjunction.right)
+
+      override def rightchoice[A, B, C](ab: A => B): C \/ A => C \/ B =
+        _.fold(Disjunction.left, a => Disjunction.right(ab(a)))
+    }
+  )
+
+  implicit val functionStrong: Strong[? => ?] = instanceOf(
+    new StrongClass[? => ?] with ProfunctorClass.DeriveDimap[? => ?] {
```
hrhino: Factor out a `private trait FunctionProfunctor extends ProfunctorClass[? => ?] with ProfunctorClass.DeriveDimap[? => ?]`?

base/shared/src/main/scala/scalaz/std/option.scala
```diff
@@ -0,0 +1,15 @@
+package scalaz
+package std
+
+import scala.{ Option, Some }
+
+import scalaz.ct._
+
+trait OptionInstances {
```
hrhino: we could, separately, use with some more of these.

purefn:
I don't think we want to do this because now these are orphan instances. It makes them harder to use and creates other potential problems.

aaronvargo:
They're no more "orphans" than they were before. With the current type class encoding, typeclasses don't have companion objects for instances to go in anyway. Either way instances are being mixed in via traits.

hrhino:
@purefn for the anyval/string/function ones, which we're putting in `base`, mixing them into `scalaz` (which I think we [ought to] do?) makes them non-orphan. 

For the stdlib instances, it appears that we decided to package those as a separate module (`scalaz-std`) so that's already unavoidable. 

Ping @NeQuissimus

base/shared/src/main/scala/scalaz/std/option.scala
aaronvargo: Apparently, yes. Some of the instances were duplicated. I just copy/pasted all the instances I deleted (or most of them. I think I missed a couple. Oops.). I guess I'll move everything to the std module then...

aaronvargo:
That decision seems questionable. What do we gain from it?

hrhino:
> That decision seems questionable. What do we gain from it?

I'll let someone who was more involved in that decision answer that.

> I guess I'll move everything to the std module then...

please leave `FunctionN`, `String`, and `AnyVal` subtypes in `base`'s `scalaz.std`. Those we don't (and can't) have good replacements for, so I think we're packaging them with base.

base/shared/src/main/scala/scalaz/std/function.scala
aaronvargo: Function1Profunctor

hrhino: of course, that :P (phone, apologies)

NeQuissimus:
Everything that we cannot/will not replace (Primitives, Function0, AnyVal, whatnot) but originates in the stdlib is in the `std` package in the `base` module. They are required to write most meaningful applications after all.

Everything we have replacements for, but we figure typeclass instances are useful for, we provide instances in the separate `std` module. (which effectively adds to the `std` package)

Absolutely everything in the `base` module ought to be linked into the `Prelude` one way or another (there are multiple options towards the bottom of the file; basically the further you scroll down, the more is included).
Everything in the std module is included in the package `scalaz.std`

We should not have instances floating around some package.

edmundnoble:
@aaronvargo 

"With the current type class encoding, typeclasses don't have companion objects for instances to go in anyway."

This isn't true, though it isn't obvious. Because `InstanceOf[A] <: A`, `EqClass` for example provides a suitable companion for `Eq`.

alexknvl:
Closing since it is superseded by #1910 AFAIU.

PR #1874, opened 2018-06-13
Add dependency on ZIO and remove effect module

ktonga:


base/shared/src/main/scala/scalaz/core/io.scala
```diff
@@ -0,0 +1,71 @@
...
+import java.lang.Throwable
+
+import com.github.ghik.silencer.silent
+
+import algebra.MonoidClass
+import ct._
+import types.IsCovariantClass
+import zio.{ IO, Fiber }
+
+trait IOModule
```
hrhino: needed?

base/shared/src/main/scala/scalaz/core/io.scala
```diff
@@ -0,0 +1,71 @@
...
+
+      final def map[A, B](ma: Fiber[E, A])(f: A => B): Fiber[E, B] = new Fiber[E, B] {
+        def join: IO[E, B]                            = ma.join.map(f)
+        def interrupt[E2](t: Throwable): IO[E2, Unit] = ma.interrupt(t)
+      }
+    })
+}
+
+@silent
+private[core] object IOImpl extends IOModule with IOInstances with FiberInstances
```
hrhino: likewise, what does this do?

base/shared/src/main/scala/scalaz/core/io.scala
ktonga: Don't think so, just following conventions. Can remove it.

ktonga:
Ping @jdegoes

jdegoes:
@ktonga Nicely done! Any ideas on how to keep this up to date with the latest ZIO?

ktonga:
Actually I'm more concerned about the opposite, it will be kept *too* up to date, since we depend on a SNAPSHOT and it will be re-checked on every scalaz build.

To make it a bit more deterministic i'd suggest:

1. Ideally, we introduce source dependency (as @fommil has been suggesting ad nauseum :smile: ) and we move the head of the submodule as required. Dunno how mature support  for that in scala is right now so maybe something to check again in the future.
2. A more realistic solution would be to start using semantic versioning in ZIO and use a version expression in scalaz instead of a fixed one.

Issue #1716, opened 2018-04-14
Make Traversables short-circuiting _and_ stack-safe


TomasMikula:
Traversals can be made both stack-safe _and_ short-circuiting, provided that the target `Applicative` implements short-circuiting. The key is to use the applicative's `unfoldrOpt` method for the implementation of `traverse`.

This has been done for `List` in https://github.com/scalaz/scalaz/commit/79c5413274b64e144049d4279c3a69ea17087e5c.

We should implement this at least for sequential data structures with natural left-to-right traversal:

 - [ ] `IList`
 - [ ] `NonEmptyList`
 - [ ] `Stream`
 - [ ] `EphemeralStream`

malancas:
Hello, I'll pick this issue up.

TomasMikula:
Awesome! It's also fine to do them one at a time.

malancas:
Hey, I plan to have a PR up for this in the next day or so.

Issue #1515, opened 2017-11-13
eager evaluation of EphemeralStream.traverse


fommil:
the following shows EphemeralStream eagerly evaluating all elements when used is `stream.traverse(f)` where the other type constructor is `Id`. Notably, this does *not* happen for stdlib `Stream`

```scala
import scala.language.higherKinds
import scalaz._, Scalaz._
import EphemeralStream.{ fromStream, EStream }

object Laziness {
  def testing[G[_]: Applicative](s: EStream[Int])(f: Int => G[Int]) =
    s.traverse(f)

  def main(args: scala.Array[String]): scala.Unit = {
    val stream = EphemeralStream.iterate(0) { i =>
      println(i)
      i + 1
    }

    val trav = testing[Id](stream)(identity)
    println(trav)
  }
}

```

fommil:
possible reason

```
  def foldRight[B](z: => B)(f: (=> A) => (=> B) => B): B =
    if (isEmpty) z else f(head())(tail().foldRight(z)(f))
```

both `head()` and `tail()` will evaluate their contents even though the `f` can take lazy parameters.

Also, note that this is actually `foldr` (curried), not the usual `Foldable.foldRight`

Same thing in the Traverse impl

```
    override def foldRight[A, B](fa: EphemeralStream[A], z: => B)(f: (A, => B) => B): B =
      if(fa.isEmpty) z else f(fa.head(), foldRight(fa.tail(), z)(f))

```

fommil:
This is the repeating stacktrace

```
background log: error: 	at scalaz.IdInstances$$anon$1.ap(Id.scala:57)
background log: error: 	at scalaz.Apply.ap2(Apply.scala:49)
background log: error: 	at scalaz.Apply.ap2$(Apply.scala:48)
background log: error: 	at scalaz.IdInstances$$anon$1.ap2(Id.scala:19)
background log: error: 	at scalaz.Applicative.apply2(Applicative.scala:32)
background log: error: 	at scalaz.Applicative.apply2$(Applicative.scala:31)
background log: error: 	at scalaz.IdInstances$$anon$1.apply2(Id.scala:19)
background log: error: 	at scalaz.EphemeralStreamInstances$$anon$1.$anonfun$traverseImpl$4(EphemeralStream.scala:208)
background log: error: 	at scalaz.EphemeralStream.foldRight(EphemeralStream.scala:37)
[...]

PR #1875, opened 2018-06-14
Make Traversables short-circuiting and stack-safe

malancas:
For [issue #1716](https://github.com/scalaz/scalaz/issues/1716#issuecomment-388029722)

### DIFF
--- a/core/src/main/scala/scalaz/EphemeralStream.scala
+++ b/core/src/main/scala/scalaz/EphemeralStream.scala
@@ -203,16 +203,16 @@ sealed abstract class EphemeralStreamInstances {
     override def zipWithR[A, B, C](fa: EphemeralStream[A], fb: EphemeralStream[B])(f: (Option[A], B) => C) =
       zipWithL(fb, fa)((b, a) => f(a, b))
     
-    def traverseImpl[F[_], A, B](fa: EphemeralStream[A])(f: A => F[B])(implicit F: Applicative[F]): F[EphemeralStream[B]] = {
-      val revOpt: Maybe[F[EphemeralStream[B]]] =
-        F.unfoldrOpt[EphemeralStream[A], B, EphemeralStream[B]](fa)(_ match {
+    def traverseImpl[F[_], A, B](fa: EphemeralStream[A])(f: A => F[B])(implicit F: Applicative[F]) = {
+      val revOpt: Maybe[F[List[B]]] =
+        F.unfoldrOpt[EphemeralStream[A], B, List[B]](fa)(_ match {
           case a ##:: as => Maybe.just((f(a), as))
           case emptyEphemeralStream => Maybe.empty
-        })(Reducer.ReverseEphemeralStreamReducer[B])
+        })(Reducer.ReverseListReducer[B])
 
-      val rev: F[EphemeralStream[B]] = revOpt getOrElse F.point(emptyEphemeralStream)
+      val rev: F[List[B]] = revOpt getOrElse F.point(Nil)
 
-      F.map(rev)(_.reverse)
+      F.map(rev)((li) => li.foldLeft(EphemeralStream[B]())((r,c) => c ##:: r))
     }
 
     override def index[A](fa: EphemeralStream[A], i: Int): Option[A] = {
--- a/core/src/main/scala/scalaz/EphemeralStream.scala
+++ b/core/src/main/scala/scalaz/EphemeralStream.scala
@@ -28,13 +28,6 @@ sealed abstract class EphemeralStream[A] {
     else Some(tail())
   }
 
-  /*
-  def :+(a: A): EphemeralStream[A] = {
-    val as = EphemeralStream(a)
-    foldRight(as)(_ ##:: _)
-  }
-  */
-
   def toList: List[A] =
     foldLeft(Nil: List[A])((xs, x) => x :: xs).reverse
 
@@ -210,6 +203,7 @@ sealed abstract class EphemeralStreamInstances {
     override def zipWithR[A, B, C](fa: EphemeralStream[A], fb: EphemeralStream[B])(f: (Option[A], B) => C) =
       zipWithL(fb, fa)((b, a) => f(a, b))
     
+    /*
     def traverseImpl[G[_], A, B](fa: EphemeralStream[A])(f: A => G[B])(implicit G: Applicative[G]): G[EphemeralStream[B]] = {
       val seed: G[EphemeralStream[B]] = G.point(EphemeralStream[B]())
 
@@ -217,8 +211,8 @@ sealed abstract class EphemeralStreamInstances {
         (x , ys) => G.apply2(f(x), ys)((b, bs) => EphemeralStream.cons(b, bs))
       }
     }
+    */
 
-    /*
     def traverseImpl[F[_], A, B](fa: EphemeralStream[A])(f: A => F[B])(implicit F: Applicative[F]): F[EphemeralStream[B]] = {
       val revOpt: Maybe[F[EphemeralStream[B]]] =
         F.unfoldrOpt[EphemeralStream[A], B, EphemeralStream[B]](fa)(_ match {
@@ -230,7 +224,6 @@ sealed abstract class EphemeralStreamInstances {
 
       F.map(rev)(_.reverse)
     }
-    */
 
     override def index[A](fa: EphemeralStream[A], i: Int): Option[A] = {
       if(i < 0)
--- a/core/src/main/scala/scalaz/EphemeralStream.scala
+++ b/core/src/main/scala/scalaz/EphemeralStream.scala
@@ -28,6 +28,13 @@ sealed abstract class EphemeralStream[A] {
     else Some(tail())
   }
 
+  /*
+  def :+(a: A): EphemeralStream[A] = {
+    val as = EphemeralStream(a)
+    foldRight(as)(_ ##:: _)
+  }
+  */
+
   def toList: List[A] =
     foldLeft(Nil: List[A])((xs, x) => x :: xs).reverse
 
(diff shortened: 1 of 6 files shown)


core/src/main/scala/scalaz/EphemeralStream.scala
```diff
@@ -202,13 +202,19 @@ sealed abstract class EphemeralStreamInstances {
     }
     override def zipWithR[A, B, C](fa: EphemeralStream[A], fb: EphemeralStream[B])(f: (Option[A], B) => C) =
       zipWithL(fb, fa)((b, a) => f(a, b))
-    def traverseImpl[G[_], A, B](fa: EphemeralStream[A])(f: A => G[B])(implicit G: Applicative[G]): G[EphemeralStream[B]] = {
-      val seed: G[EphemeralStream[B]] = G.point(EphemeralStream[B]())
+    
+    def traverseImpl[F[_], A, B](fa: EphemeralStream[A])(f: A => F[B])(implicit F: Applicative[F]): F[EphemeralStream[B]] = {
+      val revOpt: Maybe[F[EphemeralStream[B]]] =
+        F.unfoldrOpt[EphemeralStream[A], B, EphemeralStream[B]](fa)(_ match {
```
jbgi: can get ride of `_ match` I guess.

core/src/main/scala/scalaz/Reducer.scala
```diff
@@ -157,10 +157,18 @@ sealed abstract class ReducerInstances {
     unitConsReducer(_ :: INil(), _ :: _)
   }
 
+  def ReverseIListReducer[C]: Reducer[C, IList[C]] = {
```
jbgi: instead of the additional Reverse reducers, it would be better to use the `@@ Tags.Dual` reducer. Though its implementation need to be updated to properly override `cons` and `snoc`.

core/src/main/scala/scalaz/EphemeralStream.scala
```diff
@@ -202,13 +202,19 @@ sealed abstract class EphemeralStreamInstances {
...
+          case a ##:: as => Maybe.just((f(a), as))
+          case emptyEphemeralStream => Maybe.empty
+        })(Reducer.ReverseEphemeralStreamReducer[B])
 
-      fa.foldRight(seed) {
-        (x , ys) => G.apply2(f(x), ys)((b, bs) => EphemeralStream.cons(b, bs))
-      }
+      val rev: F[EphemeralStream[B]] = revOpt getOrElse F.point(emptyEphemeralStream)
+
+      F.map(rev)(_.reverse)
```
jbgi: It looks like this implementation will not work for Infinite streams. I think the previous implementation was, provided the `Applicative` was lazy enough.
I guess we probably need both a `lazyTraverse` and a `strictTraverse`...

TomasMikula: There's a test showing it works for an infinite stream (with a short-circuiting implementation of the `Applicative`'s `unfoldrOpt`).

core/src/main/scala/scalaz/Reducer.scala
TomasMikula: We don't even need all these new reverse reducers for the purposes of this PR. We could just use the existing `ReverseListReducer` and then reverse the reversed List into the appropriate data type (`IList`, `Stream`, `EphemeralStream`) using a simple `foldLeft`.

@jbgi's comment still stands for the existing `ReverseListReducer`, but that's for another PR.

core/src/main/scala/scalaz/EphemeralStream.scala
jbgi: `reverse` works for infinite stream??

core/src/main/scala/scalaz/NonEmptyList.scala
```diff
@@ -238,6 +238,11 @@ sealed abstract class NonEmptyListInstances extends NonEmptyListInstances0 {
         (BindRec[IList].tailrecM[A, B](a)(a => f(a).list): @unchecked) match {
           case ICons(h, t) => NonEmptyList.nel(h, t)
         }
+
+      override def traverseImpl[F[_], A, B](fa: NonEmptyList[A])(f: A => F[B])(implicit F: Applicative[F]): F[NonEmptyList[B]] = {
+        val fb = Traverse[IList].traverseImpl(fa.list)(f)(F)
+        F.map(fb)(fb => NonEmptyList.nel(fb.headOption.get, fb.tailOption.getOrElse(INil())))
```
TomasMikula: Let's try to avoid uses of the unsafe `Option#get`.

core/src/main/scala/scalaz/EphemeralStream.scala
TomasMikula: If the `reverse` is reached, the stream was either not infinite, or traversal terminated early and there's no stream inside that `F[EphemeralStream[B]]`.

TomasMikula: If `List` was used for that unfolding, as I suggest in another comment, it would be obvious that the reversal is safe.

jbgi: yeah.
I guess what I am concerned about is that the identity law (`traverse Identity = Identity`) is not preserved by this implementation for infinite streams. In haskell I can do
```haskell
λ> fmap (take 10) $ traverse Identity [1..]
Identity [1,2,3,4,5,6,7,8,9,10]
```

jbgi: right, looks like `apply2`/`ap2`/`ap` is too strict for this to work.

TomasMikula: Actually, it could have worked before and I'm having a [déjà vu](https://github.com/scalaz/scalaz/issues/1515#issuecomment-350547944) of thinking about this. 😃

This `Apply#unfoldrOpt` is useful for cases when appending any more `F[A]`s to an `F[B]` won't change its value.

Turns out it doesn't help when those further `F[A]`s cannot be ignored completely, but their effect part can. Examples:
 - when `F` = `Id`, the effect part of all `F[A]`s can be ignored;
 - when `F` = `Writer[Boolean @@ Conjunction, ?]`, once `false` is encountered, the effect part of all subsequent `F[A]`s can be ignored.

(I can't think of any other cases than `Writer` and `Id` (≈ `Writer[Unit, ?]`).)

I think these cases could be supported at the expense of complicating `unfoldrOpt` signature¹. I would keep it out of this PR, though.

---

¹ Such as the following, changing the return value to `Maybe[F[B]] \/ (Maybe[F[B]], S, F[A] => A)`:

```scala
def unfoldrOpt[S, A, B](seed: S)(f: S => Maybe[(F[A], S)])(implicit R: Reducer[A, B]): Maybe[F[B]] \/ (Maybe[F[B]], S, F[A] => A)
```

where

 - return value of `-\/` means that the seed has been exhausted;
 - return value of `\/-` means that there's some leftover seed, but the remaining effects can be ignored by the returned function `F[A] => A`.

Implementation for `Id` would immediately return `\/-((Maybe.empty, seed, identity[A]))`.

core/src/main/scala/scalaz/NonEmptyList.scala
```diff
@@ -238,6 +238,11 @@ sealed abstract class NonEmptyListInstances extends NonEmptyListInstances0 {
         (BindRec[IList].tailrecM[A, B](a)(a => f(a).list): @unchecked) match {
           case ICons(h, t) => NonEmptyList.nel(h, t)
         }
+
+      override def traverseImpl[F[_], A, B](fa: NonEmptyList[A])(f: A => F[B])(implicit F: Applicative[F]): F[NonEmptyList[B]] = {
+        val fb = Traverse[IList].traverseImpl(fa.list)(f)(F)
+        F.map(fb)(fb => NonEmptyList.nel(fb.headOption.getOrElse(sys.error("Head can't be empty")), fb.tailOption.getOrElse(INil())))
```
malancas: Using `sys.error` here was what first came to mind but let me know if there's another means of handling or message that is preferred.

TomasMikula: I would write a dedicated implementation for `NonEmptyList` (i.e. not delegate to `IList`).

Also, let's do that in the implementation of `traverse1` and then it suffices to keep the default `traverseImpl`.

core/src/main/scala/scalaz/IList.scala
```diff
@@ -604,8 +604,17 @@ sealed abstract class IListInstances extends IListInstance0 {
...
+        val revOpt: Maybe[F[List[B]]] =
+          F.unfoldrOpt[IList[A], B, List[B]](fa)(_ match {
+            case ICons(a, as) => Maybe.just((f(a), as))
+            case INil() => Maybe.empty
+          })(Reducer.ReverseListReducer[B])
+
+        val rev: F[List[B]] = revOpt getOrElse F.point(Nil)
+
+        F.map(rev)((rev) => rev.foldLeft(IList[B]())((r, c) => c +: r))
+      }
```
edmundnoble: Perhaps I'm missing something here, but this doesn't look like it short-circuits.

core/src/main/scala/scalaz/NonEmptyList.scala
```diff
@@ -177,6 +177,17 @@ sealed abstract class NonEmptyListInstances extends NonEmptyListInstances0 {
...
         Foldable[IList].foldMap(fa.list)(f)(M)
 
+      override def traverse1[F[_], A, B](fa: NonEmptyList[A])(f: A => F[B])(implicit F: Apply[F]) = {
+        val revOpt: Maybe[F[NonEmptyList[B]]] =
+          F.unfoldrOpt[IList[A], B, NonEmptyList[B]](fa.list)(_ match {
+            case ICons(a, as) => Maybe.just((f(a), as))
+            case INil() => Maybe.empty
+          })(Reducer.ReverseNonEmptyListReducer[B])
+
+        val rev: F[NonEmptyList[B]] = revOpt getOrElse sys.error("Head cannot be empty")
```
TomasMikula: This again uses an unsafe operation (`sys.error`).

What about
```scala
F.apply2(f(fa.head), fa.tail.traverse)(NonEmptyList.nel)
```
? (Which actually does reuse `IList.traverse`.) Should remain short-circuiting if `apply2` is, and only adds constant stack, so remains stack safe.

malancas: That makes sense to me. Since`fa.tail.traverse` expects `F` to be an `Applicative` instead of an `Apply`, is there a way we can convert the `Apply` passed into `traverse1`?

TomasMikula:
Thanks, @malancas, and sorry for the delay.

PR #1879, opened 2018-06-16
1726 - void datatype for 7.3.x

heyrutvik:
library user need to `import scalaz.Scalaz._` in order to use Void syntax.


core/src/main/scala/scalaz/Void.scala
```diff
@@ -0,0 +1,29 @@
+package scalaz
+
+import Liskov.<~<
+
+trait VoidModule {
+  type Void
+
+  def absurd[A](v: Void): A
+
+  def unsafeVoid: Void
```
hrhino: What do we have `unsafeVoid` for, again?

heyrutvik: honestly, I don't know.. I copied it from @alexknvl's implementation. So, he might have better idea. :)

alexknvl: For example, to convince the compiler that `(Int === String) => Void`.

heyrutvik:
@jdegoes Rather than creating independent `VoidModule` trait. You are asking for extending it from  `scalaz.zio.Void` in 7.3.x. So that way, 7.3.x will depend on `ZIO` and instance will be `ZIO`'s `Void` type. Correct me if I'm wrong.

jdegoes:
Yes, this is exactly what I suggest. The same instances and helpers can be provided, only they will use a safe coercion that can be hidden behind the API.

heyrutvik:
@alexknvl @jdegoes @ktonga need help..
I was thinking to make following changes in order to make scalaz 7.3 to depend on ZIO. 
https://github.com/heyrutvik/scalaz/blob/61032f1d8c0360e13f15508625b1445b18d55608/core/src/main/scala/scalaz/package.scala#L380
```
import scalaz.zio.{Void => ZVoid}
package object scalaz {
  val Void: VoidModule = VoidImpl
  type Void = ZVoid
}
```
https://github.com/heyrutvik/scalaz/blob/61032f1d8c0360e13f15508625b1445b18d55608/core/src/main/scala/scalaz/Void.scala#L5
```
trait VoidModule extends Void {
...
}
```

Is this what you have suggested?
If yes then there is a problem because `zio.Void` is `final` in its declaration. https://github.com/heyrutvik/scalaz-zio/blob/master/core/shared/src/main/scala/scalaz/zio/Void.scala#L9

thoughts?

jdegoes:
@heyrutvik 

Mostly correct, but you would not extend `Void`. Instead you can have:

```scala
trait VoidModule {
   type Void
   def absurd[A](v: Void): A
}
trait VoidImpl extends VoidModule {
  type Void = ZVoid
  def absurd[A](v: Void): A = v.absurd[A]
}
...
val Void: VoidModule = new VoidImpl{}
type Void = Void.Void
```

heyrutvik:
@jdegoes Travis is happy only with scala 2.12 and _you know why_ :)

core/src/main/scala/scalaz/Void.scala
```diff
@@ -1,29 +1,15 @@
 package scalaz
 
-import Liskov.<~<
+import scalaz.zio.{Void => ZVoid}
```
ktonga: `ZioVoid` maybe? we are in scalaz, just the Z ain't enough :slightly_smiling_face:

jdegoes:
@heyrutvik Hopefully ZIO soon supports 2.11, right?

core/src/main/scala/scalaz/Void.scala
```diff
@@ -0,0 +1,15 @@
+package scalaz
+
+import scalaz.zio.{Void => ZioVoid}
+
+trait VoidModule {
+  type Void
+
```
jdegoes: I think maybe want we want to do here is expose the type.

So maybe: `type Void = ZioVoid`. Then delete `VoidImpl`.

The reason is that then Scalaz will have the "same" Void as ZIO, which will result in better interop.

heyrutvik: Okay then implementation will be provided by `VoidModule` itself. right?

so in package object we will have something like..
```
val Void: VoidModule = new VoidModule {}
type Void = Void.Void
```

jdegoes: Yes, I'd make it an object.

core/src/main/scala/scalaz/syntax/VoidSyntax.scala
```diff
@@ -0,0 +1,9 @@
+package scalaz.syntax
+
+import scalaz._
+
+trait VoidOps {
```
jdegoes: This should no longer be needed since it's now a method on `Void`.

xuwei-k:
> I would prefer that (1) Scalaz 7.3 depend on ZIO

As of now, Scalaz 7.3 support scala 2.11, 2.12, 2.13 milestone, scala-js and scala-native. That doesn't appeal to me much.

jdegoes:
> As of now, Scalaz 7.3 support scala 2.11, 2.12, 2.13 milestone, scala-js and scala-native. That doesn't appeal to me much.

What if ZIO supports all these too?

The question is whether to have zio-interop depend on 7.3, or have 7.3 pull in zio.

ktonga:
fyi scalaz/scalaz-zio#89

jdegoes:
@heyrutvik We have discussed at length and it seems that `Nothing` is going to have to work. We will have to disable a compiler warning in order to take advantage of this.

PR #1880, opened 2018-06-16
Void syntax resolution test

heyrutvik:
added syntax resolution test code for Void type


heyrutvik:
Apparently, either I can't add reviewers or I don't know how to do that :confused: 
@alexknvl Can you help me with this?

alexknvl:
```
not found: type Nothing
[error]     v.absurd[Nothing]
```
you need to explicitly import `Nothing` in S8

Issue #1826, opened 2018-05-27
Enforce acyclic package dependencies


jdegoes:
We can use [Acyclic](https://github.com/lihaoyi/acyclic) to enforce no cyclic dependencies in our packages, moving, for example, all data instances into the `ct` package.

 - [ ] Incorporate [Acyclic](https://github.com/lihaoyi/acyclic) into the build
 - [ ] Move code around to achieve no cyclic dependencies
 - [ ] Publish dependency graph in contributors gu ide

barambani:
Sorry it took a while, I have been away. I have given a look at this and I have some observations. I added some comments to the PR. Thanks a lot for the help.

edmundnoble:
Why would we do this? All this will do is disallow good code from being written that uses things from multiple packages.

jmcardon:
B-but why

barambani:
I closed the PR as it looks like this ticket is not ready and might not go anywhere.

PR #1889, opened 2018-06-19
WIP - Enforce acyclic package dependencies

barambani:
This is a WIP for issue https://github.com/scalaz/scalaz/issues/1826. At the moment I just moved the machinery for the `InstanceOf` of the typeclasses to a package that's supposed to be very high if not the root of the hierarchy (almost everything depends on it). I called it `kernel`. I also moved `io` to another package (`external`, but the names are ugly and can be changed) as it should be pretty low level and only the prelude should depend on it. 
While doing that though I noticed that the prelude is tricky. The aliases in `BaseTypeclasses` in fact make it depend on most of the other packages but, if I'm not mistaken, a lot of those packages also depend on it as they use the aliases themselves.
If I run the plugin placing for instance `import acyclic.pkg` in the `external` package object I get something like 

```
[error] Unwanted cyclic dependency
[info] /Users/fmariotti/open-src/scalaz/base/shared/src/main/scala/scalaz/package.scala:8:10:
[info]     with BaseCore {
[info]          ^
[info] symbol: trait BaseCore
[info] More dependencies at lines 3 2 5 2 6 2 2 2 2 2 2 7 4
[info] /Users/fmariotti/open-src/scalaz/base/shared/src/main/scala/scalaz/prelude.scala:179:19:
[info]     with external.IOInstances
[info]                   ^
[info] symbol: trait IOInstances
[info] More dependencies at lines 180 150 150
[info] package scalaz.external
[warn] Choosing Sonatype OSS Snapshots for org.scalaz#scalaz-zio_2.12;0.1-SNAPSHOT
[info] /Users/fmariotti/open-src/scalaz/base/shared/src/main/scala/scalaz/external/io.scala:33:34:
[info]   implicit def ioIsCovariant[E]: IsCovariant[IO[E, ?]] = IsCovariantClass.unsafeForce[IO[E, ?]]
[info]                                  ^
[info] symbol: type InstanceOf
[info] More dependencies at lines 13 38 47 57 43 51 27 62 38
```

any thoughts ? I will keep experimenting, but before touching many files (and changing the way the aliases are used) I wanted to ask for opinions and suggestions here. Thanks a lot.

### DIFF
--- a/base/shared/src/main/scala/scalaz/algebra/monoid.scala
+++ b/base/shared/src/main/scala/scalaz/algebra/monoid.scala
@@ -1,6 +1,8 @@
 package scalaz
 package algebra
 
+import kernel.instanceOf
+
 trait MonoidClass[A] extends SemigroupClass[A] {
   def mempty: A
 }
--- a/base/shared/src/main/scala/scalaz/basehierarchy.scala
+++ b/base/shared/src/main/scala/scalaz/basehierarchy.scala
@@ -1,6 +1,7 @@
 package scalaz
 
-import scalaz.ct.FunctorClass
+import kernel.instanceOf
+import ct.FunctorClass
 
 trait BaseHierarchy extends BaseHierarchy.BH0
 
--- a/base/shared/src/main/scala/scalaz/core/eq.scala
+++ b/base/shared/src/main/scala/scalaz/core/eq.scala
@@ -3,6 +3,8 @@ package core
 
 import scala.language.experimental.macros
 
+import kernel.instanceOf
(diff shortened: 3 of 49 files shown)


edmundnoble:
I think this is a very bad idea. As John has said previously, all we will do is sacrifice cross-package functionality for no gain.

barambani:
I understand that [the ticket](https://github.com/scalaz/scalaz/issues/1826) is far from being agreed upon so there is no reason to keep this open. I'm closing the PR then.

PR #1895, opened 2018-06-24
Missing IList API

emilypi:
Sup folks, here's a basic API for `IList`

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/ilist.scala
+++ b/base/shared/src/main/scala/scalaz/data/ilist.scala
@@ -91,27 +91,25 @@ final class IListOps[A](val self: data.IList[A]) extends AnyVal {
   def take(n: Int): IList[A] = {
     @tailrec
     def go(m: Int, acc: IList[A], as: IList[A]): IList[A] =
-      if (m <= 0) IList.reverse(acc)
-      else
-        IList.uncons(as) match {
-          case Maybe2.Empty2() => IList.reverse(acc)
-          case Maybe2.Just2(a, aas) =>
-            go(m - 1, IList.cons(a, acc), aas)
-        }
-    go(n, IList.empty, self)
+      IList.uncons(as) match {
+        case Maybe2.Empty2() => IList.reverse(acc)
+        case Maybe2.Just2(a, aas) =>
+          go(m - 1, IList.cons(a, acc), aas)
+      }
+    if (n <= 0) IList.empty
+    else go(n, IList.empty, self)
   }
 
   def drop(n: Int): IList[A] = {
     @tailrec
     def go(m: Int, as: IList[A]): IList[A] =
-      if (m <= 0) as
-      else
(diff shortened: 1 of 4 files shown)


emilypi:
@xuwei-k thanks for tagging!

base/shared/src/main/scala/scalaz/data/ilist.scala
```diff
@@ -37,6 +41,87 @@ trait IListModule {
...
+      IList.uncons(self) match {
+        case Maybe2.Empty2()    => Maybe.empty
+        case Maybe2.Just2(a, _) => Maybe.just(a)
+      }
+
+    def tailOption: Maybe[IList[A]] =
+      IList.uncons(self) match {
+        case Maybe2.Empty2()     => Maybe.empty
+        case Maybe2.Just2(_, as) => Maybe.just(as)
+      }
```
jdegoes: Could we call these `head` and `tail`?

base/shared/src/main/scala/scalaz/data/ilist.scala
```diff
@@ -37,6 +41,87 @@ trait IListModule {
...
+          case Maybe2.Just2(_, aas) =>
+            go(m - 1, aas)
+        }
+
+      if (n <= 0) self
+      else go(n, self)
+    }
+
+    def size: Int =
+      IList.foldLeft[A, Int]((b, a) => 1 + b, 0, self)
```
jdegoes: This will be faster specialized.

base/shared/src/main/scala/scalaz/data/ilist.scala
emilypi: sure - i'm on board with making the safe versions mandatory 😄

base/shared/src/main/scala/scalaz/data/ilist.scala
emilypi: Good call!

emilypi:
@alexknvl would you be okay with renaming `empty` to `nil`? We have a naming conflict with `Maybe2#empty` in `AllFunctions`, and i'm sure we'd like to inline our `empty`'s for `IList`

base/shared/src/main/scala/scalaz/data/ilist.scala
alexknvl: He means that writing it out in terms of pattern matching + `@tailrec` function will be faster, because `Int`s won't be boxed and the function will be "inlined" (manually).

emilypi: oh, sure i can do that. I thought he was referring to `@specialized` and I was like 😖

emilypi:
@jdegoes I've addressed `size` as a special case and added a few more standard methods to the bunch. I think this is relatively complete as APIs go. I'm leaving the rest of the functionality up to the typeclasses.

Issue #1885, opened 2018-06-19
moar Deriving instances


fommil:
add more instances of `Decidable` and `Alt` as per #1881

Where we have `ApplicativePlus` instances, we can probably provide an `Alt`

Somewhat related to #1884

fommil:
@azanin @enzief I'm not going to be watching this repo for updates... if you have any outstanding PRs please pester @jdegoes @hrhino and @alexknvl for reviews and to merge :smile: 

and great work so far, this is going to be super useful to have in the lib.

PR #1896, opened 2018-06-24
instances: Decidable[Order], Alt[Maybe]

azanin:
This PR is working in progress for issue #1885.

Accordingly with @fommil we decided to not implement an Alt[Maybe] instance in terms of class, as it was required.

That's because it would lead to ambiguity on implicit values.

### DIFF
--- a/core/src/main/scala/scalaz/Maybe.scala
+++ b/core/src/main/scala/scalaz/Maybe.scala
@@ -299,8 +299,8 @@ sealed abstract class MaybeInstances extends MaybeInstances0 {
 
   implicit def maybeMaxMonad: Monad[MaxMaybe] = Tags.Max.subst1[Monad, Maybe](Monad[Maybe])
 
-  implicit val maybeInstance: Traverse[Maybe] with MonadPlus[Maybe] with BindRec[Maybe] with Cozip[Maybe] with Zip[Maybe] with Unzip[Maybe] with Align[Maybe] with IsEmpty[Maybe] with Cobind[Maybe] with Optional[Maybe] =
-    new Traverse[Maybe] with MonadPlus[Maybe] with BindRec[Maybe] with Cozip[Maybe] with Zip[Maybe] with Unzip[Maybe] with Align[Maybe] with IsEmpty[Maybe] with Cobind[Maybe] with Optional[Maybe] {
+  implicit val maybeInstance: Traverse[Maybe] with MonadPlus[Maybe] with Alt[Maybe] with BindRec[Maybe] with Cozip[Maybe] with Zip[Maybe] with Unzip[Maybe] with Align[Maybe] with IsEmpty[Maybe] with Cobind[Maybe] with Optional[Maybe] =
+    new Traverse[Maybe] with MonadPlus[Maybe] with Alt[Maybe] with BindRec[Maybe] with Cozip[Maybe] with Zip[Maybe] with Unzip[Maybe] with Align[Maybe] with IsEmpty[Maybe] with Cobind[Maybe] with Optional[Maybe] {
 
       def point[A](a: => A) = just(a)
 
@@ -390,6 +390,21 @@ sealed abstract class MaybeInstances extends MaybeInstances0 {
 
       override def filter[A](fa: Maybe[A])(f: A => Boolean): Maybe[A] =
         fa.filter(f)
+
+      override def alt[A](a1: => Maybe[A], a2: => Maybe[A]): Maybe[A] = a1 orElse a2
+
+      // performance optimisation
+      override def altly2[Z, A1, A2](a1: => Maybe[A1], a2: => Maybe[A2])(f: A1 \/ A2 => Z): Maybe[Z] = a1 match {
+          case Empty() => a2 match {
+            case Empty() => Empty()
+            case j => j.map(s => f(\/-(s)))
+          }
+          case j => j.map(s => f(-\/(s)))
+        }
+
+      // performance optimisation
+      override def apply2[A, B, C](fa: => Maybe[A], fb: => Maybe[B])(f: (A, B) => C): Maybe[C] =
+        fa.flatMap(a => fb.map(b => f(a, b)))
     }
 }
 
--- a/core/src/main/scala/scalaz/Order.scala
+++ b/core/src/main/scala/scalaz/Order.scala
@@ -85,20 +85,34 @@ object Order {
 
   ////
 
-  implicit val orderInstance: Divisible[Order] = new Divisible[Order] {
-    override def contramap[A, B](r: Order[A])(f: B => A) = r.contramap(f)
+  implicit val orderInstance: Decidable[Order] = new OrderDecidableInstance
 
-    override def conquer[A] = order((_, _) => Ordering.EQ)
+  private[scalaz] class OrderDecidableInstance extends Decidable[Order] with Divisible[Order] {
+    override def contramap[A, B](r: Order[A])(f: B => A): Order[B] = r.contramap(f)
 
-    override def divide2[A, B, C](fa: =>Order[A], fb: =>Order[B])(f: C => (A, B)) =
-      order[C]{ (c1, c2) =>
-        val (a1, b1) = f(c1)
-        val (a2, b2) = f(c2)
-        fa.order(a1, a2) match {
-          case Ordering.EQ => fb.order(b1, b2)
-          case o => o
+    override def choose2[Z, A1, A2](order1: => Order[A1], order2: => Order[A2])(f: Z => A1 \/ A2): Order[Z] = order[Z] { (c1, c2) =>
+      f(c1) match {
+        case -\/(c) => f(c2) match {
+          case -\/(d) => order1(c, d)
+          case _ => Ordering.LT
+        }
+        case \/-(c) => f(c2) match {
+          case \/-(d) => order2(c, d)
+          case _ => Ordering.GT
         }
       }
+    }
+
+    override def conquer[A]: Order[A] = order((_, _) => Ordering.EQ)
+
+    override def divide2[A1, A2, Z](a1: => Order[A1], a2: => Order[A2])(f: Z => (A1, A2)): Order[Z] = order[Z] { (c1, c2) =>
+      val (x1, y1) = f(c1)
+      val (x2, y2) = f(c2)
+      a1.order(x1, x2) match {
+        case Ordering.EQ => a2.order(y1, y2)
+        case o => o
+      }
+    }
   }
 
   def fromScalaOrdering[A](implicit O: SOrdering[A]): Order[A] = new Order[A] {


core/src/main/scala/scalaz/Maybe.scala
```diff
@@ -293,19 +294,20 @@ sealed abstract class MaybeInstances extends MaybeInstances0 {
...
-  implicit val maybeInstance: Traverse[Maybe] with MonadPlus[Maybe] with BindRec[Maybe] with Cozip[Maybe] with Zip[Maybe] with Unzip[Maybe] with Align[Maybe] with IsEmpty[Maybe] with Cobind[Maybe] with Optional[Maybe] =
-    new Traverse[Maybe] with MonadPlus[Maybe] with BindRec[Maybe] with Cozip[Maybe] with Zip[Maybe] with Unzip[Maybe] with Align[Maybe] with IsEmpty[Maybe] with Cobind[Maybe] with Optional[Maybe] {
+  implicit val maybeInstance: Traverse[Maybe] with MonadPlus[Maybe] with Alt[Maybe] with BindRec[Maybe] with Cozip[Maybe] with Zip[Maybe] with Unzip[Maybe] with Align[Maybe] with IsEmpty[Maybe] with Cobind[Maybe] with Optional[Maybe] =
+    new Traverse[Maybe] with MonadPlus[Maybe] with Alt[Maybe] with BindRec[Maybe] with Cozip[Maybe] with Zip[Maybe] with Unzip[Maybe] with Align[Maybe] with IsEmpty[Maybe] with Cobind[Maybe] with Optional[Maybe] {
 
       def point[A](a: => A) = just(a)
 
       override def ap[A, B](fa: => Maybe[A])(mf: => Maybe[A => B]) =
-        mf.cata(f => fa.cata(f andThen just, empty), empty)
+        apply2(fa, mf)((a, abc) => abc(a))
```
fommil: if you do this, don't forget to implement `apply2` or you'll get an infinite loop.

core/src/main/scala/scalaz/Maybe.scala
```diff
@@ -390,6 +392,11 @@ sealed abstract class MaybeInstances extends MaybeInstances0 {
 
       override def filter[A](fa: Maybe[A])(f: A => Boolean): Maybe[A] =
         fa.filter(f)
+
+      override def alt[A](a1: => Maybe[A], a2: => Maybe[A]): Maybe[A] = altly2(a1, a2) {
```
fommil: and don't forget to implement `altly2` or you'll get an infinite loop too

fommil: for `Maybe`, it might make sense to implement `ap` and `alt` AND `apply2` / `altly2` for performance reasons. Only one set is needed, but I can imagine `Maybe` being used in perf critical code a lot.

core/src/main/scala/scalaz/Order.scala
```diff
@@ -83,22 +85,32 @@ object Order {
       override def iso: F <=> G = D
     }
 
-  ////
+  implicit val orderDecidableInstance: Decidable[Order] = new OrderDecidableInstance
 
-  implicit val orderInstance: Divisible[Order] = new Divisible[Order] {
-    override def contramap[A, B](r: Order[A])(f: B => A) = r.contramap(f)
+  private[scalaz] class OrderDecidableInstance extends Decidable[Order] {
```
fommil: this should be mixed in with the `Divisible`.

For performance reasons I think it makes a lot of sense to override `equal` too if it is not already pulled in by inheriting from the `Equal[Divisible]`

core/src/main/scala/scalaz/Maybe.scala
azanin: I've implemented altly2 and apply2.

core/src/main/scala/scalaz/Maybe.scala
```diff
@@ -397,6 +397,17 @@ sealed abstract class MaybeInstances extends MaybeInstances0 {
         case -\/(a) => a
         case \/-(a) => a
       }
+
+      override def altly2[Z, A1, A2](a1: => Maybe[A1], a2: => Maybe[A2])(f: A1 \/ A2 => Z): Maybe[Z] = a1 match {
```
fommil: you still need to implement `def alt` otherwise this is incomplete. Your previous `alt` was reasonable. It is reasonable to have optimised `alt` and `altly2` for `Maybe`.

core/src/main/scala/scalaz/Maybe.scala
```diff
@@ -5,6 +5,7 @@ import scala.util.control.NonFatal
 import scala.reflect.ClassTag
 import Ordering._
 import Isomorphism.{<~>, IsoFunctorTemplate}
+import scalaz.Maybe.{Empty, empty, just}
```
fommil: it'd be best to refer to these with relative paths (which is idiomatic), e.g. `case Maybe.Empty() =>`

core/src/main/scala/scalaz/Maybe.scala
```diff
@@ -293,19 +294,20 @@ sealed abstract class MaybeInstances extends MaybeInstances0 {
     })
   }
 
+
```
fommil: whitespace

core/src/main/scala/scalaz/Order.scala
```diff
@@ -1,6 +1,8 @@
 package scalaz
 
 ////
+import scalaz.Ordering.{GT, LT}
```
fommil: here too, it's idiomatic to use the relative path instead of adding an import at the top.

fommil:
this is looking good! Just a couple of comments so far. Do you want to do everything in one go, or raise a separate PR every time you add support for a new typeclass (or a few)?

fommil:
ok, you're getting a build failure

```
[error] java.lang.RuntimeException: different number of chunks in old and new source: Order.scala
```

This is coming from the fact that much of the boilerplate in scalaz is autogenerated. The `//` indicators are markers to let you know where your code goes vs the generated code. If you edit a block that is autogenerated, the build will fail.

Run `getTypeClasses` in `sbt` to see what changes on disk and it should hopefully be obvious.

Generally in the scalaz sources, try to avoid adding imports and unnecessary whitespace, as it is 1) pollutes the history 2) might impact somebody else's code 3) usually triggers this warning :wink:

fommil:
Now, typically I should ask you to add tests... and for typeclasses with laws that is usually a case of adding a one liner (see the MaybeTests for testing the Monad laws).

But since `Alt` and `Decidable` are currently lawless (blocked on Haskell deciding what to do!) we can't do anything.

Perhaps you could add some adhoc unit tests?

core/src/main/scala/scalaz/Maybe.scala
```diff
@@ -390,6 +392,22 @@ sealed abstract class MaybeInstances extends MaybeInstances0 {
 
       override def filter[A](fa: Maybe[A])(f: A => Boolean): Maybe[A] =
         fa.filter(f)
+
+      override def alt[A](a1: => Maybe[A], a2: => Maybe[A]): Maybe[A] = altly2(a1, a2) {
+        case -\/(a) => a
+        case \/-(a) => a
+      }
+
+      override def altly2[Z, A1, A2](a1: => Maybe[A1], a2: => Maybe[A2])(f: A1 \/ A2 => Z): Maybe[Z] = a1 match {
```
fommil: it might be worthwhile putting a line comment on apply2 / altly2 to say that they are implemented for performance. In fact you might want to implement `alt` directly for performance reasons, since it has a fairly trivial implementation.

core/src/main/scala/scalaz/Maybe.scala
fommil: I'm going to change my mind on this, let's leave the `ap` as it was. It seems to be fairly well optimised already.

fommil:
Sorry for going back and forth about ap/apply2 and alt/altly2.

Unusually, for `Maybe`, `alt` is a more natural thing to implement than `altly2`. Normally it is clear that `apply2` and `altly2` are the "intuitive" thing to implement.

I think we should just implement `alt`. I wouldn't bother implementing `altly2` or `apply2` here, but since you've done it, you may as well keep them, because they are more efficient than the default implementations! Therefore leave a line comment to this effect, e.g. `// performance optimisation`. Please back out the change to `ap`, as it appears to be optimised already.

core/src/main/scala/scalaz/Order.scala
```diff
@@ -83,22 +85,32 @@ object Order {
       override def iso: F <=> G = D
     }
 
-  ////
+  implicit val orderDecidableInstance: Decidable[Order] = new OrderDecidableInstance
 
-  implicit val orderInstance: Divisible[Order] = new Divisible[Order] {
-    override def contramap[A, B](r: Order[A])(f: B => A) = r.contramap(f)
+  private[scalaz] class OrderDecidableInstance extends Decidable[Order] with Divisible[Order] {
+    override def choose2[Z, A1, A2](order1: => Order[A1], order2: => Order[A2])(f: Z => ∨[A1, A2]): Order[Z] = order[Z] { (c1, c2) =>
```
fommil: you're using `\/[A1, A2]` instead of `A1 \/ A2`... was that IntelliJ perhaps? It's definitely more idiomatic to use the infix notation.

azanin: Changing it, IntelliJ not very clever....

azanin:
Ok I should have take care of your suggestions. 
About testing probably we could have some adhoc cases, I will think about it tomorrow :)

Thanks a lot

core/src/main/scala/scalaz/Order.scala
```diff
@@ -85,20 +85,32 @@ object Order {
...
-    override def conquer[A] = order((_, _) => Ordering.EQ)
-
-    override def divide2[A, B, C](fa: =>Order[A], fb: =>Order[B])(f: C => (A, B)) =
-      order[C]{ (c1, c2) =>
-        val (a1, b1) = f(c1)
-        val (a2, b2) = f(c2)
-        fa.order(a1, a2) match {
-          case Ordering.EQ => fb.order(b1, b2)
-          case o => o
+  implicit val orderDecidableInstance: Decidable[Order] = new OrderDecidableInstance
```
fommil: this should still be called `orderInstance`

core/src/main/scala/scalaz/Maybe.scala
```diff
@@ -390,6 +390,24 @@ sealed abstract class MaybeInstances extends MaybeInstances0 {
 
       override def filter[A](fa: Maybe[A])(f: A => Boolean): Maybe[A] =
         fa.filter(f)
+
+      override def alt[A](a1: => Maybe[A], a2: => Maybe[A]): Maybe[A] = altly2(a1, a2) {
```
fommil: since `altly2` is for perf optimisation, you need to implement `alt` directly. It should be pretty easy, just evaluate the first one and if it is non-`Empty` return it, otherwise return the second one.

core/src/main/scala/scalaz/Order.scala
```diff
@@ -85,20 +85,32 @@ object Order {
...
+
+  private[scalaz] class OrderDecidableInstance extends Decidable[Order] with Divisible[Order] {
```
fommil: it's ok, let's not bother with creating classes anymore... I don't think it is going to work out.

core/src/main/scala/scalaz/Order.scala
```diff
@@ -85,20 +85,32 @@ object Order {
 
   ////
 
-  implicit val orderInstance: Divisible[Order] = new Divisible[Order] {
-    override def contramap[A, B](r: Order[A])(f: B => A) = r.contramap(f)
```
fommil: it looks like you dropped `contramap`?

azanin:
I think we could merge this when will be ready (even if there are few new instances) and then open new PR's for other instances.
Sounds good?

core/src/main/scala/scalaz/Order.scala
azanin: Yes sorry, that was because I dropped Divisible[Order] instance in first impl. Adding again

core/src/main/scala/scalaz/Maybe.scala
azanin: You're right forgot to revert to the first impl.

fommil:
looks good, thanks! I'll merge when it goes green

fommil:
(I renamed the PR to a concise description of the features that are added)

fommil:
oh, btw, we can squash, but you might want to do it yourself and use an email address that is associated to your github account if you want to claim the creds in the Contributor Chart.

PR #1901, opened 2018-06-26
Add identity traversable, foldable and monoid instances

RaasAhsan:


base/shared/src/main/scala/scalaz/data/identity.scala
```diff
@@ -1,27 +1,34 @@
...
 
-  def identityMonad: Monad[Identity]
+  def monad: Monad[Identity]
+  def traversable: Traversable[Identity]
 
-  def identityEq[A: Eq]: Eq[Identity[A]]
-  def identityDebug[A: Debug]: Debug[Identity[A]]
+  def monoid[A: Monoid]: Monoid[Identity[A]]
+  def eq[A: Eq]: Eq[Identity[A]]
+  def debug[A: Debug]: Debug[Identity[A]]
```
hrhino: It would be cool to do this with a `Identity <~> λ[⍺ => ⍺]` or something, like we have in 7.x, but not now.

emilypi: @hrhino agreed, soon!

hrhino:
Thanks, and welcome!

PR #1904, opened 2018-06-29
[WIP] Decidable[IdT] and Divisible[IdT]

azanin:
This PR is related to #1885 by @fommil .
 - Decidable and Divisable instances for IdT.
 - Unit tests for Decidable and Divisable.

### DIFF
--- a/core/src/main/scala/scalaz/IdT.scala
+++ b/core/src/main/scala/scalaz/IdT.scala
@@ -20,7 +20,21 @@ final case class IdT[F[_], A](run: F[A]) {
     new IdT(F.ap(run)(f.run))
 }
 
-sealed abstract class IdTInstances4 {
+sealed abstract class IdTInstances6 {
+  implicit def idTDivisible[F[_]](implicit F0: Divisible[F]): Divisible[IdT[F, ?]] =
+    new IdTDivisable[F] {
+      implicit def F: Divisible[F] = F0
+    }
+}
+
+sealed abstract class IdTInstances5 extends IdTInstances6 {
+  implicit def idTDecidable[F[_]](implicit F0: Decidable[F]): Decidable[IdT[F, ?]] =
+    new IdTDecidable[F] {
+      implicit def F: Decidable[F] = F0
+    }
+}
+
+sealed abstract class IdTInstances4 extends IdTInstances5 {
   implicit def idTFunctor[F[_]](implicit F0: Functor[F]): Functor[IdT[F, ?]] =
     new IdTFunctor[F] {
       implicit def F: Functor[F] = F0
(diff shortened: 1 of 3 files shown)


core/src/main/scala/scalaz/IdT.scala
```diff
@@ -20,7 +20,21 @@ final case class IdT[F[_], A](run: F[A]) {
     new IdT(F.ap(run)(f.run))
 }
 
-sealed abstract class IdTInstances4 {
+sealed abstract class IdTInstances6 {
+  implicit def idTDivisible[F[_]](implicit F0: Divisible[F]): Divisible[IdT[F, ?]] =
```
fommil: I don't think we need to introduce extra layers in the cake here.

The reason for the extra layers is implicit priority, so things in lower numbers get found first. This is used to avoid implicit ambiguity and related to typeclass coherence.

Since your instances don't conflict with anything, they can go in the same layer of the cake.

But @xuwei-k may have more to say.

fommil: I also don't think we'd want to provide an implicit for divisible if we can provide `Decidable`. I don't think `idTDivisible` can ever be discovered.

fommil:
btw, when this effort is complete, we might be able to backport it to 7.2, including the new typeclasses in scalaz-deriving. Which would be :100:

scalacheck-binding/src/main/scala/scalaz/scalacheck/ScalazArbitrary.scala
```diff
@@ -533,6 +533,9 @@ object ScalazArbitrary extends ScalazArbitraryPlatform {
   implicit def lazyEitherTArb[F[_], A, B](implicit A: Arbitrary[F[LazyEither[A, B]]]): Arbitrary[LazyEitherT[F, A, B]] =
     Functor[Arbitrary].map(A)(LazyEitherT[F, A, B](_))
 
+  implicit def constIdTArbitrary[A, F[_]](implicit A: Arbitrary[A], APP: Applicative[F]): Arbitrary[IdT[F, A]] =
+    Functor[Arbitrary].map(A)(a => IdT[F, A](APP.pure(a)))
```
xuwei-k: Why not `implicit def idTArbitrary[A, F[_]](implicit A: Arbitrary[F[A]]): Arbitrary[IdT[F, A]]` ?

core/src/main/scala/scalaz/IdT.scala
azanin: Hi @fommil I've a doubt regarding providing or not providing ```Divisible[IdT]``` instance.

```Decidible[IdT]``` instance requires a ```Decidible[F]```.
Let's assume ```F``` to be a ```type ConstInt[A] = Const[Int, A]```, for example. 

```Const``` has a ```Divisable``` instance but doesn't have a ```Decidable``` one (at least I was not able to find an implementation. http://hackage.haskell.org/package/contravariant-1.4.1/docs/Data-Functor-Contravariant-Divisible.html ).

That's the reason why I provide divisableInstance for ```IdT```.
If we don't provide it we can't ask for ```Divisible[IdT[ConstInt,?]]``` (even if it would be possible) and of course we are not able to get a ```Decidable[IdT[ConstInt,?]]``` instance since there is no ```Decidable[ConstInt]``` available.

Am I missing something?

Thanks

fommil: oh I see. Yes then this makes sense and your implicit ordering is good too. I forgot this was a monad transformer and tries to delegate as much as possible to the underlying!

azanin: Cool. 
Btw I can't figure out why the build is failing.
All tests are passing but one job got an error related to scalajs (I think).

PR #1906, opened 2018-07-01
use fromIso for IdT instances

xuwei-k:


fommil:
you really love `.fromIso` now :smile:

PR #1910, opened 2018-07-07
New package/module structure

edmundnoble:
This PR does the following:

- Removes the `scalaz.ct` and `scalaz.algebra` packages
- Moves all type classes to `scalaz.tc`
- Moves all propositions (i.e. types which can only have one inhabitant) to `scalaz.prop`
- Moves all data types to `scalaz.data`
- Removes the `scalaz-std` project and `scalaz.std` package
- Moves all type class instances to companion objects
- Removes `scalaz.Prelude`
- Moves all data type and type class aliases out of the `scalaz.Scalaz` object and into the respective packages
- Moves the Predef replacement from the `scalaz` package object to `scalaz.Predef`.

For the following reasons:
- `scalaz.ct` and `scalaz.algebra` are not discoverable for experienced users in many cases, let alone new users.
- `scalaz.data` and `scalaz.tc` are fairly clear in what they mean, and the division between `scalaz.data`, `scalaz.tc` and `scalaz.prop` is completely unambiguous (to the extent that they could be different kinds)
- `scalaz.std` is very annoying to have to import from (speaking from experience with Scalaz 7.2 and cats), and its existence isn't justified by the instances being unlawful. They are still lawful, they just required imports, which means a) orphan instances and b) dramatically slowed implicit resolution (even for other implicits being resolved from the same file that imports them) due to the way Scala's implicit search works, but most importantly c) importing them is a development hurdle.
- If we're moving the `scalaz.std` instances to the type classes to avoid orphan instances, it doesn't make sense to have `XInstances` traits for X a data type or type class. Removing these means that there are *no* imports necessary for *any* type class instances in scalaz, just for datatypes, functions, type classes and syntax.
- Without instances needing to be imported, the difference between `scalaz.Prelude` and `scalaz.Scalaz `evaporates and `scalaz.Prelude` can be eliminated.
- The idea of importing `scalaz.Scalaz._` and getting all data types and type classes as well is very confusing; it disconnects imports from the actual package structure, making it less easy to explore scalaz based on user code.
- Providing a `Predef` replacement in the root of `scalaz` conflates a lot of things, and forces users to use it. Instead, we can shove it into its own object `Predef`. The confusion between `Predef` and `Prelude` is avoided by deleting `Prelude`.

Not sure why this doesn't show up as mergeable, but I'll correct that soon enough.

Sorry that this breaks a lot, but I figure it should be done as early as possible.


.jvmopts
```diff
@@ -0,0 +1 @@
+-XX:+UseCodeCacheFlushing
```
hrhino: is this really worth checking in?

base/shared/src/main/scala/scalaz/data/disjunction.scala
```diff
@@ -14,30 +16,12 @@ sealed trait Disjunction[L, R] {
   }
 }
 
-object Disjunction extends DisjunctionInstances with DisjunctionFunctions {
+object Disjunction extends DisjunctionFunctions {
```
hrhino: disjunction function, what's your junction?

base/shared/src/main/scala/scalaz/data/package.scala
```diff
@@ -18,6 +20,14 @@ package object data {
   type ~>[F[_], G[_]]        = ∀[λ[α => F[α] => G[α]]]
   type ~~>[F[_, _], G[_, _]] = ∀∀[λ[(α, β) => F[α, β] => G[α, β]]]
 
+  type \/[L, R] = data.Disjunction[L, R]
+  val (\/): data.Disjunction.type = data.Disjunction
```
hrhino: uh. this parses? ok.

base/shared/src/main/scala/scalaz/tc/semigroup.scala
```diff
@@ -1,13 +1,18 @@
...
 
 trait SemigroupClass[A] {
   def append(a1: A, a2: => A): A
 }
 
-trait SemigroupInstances {}
+object SemigroupClass {
+  implicit val StringSemigroup: Semigroup[String] = instanceOf(new SemigroupClass[String] {
+    def append(a1: String, a2: => String) = a1 + a2
+  })
```
hrhino: this is extraneous?

.jvmopts
edmundnoble: Nope.

base/shared/src/main/scala/scalaz/tc/semigroup.scala
edmundnoble: How so?

hrhino: doesn't the `Monoid[String]` instance resolve for this? or have I become mistaken?

jdegoes:
The deletion of `ct` and `algebra` is terrible. Big 👎 from me.

* Our `algebra` package should contain a subset of what a typical textbook on _abstract algebra_ contains, with the same or similar names, in order to maximize the discoverability of content and teach users useful mathematics.
* In the same way, our `ct` package should contain a subset of what a typical textbook on _category theory_ contains, with the same or similar names, for the same reasons.
* Any other packages should be similarly informed by pedagogical concerns.

Type classes are not a real thing. They're a language feature for encoding a thing. Rather than organizing entities by whether or not they are type classes, we should organize them _as they are already widely organized in mathematics_. It would be absurd to buy a math book on "type classes" and to learn about category theory, topology, abstract algebra, and lots of other things. It's not a useful way of organizing content.

As for the other changes:

> Moves all propositions (i.e. types which can only have one inhabitant) to scalaz.prop

👍 but want @alexknvl and @TomasMikula feedback on this one.

> Moves all data types to scalaz.data

👎 Horrible way to organize. Imagine doing this to your project's code base: throwing all ADTs into one colossal `data` package. **Just because a thing is data or a type class, doesn't mean it has no other structure. There are useful ways to organize both data and type classes.**

> Removes the scalaz-std project and scalaz.std package

I'm ambivalent on this one, but I feel if people feel strongly, they should comment on the pull requests in which such issues are extensively discussed, rather than *ignoring them* and undoing the work later.

> Moves all type class instances to companion objects

Big 👍 assuming it doesn't cause any inference problems.

> Removes scalaz.Prelude

What replaces this?

> Moves all data type and type class aliases out of the scalaz.Scalaz object and into the respective packages

👎 This is strictly less flexible, since there's no way to import everything. There's no need for any of us to force their style of programming onto others. 

I and many others have stated we want a one-import-to-rule-them-all. 

> Moves the Predef replacement from the scalaz package object to scalaz.Predef.

👍

base/shared/src/main/scala/scalaz/tc/semigroup.scala
edmundnoble: Good point.

edmundnoble:
"Type classes are not a real thing" doesn't exactly encourage me. This sounds a bit ivory-tower; type classes are not an "encoding trick". Type classes are a nice way to abstract for certain cases and hit a particularly satisfying power/weight ratio. Even if conceptually you are *able* to split things up, that doesn't mean there is a benefit to doing so, nor that it's unambiguous.

In addition, several concepts fit into multiple divisions, as we've mentioned previously. What do you fit `Free[Monoid, A]` (generalized free) into? `FreeH[Monad, F, A]`? Especially in the case of higher-order abstractions it becomes clear that this division isn't useful, but even otherwise you need to come up with a place to put `Endo` (aren't endomorphisms abstract algebra? what about *higher-kinded* endomorphisms?), and then it becomes clear that the division is fundamentally between the kinds of the type classes. And there are many more kinds than `(* -> *) -> *` and `* -> *`; furthermore it's anti-pedagogic, because you have to know what your type class is before you can find it.

As well, splitting up datatypes and type classes into their own areas within their code is already a well-established convention: it's what everyone does in Haskell, and there it works perfectly (though if you have more fine-grained criteria for splitting up those types, multiple `.Types` and ` .Class` modules are often used, which would be analogous to multiple packages ending in `.data`).

There is no replacement for `scalaz.Prelude`; as I said above, the only things contained there and not contained in `scalaz.Scalaz` were type class instances, which are now all in companions.

> This is strictly less flexible, since there's no way to import everything. There's no need for any of us to force their style of programming onto others.

This is not strictly less flexible. Everything you sacrifice in the name of "flexibility" means less flexibility in a dual sense; every decision you make forces a style onto people. By providing me with a catch-all import, you are practically forcing me to use it. And what if I only want some of the names in scope? What if I want one of the `*Class` types in scope? What if I want one of the few arbitrary exceptions we've made to this strategy in scope, and I don't know which package it's from because I've been insulated from the scalaz package structure?

Already, this strategy has hidden a *massive* amount of cyclic dependencies within scalaz. I shudder to think what it might hide in user code.

I'd much prefer to:
- Have imports be connected to the layout of the code in scalaz, so people can meaningfully explore it
- Make it clear to users where to import something from, before they import from it
- Avoid using massive numbers of type aliases to "get the best of both worlds" from both a flat and nested package structure, when all this will do is confuse users. We have to choose one.

Note that with the removal of explicitly-imported instances, `scalaz.Scalaz._` just imports syntax and functions. Scalaz is much less import-heavy with this anyway, I think; but please use it to form an opinion.

> I'm ambivalent on this one, but I feel if people feel strongly, they should comment on the pull requests in which such issues are extensively discussed, rather than ignoring them and undoing the work later.

This is actually part of "moves all instances into companions", so from what I understand you support it; however, I would appreciate links to these PRs you mention, though I just commented on #1872 in case.

With that said, I'd appreciate if some more of the feedback originated from toying around with scalaz in the REPL to see practically how using it changes with this PR.

jmcardon:
@jdegoes 

> Our algebra package should contain a subset of what a typical textbook on abstract algebra contains, with the same or similar names, in order to maximize the discoverability of content and teach users useful mathematics.

We are not teaching mathematics because we teach a user to use `mempty`. We're not even close, and Scala is the furthest thing from a vehicle into proper rigor.

>  In the same way, our ct package should contain a subset of what a typical textbook on category theory contains, with the same or similar names, for the same reasons.

Similarly here, we're _far_ from rigorous. We're more like teaching a concept and how it related to programming, but we're not teaching mathematics by teaching people scalaz.

> -1 Horrible way to organize. Imagine doing this to your project's code base: throwing all ADTs into one colossal data package. Just because a thing is data or a type class, doesn't mean it has no other structure. There are useful ways to organize both data and type classes.

This feels like wanting to organize things like mathematics not because it's useful but because "look at how mathy and rigorous we are here in scalaz".

jdegoes:
^^^ I have much more to say but not much time. I will get back this week though.

Milyardo:
Category Theory is a subdiscipline of abstract algebra, so I would think having both `algebra` and `ct` packages is unnecessary. I wouldn't mind if just `algebra` stuck around. 

With this change would other scalaz projects be expected to organize typeclasses and data type similarly?

edmundnoble:
@Milyardo I myself think abstract algebra is a subdiscipline of category theory, but tbh there is no right answer here because there are a *lot* of circular conceptual dependencies between CT and algebra (and *within* CT and algebra). I think there's less likely to be consensus on this than on what is and isn't a type class.

I don't know that this makes sense in general. It would be convenient for readers to have consistency and locality, but for some projects I think it might make things harder to navigate. Not sure. I do it for testz, though there I literally have a `data.scala`.

aloiscochard:
Hey all, just my 2cts,

I did not read all the discussions but I have to say I really appreciate @edmundnoble efforts here to try to keep the design as simple as he was initially.

Having sub-packages is definitely an unnecessary practice, it just makes newcomers more confuse and does not bring anything about modularity (this is just arbitrary sub-grouping, and it's not like it helps maintaining the library).

On big point of the initial Scalaz8 design was to avoid the freaking mess with imports, this means taking as much as possible benefits from "companion" implicit resolutions, it seems like this PR ensure that this design principle is followed and I like it.

@jdegoes I really don't understand your point about "Not throwing everything in data/ct package", that was one of the key decision during initial design, it was discussed in length on IRC on PR with many people... I'm surprised to see it debated again. I really don't see any benefits adding an arbitrary structure here, that again don't add anything in term of modularity... It just makes everyone more confuse.

If you want to create a subpackage, you should ask yourself why you are not creating a new module then (by module I mean sbt project)... If you can't then I believe you need a very strong and meaningful reason to create subpackage.

aloiscochard:
> I'm ambivalent on this one, but I feel if people feel strongly, they should comment on the pull requests in which such issues are extensively discussed, rather than ignoring them and undoing the work later.

That's very easy to say and not fair, I could tell you the same John for any change you do from the original design and tell you "Why did you not comment on my PRs at the time?"... I don't think that makes any sense.

jdegoes:
I think the removal of packages was a terrible idea. Exactly as terrible as making all our packages:

```scala
package scalaz.traits
package scalaz.classes
package scalaz.objects
```

In the misguided notion that this "simplifies" things because after all, "everything is a trait, class, or object." That's true, everything is a trait, class, or object, but there are degrees of structure beyond traits, classes, and objects. Package organization should reflect that structure and help guide people toward what they're looking for. This package organization fails to do that.

Finally, it seems that most people who _have already left_ or who are _in the process of leaving_ active Scalaz contribution are have the strongest opinions while having no skin in the game. Ultimately how Scalaz 8 is organized depends only on people actively contributing code to the project. Which means, @edmundnoble, you better damn well contribute code to Scalaz for quite some time. 😆

aloiscochard:
@jdegoes welcome to open source!

Users/contributors generally knows if the abstraction they are looking for is a class or datatype (and identity them based on their unique names).

OTOH I don't think it's worth asking them to understand the relationship between abstract algebra and category theory to find an abstraction, I could take Haskell as an example but I don't want people to throw stones at me.

jdegoes:
@aloiscochard Haskell cleanly and clearly organizes a huge body of abstractions. It's not perfect, but it represents an attempt at reflecting the mathematical structure of the abstractions in the package organization, which helps people more easily find what they're looking for (e.g. `Control.Monad.Trans`). (Ed Kmett did a better job with `category-extras`, which has no become lots of little projects with fairly good organization.)

aloiscochard:
@jdegoes funny you did not use `Data` for your example. (ref: https://wiki.haskell.org/Hierarchical_module_names)

Also, I don't see how `Control` match with a specific mathematical domain, unlike "Category Theory" or "Algebra".

I did always understood it as "control flow", but maybe I'm too much of an engineer to get it's true and deep mathematical meaning.

PR #1913, opened 2018-07-10
Decidable[MaybeT]

azanin:
This is related to #1885 by @fommil .
I've implementeted `Decidable `instance for `MaybeT`.

I noticed that is possible to provide instances through `Decidable.fromIso`.
Anyway for `Decidable[MaybeT[F, ?]]` we require an instance of `Divisable[F]`.
Right now `Decidable.fromIso` requires an `implicit Decidable[F]`, so in this case we could have a problem. Any suggestion on this?

Furthermore, I could also refactor in this PR the provisioning of the other instances through `fromIso `as well if you prefer.

### DIFF
--- a/core/src/main/scala/scalaz/MaybeT.scala
+++ b/core/src/main/scala/scalaz/MaybeT.scala
@@ -141,6 +141,11 @@ sealed abstract class MaybeTInstances extends MaybeTInstances0 {
 
   implicit def maybeTShow[F[_], A](implicit F0: Show[F[Maybe[A]]]): Show[MaybeT[F, A]] =
     Contravariant[Show].contramap(F0)(_.run)
+
+  implicit def maybeTDecidable[F[_]](implicit F0: Divisible[F]): Decidable[MaybeT[F, ?]] =
+    new MaybeTDecidable[F] {
+      implicit def F: Divisible[F] = F0
+    }
 }
 
 object MaybeT extends MaybeTInstances {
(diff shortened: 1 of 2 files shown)


fommil:
The `fromIso` things are relatively new and boilerplate reduction was always a goal :smile: 

That said, we must be sure that any use of `fromIso` can be replaced with an optimised version in the future, without harming bincompat, incase the need arises. I suspect we are ok since these are not exposed as part of the public API.

azanin:
Hi @fommil , I was trying to proceed with fromIso refactor.
Unfortunately I'm  stuck with a problem.

I've definded an iso like this:

```scala
def iso[F[_]]: MaybeT[F, ?] <~> Lambda[A => F[Maybe[A]]] = new IsoFunctorTemplate[MaybeT[F, ?], Lambda[A => F[Maybe[A]]]] {
    def from[A](ga: F[Maybe[A]]): MaybeT[F,A] = MaybeT(ga)
    def to[A](fa: MaybeT[F,A]): F[Maybe[A]] = fa.run
  }
```
With this isomorphism, for example, I'm able to provide instances of `Functor[MaybeT[F,?]]` : 
```scala
 implicit def maybeTFunctor[F[_]](implicit F0: Functor[F]): Functor[MaybeT[F, ?]] = {
    Functor.fromIso[MaybeT[F, ?], Lambda[A => F[Maybe[A]]]](MaybeT.iso[F])(F0.compose(Functor[Maybe]))
  }
```

As you can see I need to compose `F0` with `Functor[Maybe]` because `Functor.fromIso` requires an `implicit Functor[G]`.

My problem is with `MonadError` (and related) instances in which I can't compose instances (like in `Functor` case) to provide a `MonadError[G, E`].

A solution could be to implement a proper instance like in this way:
```scala
val moerr = new MonadError[ Lambda[A => F[Maybe[A]]], E] {
        def point[A](a: => A): F[scalaz.Maybe[A]] = ???
        def bind[A, B](fa: F[scalaz.Maybe[A]])(f: A => F[scalaz.Maybe[B]]): F[scalaz.Maybe[B]] = ???
        def handleError[A](fa: F[scalaz.Maybe[A]])(f: E => F[scalaz.Maybe[A]]): F[scalaz.Maybe[A]] = ???
        def raiseError[A](e: E): F[scalaz.Maybe[A]] = ???
      }

      MonadError.fromIso[MaybeT[F, ?], Lambda[A => F[A]], E](MaybeT.iso[F])(moerr)
```
but I don't see the real advantage to use fromIso then.

Am I missing something? Any suggestion on this?

fommil:
hmm, yeah, I hit something similar myself recently. Nah, forget iso, sorry for the digression!

azanin:
Ok I think that PR should be ready then.
I'll proceed with other instances in separate PR

fommil:
:tada: :clap:

Issue #1915, opened 2018-07-13
Flatten ApplicativeBuilder


cryeo:
`ApplicativeBuilder`, which is the wrapper class to construct applicative expressions, is adopted a nested structure as follows. 

- `ApplicativeBuilder3` is defined as an inner class of `ApplicativeBuilder`.
- `ApplicativeBuilder{N}` is defined as an inner class of `ApplicativeBuilder{N-1}` for 4 <= N <= 12.

This structure can reduce code redundancy, but it makes class names related with `ApplicationBuilder` too long.
It leads to `NoSuchMethodError` at runtime when `-Xmax-classfile-name` is set with not enough value.

How about flattening it?

```scala
scala> import scalaz._
import scalaz._

scala> import Scalaz._
import Scalaz._

scala> val x: Option[Int] = 1.some
x: Option[Int] = Some(1)

scala> (x |@| x).getClass.getName.some.map(c => (c.length, c)) // ApplicationBuilder2
res77: Option[(Int, String)] = Some((32,scalaz.syntax.ApplicativeBuilder))

scala> (x |@| x |@| x).getClass.getName.some.map(c => (c.length, c)) // ApplicativeBuilder3
res78: Option[(Int, String)] = Some((52,scalaz.syntax.ApplicativeBuilder$ApplicativeBuilder3))

scala> (x |@| x |@| x |@| x).getClass.getName.some.map(c => (c.length, c)) // ApplicativeBuilder4
res79: Option[(Int, String)] = Some((72,scalaz.syntax.ApplicativeBuilder$ApplicativeBuilder3$ApplicativeBuilder4))

scala> (x |@| x |@| x |@| x |@| x).getClass.getName.some.map(c => (c.length, c)) // ApplicativeBuilder5
res80: Option[(Int, String)] = Some((92,scalaz.syntax.ApplicativeBuilder$ApplicativeBuilder3$ApplicativeBuilder4$ApplicativeBuilder5))

scala> (x |@| x |@| x |@| x |@| x |@| x).getClass.getName.some.map(c => (c.length, c)) // ApplicativeBuilder6
res81: Option[(Int, String)] = Some((112,scalaz.syntax.ApplicativeBuilder$ApplicativeBuilder3$ApplicativeBuilder4$ApplicativeBuilder5$ApplicativeBuilder6))

scala> (x |@| x |@| x |@| x |@| x |@| x |@| x).getClass.getName.some.map(c => (c.length, c)) // ApplicativeBuilder7
[...]

PR #1916, opened 2018-07-13
Flatten ApplicativeBuilder

cryeo:
This is related to #1915


fommil:
I like this, but it will conflict with some changes in scalaz 7.2... that's ok, they are easy to migrate.

cryeo:
Thank you for reviewing.


> but it will conflict with some changes in scalaz 7.2

I will try to make PR for 7.2.x after this PR is accepted.


> can you please prefer sealed abstract class instead of sealed trait? (bincompat)

Sorry, I did not consider binary compatibility. 
This is my rough understanding: The trait with default implementation is going to break binary compatibility. Is it correct?

Then, It seems to be better to change `trait ApplicativeBuilder` into `abstract class`.
However, `sealed trait ApplicativeBuilderOps` is not able to change cause it is used as mix-in.

> one downside is that now we are exporting a lot more symbols (more classes). Can you think of any way of hiding them from downstream users?

A simple way of hiding them is to define them in companion object like a following code. But I'm not sure whether it is good or not.

```scala
object ApplicativeBuilder {
  // Move classes(ApplicativeBuilderN) into companion object
  final class ApplicativeBuilder2
  ...
  final class ApplicativeBuilder12

  // Add apply methods for wrapper classes
  def apply[F, A, B](fa: F[A], fb: F[B])(implicit ap: Apply[F]): ApplicativeBuilder2[F, A, B] = new ApplicativeBuilder2(fa, fb)
  ...
}

**Comment by xuwei-k** [comment]
> I will try to make PR for 7.2.x after this PR is accepted.

Can't change 7.2.x for binary compatibility. I think @fommil say we should forward port https://github.com/scalaz/scalaz/pull/1891 from 7.2. to 7.3.x

> This is my rough understanding: The trait with default implementation is going to break binary compatibility. Is it correct?

You`re right. if Scala 2.11

**Comment by cryeo** [comment]
Thank you for description.

> Can't change 7.2.x for binary compatibility. I think @fommil say we should forward port #1891 from 7.2. to 7.3.x

I totally misunderstood about that. I see!

**Comment by hrhino** [comment]
I vaguely recall that you should be able to make all of the `ApplicativeBuilderN` classes `private[scalaz]` so that they won't show up in the public API, but you'll still be able to call them.

At the very least, in 2.12.6 this works:

```scala
package a {
  object X { def y: Y = new Y }
  private[a] class Y { def z: Z = new Z }
  private[a] class Z { def value: Int = 1 }
}

package b {
  object Test { def test() = println(a.X.y.z.value) }
}
```

cryeo:
Thanks for your review.

It seems that adding access modifier to classes `ApplicativeBuilderN` is definitely simple way to hide them from downstream users.

I'll add it.

Issue #1886, opened 2018-06-19
Augment IO and IO.type with helper methods for Disjunction, Maybe


jdegoes:
Now that Scalaz 8 depends on ZIO (which defaults to `Either` and `Option`), some convenience functions are missing. This ticket is to introduce such convenience functions by adding methods to `IO` and `IO.type`. Any method which currently uses `Either` or `Option` or `List` should have an analogue that uses `\/`, `Maybe`, or `IList`.

ktonga:
You forgot `TraversableOnce` in the list of types it defaults to :smile:

ktonga:
Do you think we should introduce a `newtype` to implement `Applicative` (without `Monad`) ala `Validation` to combine using `par` instead of `bind`?

jdegoes:
@ktonga Yes, for sure. `ParallelIO` perhaps.

ktonga:
Then we get for free sequenced and parallel flavors for all the methods we have accepting `TraversableOnce` added recently

sideeffffect:
Hmm, is `ParallelIO` `newtype` really necessary? :thinking:
I mean, if we would implement `Applicative`'s `ap` in terms of `Monad`'s `flatMap` then yes, it would be sequential.
But if we implement `Applicative`'s `ap` using `par`, then every time we use the `Applicative` instance, we would be running code in parallel. And that would imply no need for any special `newtype`s.
Parallel implementation of `ap` in `Applicative` and sequential for `flatMap` in `Monad` should be able to live peacefully together.
Right? Or am I missing something?

jdegoes:
@sideeffffect A `Parallel`-like new type is necessary because `ap` must be consistent with `bind` for monads.

PR #1917, opened 2018-07-16
Add convenience functions for IO

AL333Z:
This PR is related to https://github.com/scalaz/scalaz/issues/1886.
Basically, it's just porting some convenience functions using scalaz adts (Either => \\/, Option => Maybe, etc..).

My remaining concerns are about:
 - implementing `traverse`, `sequence`, `parTraverse` and `parAll` or not, since we may just use traversable syntax and get them for free

Thanks guys, coding under scalaz compiler restrictions is a really nice experience :)

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/io.scala
+++ b/base/shared/src/main/scala/scalaz/data/io.scala
@@ -1,0 +1,96 @@
+package scalaz
+package data
+
+import scalaz.zio.{ ExitResult, Fiber, IO }
+import scalaz.tc.{ Foldable, Unfoldable }
+
+import scala.{ AnyVal, Nothing, Throwable, Unit }
+
+object io {
+
+  implicit class IOObjOps(private val ioObj: IO.type) extends AnyVal {
+
+    final def absolvez[E, A](v: IO[E, E \/ A]): IO[E, A] =
+      v.flatMap(fromDisjunction)
+
+    final def fromDisjunction[E, A](v: E \/ A): IO[E, A] =
+      v.fold(IO.fail, IO.now)
+
+    final def fromMaybe[A](v: Maybe[A]): IO[Unit, A] =
+      Maybe.fold(v)(IO.now, IO.fail(()))
+
+    final def requirez[E, A](error: E): IO[E, Maybe[A]] => IO[E, A] =
+      (io: IO[E, Maybe[A]]) =>
+        io.flatMap { maybe =>
+          Maybe.fold[A, IO[E, A]](maybe)(IO.now[A], IO.fail[E](error))
+      }
+
+    final def forkAllz[E, A, F[_]](
+      as: F[IO[E, A]]
+    )(implicit f: Foldable[F], u: Unfoldable[F]): IO[Nothing, Fiber[E, F[A]]] =
+      f.foldRight(as, IO.point(Fiber.point[E, IList[A]](IList.empty))) { (aIO, asFiberIO) =>
+          asFiberIO.par(aIO.fork).map {
+            case (asFiber, aFiber) =>
+              asFiber.zipWith(aFiber)((as, a) => a :: as)
+          }
+        }
+        .map(
+          fas =>
+            fas.zipWith(Fiber.point(())) { (res, _) =>
+              u.fromList(res)
+          }
+        )
+
+    final def raceAllz[E, A, F[_]](io: IO[E, A], ios: F[IO[E, A]])(implicit f: Foldable[F]): IO[E, A] =
+      f.foldLeft[IO[E, A], IO[E, A]](ios, io)(_ race _)
+
+    final def reduceAllz[E, A, F[_]](a: IO[E, A], as: F[IO[E, A]])(f: (A, A) => A)(implicit fd: Foldable[F]): IO[E, A] =
+      fd.foldLeft(as, a) { (l, r) =>
+        l.par(r).map(f.tupled)
+      }
+
+    final def mergeAllz[E, A, B, F[_]](in: F[IO[E, A]])(zero: B, f: (B, A) => B)(implicit fd: Foldable[F]): IO[E, B] =
+      fd.foldLeft[IO[E, A], IO[E, B]](in, IO.point[B](zero))((acc, a) => acc.par(a).map(f.tupled))
+
+    final def terminate0z[F[_]](ts: F[Throwable])(implicit f: Foldable[F]): IO[Nothing, Nothing] =
+      IO.terminate0(f.toList(ts))
+
+    final def unsandboxz[E, A, F[_]](v: IO[F[Throwable] \/ E, A])(implicit f: Foldable[F]): IO[E, A] =
+      v.catchAll[E, A] {
+        case \/-(e)  => IO.fail(e)
+        case -\/(ts) => IO.terminate0z(ts)
+      }
+
+  }
+
+  implicit class IOOps[E, A](private val io: IO[E, A]) extends AnyVal {
+    final def attemptz: IO[Nothing, E \/ A] =
+      io.redeem[Nothing, E \/ A](e => IO.now(-\/(e)), a => IO.now(\/-(a)))
+
+    final def raceBothz[E1 >: E, B](that: IO[E1, B]): IO[E1, A \/ B] =
+      io.raceWith(that)((a, fiber) => fiber.interrupt.const(-\/(a)), (b, fiber) => fiber.interrupt.const(\/-(b)))
+
+    final def or[E2, B](that: => IO[E2, B]): IO[E2, A \/ B] =
+      io.redeem(_ => that.map(\/-(_)), e => IO.now(-\/(e)))
+
+    final def sandboxedz[F[_]: Foldable](implicit u: Unfoldable[F]): IO[\/[F[Throwable], E], A] =
+      io.run.flatMap[\/[F[Throwable], E], A](_ match {
+        case ExitResult.Completed(value) =>
+          IO.now(value)
+        case ExitResult.Failed(error, _) =>
+          IO.fail(\/-[F[Throwable], E](error))
+        case ExitResult.Terminated(ts) =>
+          IO.fail(-\/[F[Throwable], E](u.fromList(IList(ts: _*))))
+      })
+
+    final def sandboxWithz[E2, B, F[_]: Foldable](
+      f: IO[\/[F[Throwable], E], A] => IO[\/[F[Throwable], E2], B]
+    )(implicit u: Unfoldable[F]): IO[E2, B] =
+      f(io.sandboxedz).catchAll[E2, B] {
+        case \/-(e)  => IO.fail(e)
+        case -\/(ts) => IO.terminate0z(ts)
+      }
+
+  }
+
+}
--- a/tests/src/main/scala/scalaz/tests/IOTests.scala
+++ b/tests/src/main/scala/scalaz/tests/IOTests.scala
@@ -1,15 +1,18 @@
 package scalaz
 package tests
 
-import scala.{ List, Range, StringContext }
+import scala.{ Exception, Int, List, Nothing, Range, StringContext, Throwable, Unit }
 import scala.Predef.String
-
 import Scalaz._
-
-import testz._, z._
-import laws._, FunctorLaws._
-
+import testz._
+import z._
+import laws._
+import FunctorLaws._
+import scalaz.data._
+import scalaz.data.io._
 import zio._
+import scala.concurrent.duration._
+import Result._
 
 object IOTests {
 
@@ -41,10 +44,12 @@ object IOTests {
     seeds.flatMap(s => endos.map(_(s)))
   }
 
-  object RTS extends RTS
+  val exampleError    = new Exception("Oh noes!")
+  val interruptCause1 = new Exception("Oh noes 1!")
 
-  def ioEql[E, A](l: IO[E, A], r: IO[E, A]): Result =
-    assert(RTS.unsafeRunSync(l.attempt) == RTS.unsafeRunSync(r.attempt))
+  object RTS extends RTS {
+    override def defaultHandler: List[Throwable] => IO[Nothing, Unit] = i => IO.traverse(i)(_ => IO.unit) *> IO.unit
+  }
 
   object fns {
     val s1 = (s: String) => IO.point(s"foo$s")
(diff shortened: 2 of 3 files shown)


base/shared/src/main/scala/scalaz/data/io.scala
```diff
@@ -0,0 +1,109 @@
...
+  )(fn: A => IO[E, B])(implicit cbf: CanBuildFrom[F[A], B, F[B]]): IO[E, F[B]] =
+    implicitly[Foldable[F]]
+      .foldLeft(in, Ref[E, mutable.Builder[B, F[B]]](cbf(in))) {
+        case (ioRef, a) =>
+          ioRef.par(fn(a)).flatMap {
+            case (ref, b) => ref.modify(_ += b) *> IO.point(ref)
+          }
+      }
+      .flatMap(_.read)
+      .map(_.result())
```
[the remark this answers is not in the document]
AL333Z: Hi @sideeffffect , I noticed your PR yesterday and I had the chance to play with it for a while, and
I wasn't able to proof the incorrectness of your implementation :)

If your assumptions hold, I'll definitely amend my implementation as well, removing `Ref` and cutting off some complexity (not to mention performance..).
Will also need to fix the build, since scalaz-zio apis are moving, but that's another story.

base/shared/src/main/scala/scalaz/data/io.scala
```diff
@@ -0,0 +1,109 @@
...
+      .map(as => as.zipWith(Fiber.point(())) { case (as, _) => as.result })
+
+  final def traverse[E, A, B, F[_]: Foldable](
+    in: F[A]
+  )(fn: A => IO[E, B])(implicit cbf: CanBuildFrom[F[A], B, F[B]]): IO[E, F[B]] =
+    implicitly[Foldable[F]]
+      .foldLeft(in, IO.point[E, mutable.Builder[B, F[B]]](cbf(in)))((io, b) => io.zipWith(fn(b))(_ += _))
+      .map(_.result())
+
+  def parTraverse[E, A, B, F[_]: Foldable](
```
sideeffffect: `final`?

AL333Z:
I pulled the latest updates from scalaz-zio and took the chance to amend parTraverse as well.
The build is now red due to `IOInstances` still depending on an API that has been changed in zio.

edmundnoble:
Is the intention here to make `absolve` a top-level function, visible if you've imported `Scalaz._`?

AL333Z:
Hi @edmundnoble and @jdegoes, thanks for the feedback.

It should be fine now.
Still need to wait for `Unfoldable` to complete this PR.

base/shared/src/main/scala/scalaz/data/io.scala
```diff
@@ -0,0 +1,109 @@
...
+      fd.foldLeft(as, a) { (l, r) =>
+        l.par(r).map(f.tupled)
+      }
+
+    final def mergeAll[E, A, B, F[_]](in: F[IO[E, A]])(zero: B, f: (B, A) => B)(implicit fd: Foldable[F]): IO[E, B] =
+      fd.foldLeft[IO[E, A], IO[E, B]](in, IO.point[B](zero))((acc, a) => acc.par(a).map(f.tupled))
+  }
+
+  implicit class IOOps[E, A](private val io: IO[E, A]) extends AnyVal {
+    final def attempt[E2]: IO[E2, E \/ A] =
```
jdegoes: This should perhaps be `IO[Nothing, E \/ A]`, with no `E2` type parameter.

base/shared/src/main/scala/scalaz/data/io.scala
```diff
@@ -0,0 +1,109 @@
...
+    final def require[E, A](error: E): IO[E, Maybe[A]] => IO[E, A] =
+      (io: IO[E, Maybe[A]]) =>
+        io.flatMap { maybe =>
+          Maybe.fold[A, IO[E, A]](maybe)(IO.now[A], IO.fail[E](error))
+      }
+
+    // TODO replace CanBuildFrom with Unfoldable
+    final def forkAll[E, A, F[_]](
+      as: F[IO[E, A]]
+    )(implicit f: Foldable[F], cbf: CanBuildFrom[F[IO[E, A]], A, F[A]]): IO[E, Fiber[E, F[A]]] =
```
jdegoes: Yeah let's replace this by `Unfoldable` as soon as your other PR is merged. 😄

base/shared/src/main/scala/scalaz/data/io.scala
```diff
@@ -0,0 +1,109 @@
...
+import scala.{ AnyVal, Either, Nothing, Option, Unit }
+import Predef.identity
+import scala.collection.generic.CanBuildFrom
+import scala.collection.mutable
+import scala.inline
+
+object io {
+
+  // TODO check if this can be optimized
+  private final def _bracket0[E, A, B](
```
jdegoes: This private method is only used in one place, why not inline it below?

base/shared/src/main/scala/scalaz/data/io.scala
AL333Z: Should be pretty easy, since `Foldable` has a `toList` (which i'm gonna update returning an IList) and `Unfoldable` has a `fromList`.

But I'm afraid this approach will be inefficient (but elegant, of course).
What're your thoughts about it?

AL333Z:
I'm almost there, but I think I spotted a limit of this approach.
Unfoldable's `fromList` is returning a ` Maybe[F[B]]`, indicating that the unfold operation might fail.
`traverse` is expecting an `F[B]`, instead.

I guess we'd need to find a way to flatten the external Maybe?
\cc @jdegoes @ktonga 

```scala
    final def traverse[E, A, B, F[_]](
      in: F[A]
    )(fn: A => IO[E, B])(implicit f: Foldable[F], u: Unfoldable[F]): IO[E, F[B]] =
      f.foldRight[A, IO[E, IList[B]]](in, IO.sync(IList.empty[B])) { (a, io) =>
          fn(a).zipWith(io)((b, bs) => b :: bs)
        }
        .map[F[B]](u.fromList(_) match {
          case Maybe.Just(bs) => bs
          case Maybe.Empty()    => ??? // <- here!
        })
```

ktonga:
`_.toOption.get` :trollface: 

In all seriousness, why do we need the intermediate list in first place?
Couldn't we constrain to `Applicative` (A) and `PlusEmpty` (PE) and fold like this:

```scala
F.foldRight(in, IO.pure(PE.empty[B])) { (a, io) =>
  fn(a).zipWith(io)((b, bs) => PE.plus(A.pure(b), bs))
}
```

Just asking, maybe someone smarter than me will say why that's stupid.

AL333Z:
An easy solution may come introducind a Monoid on `F[B]`, but I'm still not sure this is what we really want.

```scala
final def traverse[E, A, B, F[_]](in: F[A])(fn: A => IO[E, B])(
      implicit f: Foldable[F], u: Unfoldable[F], m: Monoid[F[B]]): IO[E, F[B]] =
      f.foldRight[A, IO[E, IList[B]]](in, IO.sync(IList.empty[B])) { (a, io) =>
          fn(a).zipWith(io)((b, bs) => b :: bs)
        }
        .map[F[B]](u.fromList(_) match {
          case Maybe.Just(bs) => bs
          case Maybe.Empty()    => m.mempty
        })
```

@ktonga I guess that your proposal may be more performant than folding and unfolding an F..
We'll need to introduce a PlusEmpty, of course (related: #1941).

jdegoes:
Why can't `Unfoldable` promise to return something? We can always change the type of unfoldable to guarantee it can produce something even from an empty structure.

AL333Z:
Yep, that would look nicer, of course.
We’ll need a notion of empty for F (monoid?).

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ktonga:
Why are we writing this functions instead of getting them for free, for instance, from traverse syntax? `IO` is `Applicative` already.

ktonga:
I do `myIList.traverse(myFuncReturningIO)` all the time with sz 7.2. And there is also a `parTraverse` Sam added that is really sweet.

base/shared/src/main/scala/scalaz/data/io.scala
```diff
@@ -0,0 +1,109 @@
...
+      (a: A, o: Option[Either[E, B]]) => release(a, Maybe.fromOption(o.map(Disjunction.fromEither[E, B])))
+    IO.bracket0[E, A, B](acquire)(release1)(use)
+  }
+
+  implicit class IOObjOps(private val ioObj: IO.type) extends AnyVal {
+
+    final def absolve[E, A](v: IO[E, E \/ A]): IO[E, A] =
+      v.flatMap(fromEither)
+
+    final def fromEither[E, A](v: E \/ A): IO[E, A] =
```
ktonga: `fromDisjunction`

AL333Z:
Hi @jdegoes and @ktonga ,

@jdegoes I can of course update `Unfoldable`, like this i guess:

```scala
trait UnfoldableClass[T[_]] {
  def unfoldRight[A, B](f: B => (A, Maybe[B]))(b: B): T[A]
  ...
}
```

That will definitely solve the matter, but at what cost?
We'll need to fold F, and then unfold, building an intermediate structure), and i feel that this may bring in a significant performance overhead (which can be measured, of course..).

In the end I think that what @ktonga is proposing is just what we want, and we already have those in (or at least, most of it).

So, we already have `traverse`, `sequence`, `parTraverse` and `parAll` (aka parSequence) for free.
The only one remaining is `forkAll`, but I guess it's not a big deal to implement that in term of Foldable, Traversable and Applicative (i guess..). 

Thoughts?

edmundnoble:
I would suggest that instead of altering `Unfoldable`, `Unfoldable1` be added too, *if* we want to do that.

Also, I believe using `Foldable` and `Unfoldable` for the same type has no point, so I think that means I agree with @ktonga.

There's another point to be made here: we could implement this using different `F[_]` and `G[_]` with `Foldable[F]` and `Unfoldable[G]` which would give us the same expressiveness as `CanBuildFrom`. However, `G[_]` will basically never be inferred, and it's more complicated than it should be IMO.

jdegoes:
@edmundnoble I agree. What we have right now is `Unfoldable1` and not the dual of `Foldable` (but rather `Foldable1`. If we fix this problem, the current one goes away.

Btw I think `Foldable` + `Unfoldable` could make an OK collection. Maybe with some faster slicing capabilities, e.g. something like:

```scala
trait Slice[A] extends Monoid[A] {
  // Lossless: forall n a. a === slice(a, n)._1 <> slice(a, n)._2
  // Empty: forall n. slice(mempty, n) === mempty
  // Etc.
  def slice(a: A, n: Int): (A, A)
}

trait Collection[F] extends Foldable[F] with Unfoldable[F] {
  implicit def slice[A]: Slice[F[A]]
}
```

AL333Z:
Actually, our current definition for `Unfoldable` is

```scala
def unfoldRight0[A, B](f: B => Maybe2[A, B])(b: B): Maybe[T[A]]
```

which is not the dual for `Foldable`, as you all noted, because of it's possibility of returning nothing.
For this reason, I propose to rename this to `Unfoldable0`.

We can then introduce a proper `Unfoldable`, which will make the promise of always return something, like this:

```scala
def unfoldRight[A, B](f: B => (A, Maybe[B]))(b: B): T[A]
```

In this way, the names align as well (Foldable <-> Unfoldable, and the 0 in Unfoldable0 shows the possibility of an empty result).
If there's interest, I can fix this.

The other point to be made here is if it makes sense to use `Unfoldable` to implement `traverse`, `sequence`, etc.. since we already have most of them for free in TraversableSyntax.

base/shared/src/main/scala/scalaz/data/io.scala
```diff
@@ -0,0 +1,83 @@
...
+    )(implicit f: Foldable[F], cbf: CanBuildFrom[F[IO[E, A]], A, F[A]]): IO[E, Fiber[E, F[A]]] =
+      f.foldRight(as, IO.point[Fiber[E, mutable.Builder[A, F[A]]]](Fiber.point(cbf(as)))) {
+          case (a, as) => as.par(a.fork).map { case (as, a) => as.zipWith(a)(_ += _) }
+        }
+        .map(as => as.zipWith(Fiber.point(())) { case (as, _) => as.result })
+
+    final def raceAll[E, A, F[_]](t: F[IO[E, A]])(implicit f: Foldable[F]): IO[E, A] =
+      f.foldLeft[IO[E, A], IO[E, A]](t, IO.terminate(NothingRaced))(_ race _)
+
+    final def reduceAll[E, A, F[_]](a: IO[E, A], as: F[IO[E, A]])(f: (A, A) => A)(implicit fd: Foldable[F]): IO[E, A] =
```
ktonga: Don't we have `OneAnd` yet in z8?

base/shared/src/main/scala/scalaz/data/io.scala
```diff
@@ -0,0 +1,83 @@
...
+    acquire: IO[E, A]
+  )(release: (A, Maybe[E \/ B]) => IO[Nothing, Unit])(use: A => IO[E, B]): IO[E, B] = {
+    val release1: (A, Option[Either[E, B]]) => IO[Nothing, Unit] =
+      (a: A, o: Option[Either[E, B]]) => release(a, Maybe.fromOption(o.map(Disjunction.fromEither[E, B])))
+    IO.bracket0[E, A, B](acquire)(release1)(use)
+  }
+
+  implicit class IOObjOps(private val ioObj: IO.type) extends AnyVal {
+
+    final def absolve[E, A](v: IO[E, E \/ A]): IO[E, A] =
```
ktonga: Why is `absolve` in the companion instead of in the `IO` type? That's something I always wondered in zio too. It always breaks dot syntax for me. Did you have a reason @jdegoes ?

base/shared/src/main/scala/scalaz/data/io.scala
AL333Z: Not yet I guess

AL333Z:
Hi guys, 
I rebased bringing in latests updates.
I updated the PR description as well, reflecting what is still to be discussed.

base/shared/src/main/scala/scalaz/data/io.scala
jdegoes: We would need `ev` magic to do it into `IO`. I guess I'm fine with that. In the past I've stayed away.

base/shared/src/main/scala/scalaz/data/io.scala
jdegoes: It will definitely be inefficient. But we can deal with that later.

We need `Collection[F] extends Foldable[F] with Unfoldable[F]`, and then we can specialize on more common compositions.

base/shared/src/main/scala/scalaz/data/io.scala
```diff
@@ -0,0 +1,83 @@
...
+      fd.foldLeft(as, a) { (l, r) =>
+        l.par(r).map(f.tupled)
+      }
+
+    final def mergeAll[E, A, B, F[_]](in: F[IO[E, A]])(zero: B, f: (B, A) => B)(implicit fd: Foldable[F]): IO[E, B] =
+      fd.foldLeft[IO[E, A], IO[E, B]](in, IO.point[B](zero))((acc, a) => acc.par(a).map(f.tupled))
+  }
+
+  implicit class IOOps[E, A](private val io: IO[E, A]) extends AnyVal {
+    final def attempt: IO[Nothing, E \/ A] =
```
jdegoes: Not sure Scala will find this one, unfortunately.

base/shared/src/main/scala/scalaz/data/io.scala
AL333Z: I actually run a couple of benchmark with the version i'm working on, and.. it's way too slow.
Will push that anyway if that's ok with you,, and optimize later on.

base/shared/src/main/scala/scalaz/data/io.scala
AL333Z: Ouch.. just got bitten by this while writing tests..

```scala
test("bracket0") { () =>
  def release[A, E, B]: (A, Maybe[E \/ B]) => IO[Nothing, Unit] = (a, meb) => IO.point(())
  def use[A, E]: A => IO[E, A] = a => IO.point(a)

  val res: Int = unsafeRun(IO.point[Int](1).bracket0(release[Int, Nothing, Int])(use[Int, Nothing]))
  assertEqual(1, 1) // TODO
}
```

This is returning

> [error] type mismatch;
[error]  found   : (Int, scalaz.data.Maybe[Nothing \/ Int]) => scalaz.zio.IO[Nothing,Unit]
[error]     (which expands to)  (Int, scalaz.data.Maybe.Maybe[scalaz.data.Disjunction[Nothing,Int]]) => scalaz.zio.IO[Nothing,Unit]
[error]  required: (Int, Option[Either[?,?]]) => scalaz.zio.IO[Nothing,Unit]
[error]           val res: Int = unsafeRun(IO.point[Int](1).bracket0(release[Int, Nothing, Int])(use[Int, Nothing]))
    
Any ideas?

jdegoes: In order for Scala to use the implicit, there cannot exist a method of the same name on the target with the same erased type. So the options are to change the type, so Scala can find it, or to change the name.

AL333Z: I guess we should find new names then..
I wouldn't touch the types.
Any conventions here, or I'm free to propose new ones?
Thanks @jdegoes , I'm learning a lot.

AL333Z:
Hi guys,

I completed the port of all the methods impacted and I added tests as well.
Not sure if this is what you'd expect.

In order for Scala to use the extensions I choose to add a trailing `z` to all the names, so Scala can find them. I really don't like this, and i'm open to suggestions.

I'm still unsure if this is the right moment for having this merged, since zio apis are still evolving and I found myself updating the code quite a lot.

jdegoes:
@AL333Z I think if we depend on a non-snapshot version we should be safe.

Also I hope to argue to change to Option and Either in Sz8 (but disable unsafe methods and with invariant constructors), so hopefully that will ease the issue in the future.

AL333Z:
Thanks @jdegoes ,
Sorry for the late response, will do in a couple of days or so.
Should i depend on 0.2.7 which is marked as Pre-Release?

jdegoes:
I think 0.2.7 is fine. It just needs to be fixed to exactly that version.

AL333Z:
Hi @jdegoes , all done. 
Please note that tests are now reporting out a bunch of `UnhandledError`s.
@edmundnoble let me know if this is fine for you or if I can improve something.

AL333Z:
I rebased once again.
Looks like CI is failing due to `TestMain` not exiting after tests run.
Maybe something related to threads in WAIT status not letting the jvm to get down?

jdegoes:
@AL333Z You can force shutdown of the test `RTS` if you want, which should solve the problem.

Thanks for all your hard work on this!

AL333Z:
Hi @jdegoes,
all done :smiley: 

I'd like to understand more about why the jvm was kept alive. Anything related to the RTS internals, I guess. Any pointers?
Thanks a lot.

jdegoes:
@AL333Z The thread pool does not shut down after inactivity... this could be changed, of course (and maybe should be?).

Issue #1385, opened 2017-05-09
Slow Scaladocs


fosskers:
Hi there. Every page load of the Scaladocs freshly downloads a ~9mb javascript file:
![slow](https://cloud.githubusercontent.com/assets/229679/25866376/270f4232-34ab-11e7-904d-2dcdde23e1d5.png)

which severely slows down interaction with the docs. I haven't noticed this issue with other scala projects that also produce 2.12-style docs.

Thoughts? Thanks.

xuwei-k:
Thanks for the report. Are you using sonatyp? (e.g. https://oss.sonatype.org/service/local/repositories/releases/archive/org/scalaz/scalaz_2.12/7.2.12/scalaz_2.12-7.2.12-javadoc.jar/!/scalaz/index.html )

It's just as you say, sonatype is so slow.

Now, scalaz unidoc indexed on <http://javadoc.io> ( I just added `pom.xml` for unidoc )

Maybe <http://javadoc.io> has some nice features(Optimized HTTP caching, Accelerated with CDN, gzip support)

Could you try to following url?
https://static.javadoc.io/org.scalaz/scalaz_2.12/7.2.12/scalaz/index.html


![javadoc.io](https://cloud.githubusercontent.com/assets/389787/26062875/33f061d6-39c7-11e7-9240-8c956fc5ef70.png)

fosskers:
Nice and fast! Thanks for the fix.

PR #1919, opened 2018-07-17
Add Scaladoc link to microsite

NeQuissimus:
This adds a link in the top right corner that points to the Scaladoc


xuwei-k:
use javadoc.io instead of sonatype? https://github.com/scalaz/scalaz/issues/1385

NeQuissimus:
@xuwei-k I had used Sonatype because that's where the SNAPSHOT docs are but I suppose that's not really relevant :) Changed it

PR #1927, opened 2018-07-24
Typeclass diagram

NeQuissimus:
Adds an interactive diagram of the type class hierarchy.

Screenshot: https://screenshots.firefox.com/1P3eySReEsEb3bFM/localhost

The nodes support drag/drop.
I think as I build this out, I will split the tiny graphs (<5 nodes with separate root) off the graph with `InvariantFunctor`.
It will make things much cleaner.

Edit: Nodes in the diagram can be made clickable (see `Equal` as an example in the data set)


fommil:
is there no way we can put the docs in a separate branch or something? It feels super hacky to be adding all these javascript files to the scalaz source code itself.

NeQuissimus:
I can't think of a good way to do that. You can have a separate sbt project on a branch and depend on an artifact but that's not really the idea, I think...
FWIW, I don't plan on adding any more JS or CSS. I just needed something to draw graphs.

jdegoes:
@NeQuissimus When do we get it for 8? 😄

NeQuissimus:
@jdegoes When I have a bit of time or somebody ports my improvements from 7.x to 8.x :) There are other things I want to port.

PR #1929, opened 2018-07-31
Fix build after scalaz-zio changes.

paulp:
Not sure of the wisdom of depending on a snapshot.


ktonga:
https://github.com/scalaz/scalaz/pull/1874#issuecomment-397914432 :slightly_smiling_face: 

We should change it as soon as we start producing non-snapshot versions for zio scalaz/scalaz-zio#145

PR #1931, opened 2018-07-31
Make scalaz-zio a source dependency.

paulp:
Figuring out how to do things in a cross project sends my blood pressure through the roof, but this seems to work. Mostly I had to figure out that I needed to re-make all the same changes I made to the scalaz-zio build.


fommil:
source deps work reasonably well, especially if there are no transitives or compiler plugins. I've been experimenting with them in https://gitlab.com/fommil/upgrade-typelevel-to-scalaz
which has installation instructions along the lines of

```scala
    val scalaz = ProjectRef(
      uri("https://gitlab.com/fommil/upgrade-typelevel-to-scalaz.git#28a9bdf"),
      "upgrade-typelevel-to-scalaz"
    )
    val myproject = ... .dependsOn(scalaz)
```

sbt could do a better job, of course, but that's what Fury is all about.

build.sbt
```diff
@@ -31,11 +33,15 @@ lazy val root = project
   .aggregate(baseJVM, baseJS, metaJVM, metaJS, stdJVM, stdJS, microsite, benchmarks)
   .enablePlugins(ScalaJSPlugin)
 
-resolvers in ThisBuild += "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots"
+lazy val zioJVM = ProjectRef(uri("git://github.com/scalaz/scalaz-zio.git#master"), "coreJVM")
```
ktonga: Isn't pointing to `master` equivalent to depending on the snapshot?
Should we change it to a commit SHA until we start tagging?

paulp: Not quite the same, but close enough. It was a placeholder since I don't have a good project intuition. I can change it to the current sha if that is desired.

paulp:
I am unsure how a source dependency affects downstream projects.

build.sbt
ktonga: I think current SHA would do it by now. Mostly to avoid what happened that you had to fix. We can bump it from time to time.
The other option is to point to master until it breaks and only then we fix the ref to the last working commit while we fix it.

paulp: In a perfect world I think it would track the repository head until the tests fail here, at which point it fixes on the last working commit until new ones are passing again. The git/github/travis machinery is probably way too coarse to allow such a flow. I'll inject the current sha.

NeQuissimus:
Somehow, `First-time contributor` is my favourite part of this PR :rofl:

paulp:
@NeQuissimus everyone has to start somewhere 😀

paulp:
@ktonga I'm not seeing where it says "first-time contributor" so I'm only able to discuss it on the most hypothetical of terms.

paulp:
[Attempt to push cross-source-project logic upstream.](https://github.com/portable-scala/sbt-crossproject/issues/102)

hrhino:
So (since it came up on gitter) I cloned this, set the version to something specific, and ran `publishLocal`, and I got this as my `scalaz-base` POM:
```xml
<?xml version='1.0' encoding='UTF-8'?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0">
    <modelVersion>4.0.0</modelVersion>
    <groupId>org.scalaz</groupId>
    <artifactId>scalaz-base_2.12</artifactId>
    <packaging>jar</packaging>
    <description>scalaz-base</description>
    <version>8.0.0-sourcedep-1-SNAPSHOT</version>
    <name>scalaz-base</name>
    <organization>
        <name>org.scalaz</name>
    </organization>
    <scm>
        <url>https://github.com/hrhino/scalaz</url>
        <connection>scm:git:https://github.com/hrhino/scalaz.git</connection>
        <developerConnection>scm:git:git@github.com:hrhino/scalaz.git</developerConnection>
    </scm>
    <dependencies>
        <dependency>
            <groupId>org.scala-lang</groupId>
            <artifactId>scala-library</artifactId>
            <version>2.12.6</version>
        </dependency>
        <dependency>
            <groupId>org.scalaz</groupId>
            <artifactId>scalaz-meta_2.12</artifactId>
            <version>8.0.0-sourcedep-1-SNAPSHOT</version>
        </dependency>
        <dependency>
            <groupId>org.scalaz</groupId>
            <artifactId>scalaz-zio_2.12</artifactId>
            <version>8.0.0-sourcedep-1-SNAPSHOT</version>
        </dependency>
        <dependency>
            <groupId>com.github.ghik</groupId>
            <artifactId>silencer-lib_2.12</artifactId>
            <version>1.0</version>
            <scope>provided</scope>
        </dependency>
        <dependency>
            <groupId>org.scalacheck</groupId>
            <artifactId>scalacheck_2.12</artifactId>
            <version>1.14.0</version>
            <scope>test</scope>
        </dependency>
    </dependencies>
</project>
```

which references a possibly-not-yet-published version of `scalaz-zio`. 

I don't know what this means for projects downstream of this one; presumably if they too use source dependencies sbt will figure it out, but I don't think we can get everyone on board with that.

@fommil / @xuwei-k / @paulp do you have any opinion on this? Sam mentioned "some sbt magic to reexport the jar"; as I understand it that would more-or-less mean that this project and `zio` release in sync.

paulp:
@hrhino Mmm. This gets into weeds where I'd be speculating unacceptably. If someone else has a good idea of the right thing to do, I can try to implement it.

fommil:
basically if you have a source dependency you can't publish binaries.

Ask yourself: are we really ready for publishing binaries? I don't think so. Anybody using scalaz8 should be smart enough to figure this out for themselves and either add a source dep on scalaz8 or a source dep on zio or both.

If you want to workaround this, you could create a project that has a source dependency on zio and then builds a one jar or something. But I don't think it's worth the effort.

edmundnoble:
I think we missed the opportunity for this. Should this be closed?

PR #1932, opened 2018-08-01
Add more decidable/divisible instances

azanin:
This is related to #1885 by @fommil .

### DIFF
--- a/core/src/main/scala/scalaz/Kleisli.scala
+++ b/core/src/main/scala/scalaz/Kleisli.scala
@@ -97,7 +97,21 @@ final case class Kleisli[M[_], A, B](run: A => M[B]) { self =>
 // Prioritized Implicits for type class instances
 //
 
-sealed abstract class KleisliInstances13 {
+sealed abstract class KleisliInstances15 {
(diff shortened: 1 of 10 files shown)


fommil:
great! Is there any precedent for these instances from Haskell? I'm concerned that some of them may be lawless. In particular it would be good if we could add tests for `Divisible[WriterT]` and perhaps introduce new laws that check the consistency of derived combinators for `Decidable`. e.g. `alt2(alt2, alt2)` should equal `alt4`

azanin:
I have added only instances present in Haskell. (http://hackage.haskell.org/package/contravariant-1.5/docs/Data-Functor-Contravariant-Divisible.html)

You can find unit test for for `Divisible[WriterT]` here: https://github.com/scalaz/scalaz/pull/1932/commits/e07d502f3879793653930d0a49c40277ea1cb0c1 . (Let me know it it is enough).

Not sure to have understood `alt4` as derived combinator of `Decidable`, I thought it was for `Alt`. Am I missing something?

fommil:
sorry I meant `choose`, not `alt`. Awesome, if it's in Haskell we can assume it's been given more thought already by Ed Kmett than I can give it...

PR #1936, opened 2018-08-04
Laws 2

edmundnoble:
This depends on #1934. 
It adds laws for comonads, cobind, and traverse.

It also adds all laws for homomorphisms.


edmundnoble:
@NeQuissimus Just so you know in advance, I'll add a microsite page to explain the laws as soon as I've integrated them with testz, which is the next step I want to take.

laws/shared/src/main/scala/scalaz/laws/HomomorphismLaws.scala
```diff
@@ -0,0 +1,191 @@
...
+package laws
+
+import scala.{ Nothing, Unit }
+
+import data._, tc._, Scalaz._
+
+object HomomorphismLaws {
+  // To be a monoid homomorphism, a function must satisfy both `monoidIdentity`
+  // and `semigroupAppend`.
+  def monoidIdentity[A, B, T](f: A => B)(assert: (B, B) => T)(implicit A: Monoid[A], B: Monoid[B]): T =
```
Jacoby6000: If this is testing homomorphisms, then shouldn't it be testing in both directions?  This only verifies half of the homomorphism, unless I'm misunderstanding something.

(Same question applies to other assertions in this file)

edmundnoble: I don't understand; homomorphisms go in one direction.

edmundnoble: Though that reminds me, it would be nice to have isomorphism laws too.

Jacoby6000: Oh yeah my bad. I got mixed up. Pretend that I was telling you to write `Iso` laws.

project/ScalazBuild.scala
```diff
@@ -22,7 +22,6 @@ object Scalaz {
     "-opt-warnings",
     "-Xlint:_,-type-parameter-shadow",
     "-Xsource:2.13",
-    "-Ywarn-dead-code",
```
alexknvl: :heart:

PR #1938, opened 2018-08-06
Align with latest updates in zio

AL333Z:
Once again, this will make it compile again :)

Related #1931


jdegoes:
@AL333Z Thank you! 🙏

jdegoes:
Also we need to get off snapshot builds. 😆

PR #1939, opened 2018-08-07
IList tests

edmundnoble:
The first wave of tests using the laws and testz!

CI will not actually fail if tests fail right now, unfortunately, until https://github.com/scalaz/testz/pull/13 is merged and released into testz.

I also had to add a few instances to `IList`, and had a good opportunity to test the laws; replacing `ap` with the zipping variation causes test failures, which is nice.


base/shared/src/main/scala/scalaz/data/ilist.scala
```diff
@@ -97,12 +154,6 @@ object IListModule {
     def uncons: Maybe2[A, IList[A]] =
       IList.uncons(self)
 
-    def foldLeft[B](z: => B)(f: (B, A) => B): B =
-      IList.foldLeft(f, z, self)
-
-    def foldRight[B](z: => B)(f: (A, B) => B): B =
-      self.foldLeft(z)((b, a) => f(a, b))
```
edmundnoble: This actually was incorrect, you need to reverse first before using `foldLeft`.

laws/shared/src/main/scala/scalaz/laws/HomomorphismLaws.scala
```diff
@@ -0,0 +1,191 @@
+package scalaz
+package laws
+
+import scala.{ Nothing, Unit }
+
+import data._, tc._, Scalaz._
+
+object HomomorphismLaws {
```
hrhino: why this and not `*Laws`s for the various implicated typeclasses?

laws/shared/src/main/scala/scalaz/laws/TypeclassLaws.scala
```diff
@@ -0,0 +1,217 @@
...
+  def antisymmetry[A, T](fst: A, snd: A)(assert: Boolean => T)(implicit A: Ord[A]): T = {
+    val mustBeEqual =
+      ((A.<=(fst, snd)) && (A.<=(snd, fst)))
+    val satisfiesAntisymmetry =
+      mustBeEqual == A.equal(fst, snd)
+    assert(satisfiesAntisymmetry)
+  }
+
+  @inline
+  def connex[A, T](fst: A, snd: A)(assert: Boolean => T)(implicit A: Ord[A]): T = {
```
hrhino: it feels like there should be a better `assert` type than this, but I don't think it's in testz yet, is it?

laws/shared/src/main/scala/scalaz/laws/HomomorphismLaws.scala
edmundnoble: I believe this is from #1936. @alexknvl has suggested putting the laws into the type classes themselves which seems like an okay idea to me. I'll try that first.

edmundnoble:
If this isn't merged soon, I'll have to add more tests :wink:

base/shared/src/main/scala/scalaz/data/ilist.scala
```diff
@@ -72,6 +73,58 @@ object IListModule {
...
 
+  implicit final def ilistMonad: Monad[IList] =
+    instanceOf(new MonadClass[IList] {
+      def pure[A](a: A): IList[A] = IList.cons(a, IList.empty)
+      def ap[A, B](fa: IList[A])(ff: IList[A => B]): IList[B] =
+        flatten(map(fa)(a => map(ff)(f => f(a))))
+      def flatMap[A, B](ma: IList[A])(f: A => IList[B]): IList[B] =
+        flatten(map(ma)(f))
+      def flatten[A](nested: IList[IList[A]]): IList[A] =
+        nested.foldRight(IList.empty[A])((out, as) => as.append(out))
```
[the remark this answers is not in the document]
edmundnoble: I'm not sure whether to rename `prepend` or fix `append`. I'll address this in a follow-up if that's alright.

base/shared/src/main/scala/scalaz/data/ilist.scala
```diff
@@ -72,6 +73,58 @@ object IListModule {
...
+      def foldRight[A, B](fa: IList[A], z: => B)(f: (A, => B) => B): B =
+        fa.reverse.foldLeft(z)((b, a) => f(a, b))
+
+      def toList[A](fa: IList[A]): scala.List[A] = {
+        import scala.{ ::, List, Nil }
+        foldRight[A, List[A]](fa, Nil)(new ::(_, _))
+      }
+
+      def sequence[F[_], A](ta: IList[F[A]])(implicit F: Applicative[F]): F[IList[A]] =
+        traverse(ta)(fa => fa)
```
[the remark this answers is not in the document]
edmundnoble: For sure, I was just feeling lazy. Will address in a follow-up.

Issue #1882, opened 2018-06-18
MonadPlus reform


fommil:
https://wiki.haskell.org/MonadPlus_reform_proposal

I'm not sure if we can fix this in time for scalaz 7.3, but we should definitely split up Alternative and Plus in scalaz 8.

fommil:
(on the co side we should split up decidable into decide + the identity version)

fommil:
(this is pretty tricky...)

PR #1941, opened 2018-08-07
[WIP] MonadPlus hierarchy

azanin:
This is just a draft proposal for #1882  opened by @fommil .
This is totally inspired from https://github.com/purescript/purescript-control .

Laws that justify each typeclass are reported here: https://pursuit.purescript.org/packages/purescript-control/4.1.0/docs/Control.MonadPlus

I add other references on this topic:

- https://stackoverflow.com/questions/10167879/distinction-between-typeclasses-monadplus-alternative-and-monoid

- https://stackoverflow.com/questions/13122489/what-s-an-example-of-a-monad-which-is-an-alternative-but-not-a-monadplus

- https://en.wikibooks.org/wiki/Haskell/Alternative_and_MonadPlus

I'd like to have some feedback on this topic and if it make sense having such hierarchy organization

### DIFF
--- a/base/shared/src/main/scala/scalaz/tc/alt.scala
+++ b/base/shared/src/main/scala/scalaz/tc/alt.scala
@@ -1,0 +1,5 @@
+package scalaz.tc
+
+trait AltClass[F[_]] extends FunctorClass[F] {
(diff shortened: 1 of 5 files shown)


fommil:
@tonymorris has thoughts on this...

Issue #1940, opened 2018-08-07
Comprehensive instances for ZIO


jdegoes:
I'm sure I'm missing some:

 - [X] `Monad` for `IO`
 - [X] `Bifunctor` for `IO`
 - [ ] `MonadPlus`, `Alternative` for `IO`
 - [ ] `Monoid` for `IO`
 - [ ] `Applicative` for `Fiber`
 - [ ] `Monoid` for `Fiber`
 - [ ] `Applicative` for `Retry`
 - [ ] `Monoid` for `Retry`

edmundnoble:
We have the `Applicative` for `Fiber`, see https://github.com/scalaz/scalaz/blob/series/8.0.x/base/shared/src/main/scala/scalaz/tc/applicative.scala#L20

jdegoes:
@azanin `Schedule` has shipped and therefore it's now possible to write instances for it.

azanin:
@jdegoes I was opening PR for Schedule instances but I just realized that https://github.com/scalaz/scalaz/pull/1998 has been opened. 

I will look for something else, thanks

PR #1945, opened 2018-08-08
[WIP] Add IO Monoid instance

azanin:
This relates to #1940 .
I've added just ioMonoid instance, since:

-  ~~`Retry` is still wip (https://github.com/scalaz/scalaz-zio/pull/141)~~

- `Alternative` and `MonadPlus` have not been defined yet ( see https://github.com/scalaz/scalaz/issues/1882 and https://github.com/scalaz/scalaz/pull/1941 )

### DIFF
--- a/base/shared/src/main/scala/scalaz/tc/monoid.scala
+++ b/base/shared/src/main/scala/scalaz/tc/monoid.scala
@@ -38,9 +38,11 @@ object MonoidClass {
       }
     })
 
-  implicit def ioMonoid[E, A](implicit A: Monoid[A]): Monoid[IO[E, A]] =
+  implicit def ioMonoid[E, A](implicit SE: Semigroup[E], A: Monoid[A]): Monoid[IO[E, A]] =
     instanceOf(new MonoidClass[IO[E, A]] {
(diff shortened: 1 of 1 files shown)


edmundnoble:
I don't think this instance is necessarily a good idea, see https://failex.blogspot.ca/2015/10/the-uninteresting-monoids-of-certain.html. I would like the more general version if possible, unless this is the most interesting monoid for `IO` (and I suspect it isn't, it would probably have to do with choice/failure).

jbgi:
Gabriel has provided some motivations for the haskell IO instance in those posts:
 - http://www.haskellforall.com/2014/07/equational-reasoning-at-scale.html
 - http://www.haskellforall.com/2018/02/the-wizard-monoid.html

But as @S11001001 points out, great care is required for default instances (because if not the most useful, people will be required to wrap/unwrap or it will necessitate a breaking change).

azanin:
Yep. I've already read these post and I've privately spoken with @edmundnoble about this topic. I get the point of what @emilypi is talking about.

I think we can wait for @edmundnoble feedback  and then take a final decision on the subject.

Thanks guys :)

emilypi:
@azanin are you still working on this?

azanin:
@emilypi not anymore. I'm going to close this (there are just few lines of code). Thanks

PR #1953, opened 2018-08-10
Add more IList tests

edmundnoble:
Should test every method on `IList`.


base/shared/src/main/scala/scalaz/data/ilist.scala
```diff
@@ -160,7 +163,7 @@ object IListModule {
       IList.uncons(self)
 
     def append(that: IList[A]): IList[A] =
-      IList.reverse(that).foldLeft(self)((b, a) => IList.cons(a, b))
+      self.foldRight(that)(_ :: _)
 
     def :::(that: IList[A]): IList[A] =
       self.append(that)
```
alexknvl: `that ::: self = that ++ self != self.append(that) = self ++ that`
Same thing with `++`. I guess both would eventually get caught by the tests :)

base/shared/src/main/scala/scalaz/data/maybe2.scala
```diff
@@ -42,31 +42,39 @@ final class Just2Extractor[A, B] private[data] (private val value: Option2[A, B]
   def _2: B                     = value._2
 }
 
+object Maybe2Module {
+  implicit def isCovariant_1[B]: IsCovariant[Maybe2[?, B]] = Maybe2.isCovariant_1
+  implicit def isCovariant_2[A]: IsCovariant[Maybe2[A, ?]] = Maybe2.isCovariant_2
+  implicit def eq[A, B](implicit A: Eq[A], B: Eq[B]): Eq[Maybe2[A, B]] =
+    Maybe2.eq[A, B]
+  implicit def bifunctor: Bifunctor[Maybe2] = Maybe2.bifunctor
```
alexknvl: `Traversable` etc?

base/shared/src/main/scala/scalaz/tc/foldable.scala
```diff
@@ -30,6 +32,9 @@ object FoldableClass {
   trait DeriveFoldMap[F[_]] extends FoldableClass[F] with Alt[DeriveFoldMap[F]] {
     final override def foldMap[A, B](fa: F[A])(f: A => B)(implicit B: Monoid[B]) =
       foldRight(fa, B.mempty)((a, b) => B.mappend(f(a), b))
+
+    final override def msuml[A](fa: F[A])(implicit A: Monoid[A]) =
+      foldLeft(fa, A.mempty)((a, s) => A.mappend(a, s))
```
alexknvl: Maybe `A.mappend(_, _)`? Btw if `s` here means `sum`, it should be on the left. Semantically it looks correct.

meta/shared/src/main/scala/scalaz/meta/Ops.scala
```diff
@@ -39,6 +39,10 @@ class Ops(val c: blackbox.Context) {
   def n_0[T: c.WeakTypeTag]: Tree            = q"$ev.${typeName[T]}($lhs)"
   def ni_0[T: c.WeakTypeTag](ev: Tree): Tree = q"$ev.${typeName[T]}($lhs)"
 
+  /* def method(implicit g: B): R */
```
alexknvl: `def method(lhs: T)(implicit g: B): R`?

alexknvl: I really want to get rid of all these macros and make scalaz-plugin just inline syntax all the time.

tests/src/main/scala/scalaz/tests/IListTests.scala
```diff
@@ -22,93 +22,272 @@ final class IListTests {
...
-  def tests[T](harness: Harness[T]): T = {
+  def tests[T](harness: Harness[T], sequence: (T, T) => T): T = {
     import harness._
-    section("laws")(
-      section("eq laws")(
-        test("reflexivity") { () =>
-          lists.foldMap(
-            EqLaws.reflexivity(_)(assert)
-          )
+    sequence(
```
alexknvl: Why is this necessary?

base/shared/src/main/scala/scalaz/data/ilist.scala
edmundnoble: `self.append(that)` is supposed to be `self ++ that`. Don't like the name?

base/shared/src/main/scala/scalaz/data/maybe2.scala
edmundnoble: I just exposed the instances that were not being exposed properly; adding new ones is for later.

tests/src/main/scala/scalaz/tests/IListTests.scala
edmundnoble: Two sections at the same level with no parent section.

base/shared/src/main/scala/scalaz/data/ilist.scala
alexknvl: yes but `that ::: self` is not `self.append(that)`.

alexknvl: Just add tests for `:::` and `++` and you'll see what I mean ;)

base/shared/src/main/scala/scalaz/tc/foldable.scala
edmundnoble: Why on the left? I don't follow.

base/shared/src/main/scala/scalaz/data/ilist.scala
edmundnoble: Poop, fixed lol.

PR #1957, opened 2018-08-13
Fixed Eq[Double].

alexknvl:


NeQuissimus:
We don't want `-0.0d` and `0.0d` to be equal?

alexknvl:
They are clearly not equal, since they have different observable behavior:
```scala
def check(x: Double): Boolean = (1/x) > 0
println(check(+0.0), check(-0.0))
// (true,false)
```
They might be equivalent in a particular domain if you don't care about the distinction, but then you can just use a newtype that does not allow you to observe the difference.

base/shared/src/main/scala/scalaz/data/ilist.scala
```diff
@@ -210,6 +210,9 @@ object IListModule {
 
     def !!(n: Int): Maybe[A] = self.index(n)
 
+    def cross[B](b: IList[B]): IList[(A, B)] =
+      self.flatMap(a => b.map(b => (a, b)))
```
edmundnoble: Wait, isn't this just `.tuple2`?

alexknvl: Yes, I think so, will fix.

alexknvl: Nvm, there is no tuple2 yet.

PR #1960, opened 2018-08-14
Tests for ACatenable1

edmundnoble:
Also added a `Category[Function1]` (untested) and found a `Semicategory[Map]` (tested), added the `|+|` alias for `mappend`, the `<<<` and `>>>` aliases for `compose` and `andThen` respectively.

P.S.: Removed unnecessary indirection in `Const` instances.

P.P.S: Started running tests in parallel, just at the suite-level. Gives me almost a 2x speedup on my machine, 80 ms to 40 ms 😝

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/acatenable1.scala
+++ b/base/shared/src/main/scala/scalaz/data/acatenable1.scala
@@ -3,7 +3,7 @@ package data
 
 import scala.annotation.tailrec
 
-import tc.{instanceOf, Semicategory, SemicategoryClass}
+import tc.{ instanceOf, Semicategory, SemicategoryClass }
 
 /**
  * Non-empty type-aligned sequence represented as a (non-balanced) binary tree,
@@ -77,7 +77,7 @@ object ACatenable1 {
   // (depending on user code) and never asymptotically slower.
   implicit def acatenable1FreeSemicategory[=>:[_, _]]: Semicategory[ACatenable1[=>:, ?, ?]] =
     instanceOf(new SemicategoryClass[ACatenable1[=>:, ?, ?]] {
-      def compose[A, B, C](f: ACatenable1[=>:,B,C], g: ACatenable1[=>:,A,B]): ACatenable1[=>:,A,C] =
+      def compose[A, B, C](f: ACatenable1[=>:, B, C], g: ACatenable1[=>:, A, B]): ACatenable1[=>:, A, C] =
         f.compose(g)
     })
 }
--- a/base/shared/src/main/scala/scalaz/data/acatenable1.scala
+++ b/base/shared/src/main/scala/scalaz/data/acatenable1.scala
@@ -3,7 +3,7 @@ package data
 
 import scala.annotation.tailrec
 
-import tc.Semicategory
+import tc.{instanceOf, Semicategory, SemicategoryClass}
 
 /**
  * Non-empty type-aligned sequence represented as a (non-balanced) binary tree,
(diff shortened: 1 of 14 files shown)


tests/src/main/scala/scalaz/tests/ACatenable1Tests.scala
```diff
@@ -0,0 +1,147 @@
...
+    import harness._
+
+    sequence(
+      section("laws")(
+        section("lawful semicategory")(
+          test("associativity") { () =>
+            SemicategoryLaws.composeAssoc(fst, snd, thd)(assertEqualNonEmptyMaps)
+          },
+        ),
+        section("free semicategory")(
```
edmundnoble: This is the idea behind testing free structures: we test that `foldMap` is a homomorphism, and that `lift` is a homomorphism, when observed by `foldMap`.

base/shared/src/main/scala/scalaz/tc/semicategory.scala
```diff
@@ -1,14 +1,31 @@
...
+    def andThen[D](f: C =>: D)(implicit ev: Semicategory[=>:]): B =>: D = macro meta.Ops.ia_1
+    type compose
+    type andThen
+    def <<<[A](f: A =>: B)(implicit ev: Semicategory[=>:]): A =>: C = macro meta.Ops.nia_1[compose]
+    def >>>[D](f: C =>: D)(implicit ev: Semicategory[=>:]): B =>: D = macro meta.Ops.nia_1[andThen]
   }
 }
+
+object SemicategoryClass {
+  implicit def mapSemicategory: Semicategory[SMap] =
```
alexknvl: Once we get our own `Map`, we should move this to `outlaws`.

tests/src/main/scala/scalaz/tests/ACatenable1Tests.scala
alexknvl: Strictly speaking you should also ensure that all observations are done through `foldMap`, right? That ensures that there is no extra fluff.

base/shared/src/main/scala/scalaz/data/acatenable1.scala
```diff
@@ -15,17 +15,20 @@ sealed abstract class ACatenable1[=>:[_, _], A, B] {
...
     this compose that
 
   def >>>[C](that: ACatenable1[=>:, B, C]): ACatenable1[=>:, A, C] =
-    that compose this
+    this andThen that
 
-  def :+[C](f: B =>: C): ACatenable1[=>:, A, C] =
-    Chain(this, Lift(f))
+  def :+[Z](f: Z =>: A): ACatenable1[=>:, Z, B] =
+    compose(Lift(f))
```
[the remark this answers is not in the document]
alexknvl: ah, tests are passed, so it must be correct.

tests/src/main/scala/scalaz/tests/ACatenable1Tests.scala
edmundnoble: Right, that's what I'm doing, isn't it?

base/shared/src/main/scala/scalaz/data/acatenable1.scala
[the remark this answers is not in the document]
edmundnoble: I know, right-assoc is fucked.

base/shared/src/main/scala/scalaz/data/acatenable1.scala
```diff
@@ -68,4 +71,13 @@ object ACatenable1 {
 
   def lift[F[_, _], A, B](f: F[A, B]): ACatenable1[F, A, B] =
     Lift(f)
+
+  // ACatenable1 is *the* free semicategory in lazy languages like Haskell.
```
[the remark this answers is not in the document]
edmundnoble: Oh in Haskell, it's the only one, because of infinitary reassociation.

base/shared/src/main/scala/scalaz/tc/semicategory.scala
edmundnoble: Is this not lawful?

PR #1965, opened 2018-08-20
Hundred columns in scalafmt

edmundnoble:
120 just seems like too many to me; if we're going to rely on it at all, we should at least get somewhat readable lines.


NeQuissimus:
:rofl:

NeQuissimus:
Punchcard terrorists! :rofl:

alexknvl:
We'll need to coordinate better next time if we want to actually get this in.

PR #1966, opened 2018-08-20
Add Unfoldable1, update Unfoldable

AL333Z:
This PR will:
- add Unfoldable1
- update Unfoldable, removing the annoying `Maybe` in the output of `unfoldRight`

The representation is now just like the one in [Purescript](https://pursuit.purescript.org/packages/purescript-unfoldable/3.0.0/docs/Data.Unfoldable) and #1626 (with a difference in the encoding).

I also updated the microsite, even if I think a resonable example may arise with `NonEmptyList`, which is still missing in Scalaz 8.

### DIFF
--- a/microsite/src/main/tut/typeclass/Unfoldable1.md
+++ b/microsite/src/main/tut/typeclass/Unfoldable1.md
@@ -1,0 +1,26 @@
+---
+layout: docs
+title:  "Unfoldable1"
+---
+
+# Unfoldable1 [![GitHub](../img/github.png)](https://github.com/scalaz/scalaz/blob/series/8.0.x/base/shared/src/main/scala/scalaz/tc/unfoldable1.scala)
+
+* `Unfoldable1` identifies non-empty data structures which can be unfolded*, and it's a more general form than `Unfoldable`.
+Given a seed value `b` of type `B`, `unfoldRight1(f)(b)` let's you define how to generate (*unfold*) a non-empty data structure of type `F[A]`.
+The generating function `f` always return a value, and then optionally a value to continue unfolding from.
+
+**Typical imports**
+
+```tut:silent
+import scalaz.tc._
+import scalaz.data._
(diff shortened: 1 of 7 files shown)


microsite/src/main/tut/typeclass/Unfoldable1.md
```diff
@@ -0,0 +1,22 @@
+---
+layout: docs
+title:  "Unfoldable1"
+---
+
+# Unfoldable1 [![GitHub](../img/github.png)](https://github.com/scalaz/scalaz/blob/series/8.0.x/base/shared/src/main/scala/scalaz/tc/unfoldable1.scala)
+
+* `Unfoldable1` identifies non-empty data structures which can be unfolded.*
```
edmundnoble: I think this is *very* unclear and can be improved on, but it's fine to me to leave it for this PR.

emilypi: Something to the effect that `Unfoldable1` is analogous to `Foldable1`, which corresponds to the anamorphisms of coalgebraic structures would be nice.

microsite/src/main/tut/typeclass/Unfoldable1.md
```diff
@@ -0,0 +1,22 @@
...
+
+```tut:silent
+import scalaz.tc._
+import scalaz.data._
+import scalaz.Scalaz._
+```
+
+# Usage
+
+TBD A sample with NonEmptyList
```
emilypi: If you expand this, I'll approve

AL333Z: I pushed what I'd in mind, and tried a better wording as well.

AL333Z: /cc @emilypi :)

emilypi:
thanks @AL333Z ! Great job :)

PR #1969, opened 2018-08-24
Lattice Hierarchy

emilypi:
Hi y'all I have some lattices for your pies

Discussion point: do we care about bounded variants of Join, Meet, and Lattice? If so, easily done. I've also taken the liberty of bumping to the jvm option for using concurrent mark-sweep to G1GC, as it has been deprecated since J9


base/shared/src/main/scala/scalaz/tc/band.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  /**
+   * For `n = 0`, `value`
+   * For `n = 1`, `append(value, value)`
+   * For `n = 2`, `append(append(value, value), value)`
+   *
+   * The default definition uses peasant multiplication, exploiting associativity to only
+   * require `O(log n)` uses of [[mappend]]
+   */
+  def multiply1(value: A, n: Int): A = {
```
TomasMikula: Doesn't this always return `value`, due to idempotency?

base/shared/src/main/scala/scalaz/tc/lattice.scala
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Lattices are an algebraic structures consisting of a set L together with two operations
+ * "join" and "meet" that compute a unique supremum (least upper bound) and a unique infimum
+ * greatest lower bound) for every pair of finite, non-empty subsets of L
+ *
+ * See [[JoinSemiLatticeClass]] and [[MeetSemiLatticeClass]] for laws.
+ *
+ * @tparam A
+ */
+trait LatticeClass[A] extends JoinSemiLatticeClass[A] with MeetSemiLatticeClass[A]
```
TomasMikula: What should `mappend` do?

base/shared/src/main/scala/scalaz/tc/joinsemilattice.scala
```diff
@@ -0,0 +1,59 @@
+package scalaz.tc
+
+import Predef._
+import scalaz.data.{ Const, Identity }
+
+/**
+ * In abstract algebra (and order theory), a join-semilattice is a partially ordered set S together with
+ * with an operation "join" which computes the supremum (least upper bound) for every pair of finite,
+ * nonempty subsets of S.
```
TomasMikula: Either "for every pair of elements", or "for a finite, nonempty subset", but not "for every pair of finite, non-empty subsets" 😃

emilypi: Good catches all around. I'll formalize what I mean

base/shared/src/main/scala/scalaz/tc/lattice.scala
emilypi: Good point!

base/shared/src/main/scala/scalaz/tc/band.scala
emilypi: Yes. But do we want to exclusively compute the idempotent behavior or do we just want say "hey, this is an idempotent semigroup" and have it be a marker trait for the extension?

base/shared/src/main/scala/scalaz/tc/lattice.scala
TomasMikula: 👍

emilypi:
@TomasMikula I've addressed your comments + my comments, and went ahead with adding the bounded lattice heirarchy as well

base/shared/src/main/scala/scalaz/tc/band.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import scala.Int
+
+/**
+ * Band is a semigroup with idempotency.
+ *
+ */
+trait BandClass[A] extends SemigroupClass[A] {
+  override def exponent(a: A, i: Int): A = {
+    val _ = i // allocates, unfortunately - if we can remove this, that would be great
```
edmundnoble: This doesn't allocate anything on the heap, just on the stack. Really not a problem.

emilypi:
To recap what we've discussed in Gitter, there will be newtypes over `A` so that `A @@ Meet` will form a semilattice, and `A @@ Join` will form a semilattice, and the inheritance relationship between Lattice and JoinSemilattice + Meetsemilattices for normal and bounded variants will be left out.

base/shared/src/main/scala/scalaz/tc/semigroup.scala
```diff
@@ -1,11 +1,26 @@
...
+   * Compute the exponent of an element. a^n = exponent(a, n) = a * ... * a (n-times)
+   * If an algebra is idempotent, every element is an identity for itself modulo
+   * exponents, i.e. x*x = x and x^n = x - see [[Band]]
+   */
+  def exponent(a: A, i: Int): A = {
+    @tailrec
+    def go(a0: A, j: Int): A =
+      if (j == 0) a0 else go(mappend(a, a), j - 1)
+    go(a, i)
+  }
```
TomasMikula: The implementation doesn't look right.

TomasMikula:
That's also an option :)

base/shared/src/main/scala/scalaz/tc/semigroup.scala
emilypi: ah - it doesn't matter, but just for sanity's sake those should be `a0` in `mappend`

jdegoes:
Wow. Looking sharp. 🎉

jdegoes:
😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢 😢

emilypi:
At @jdegoes' request, this is reopened and merged so people can work off of it in the future.

PR #1971, opened 2018-08-31
AMaybe tests

edmundnoble:
Depends on #1963.

Also rephrases the `Maybe2` docs, and adds some extra instances.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/acatenable1.scala
+++ b/base/shared/src/main/scala/scalaz/data/acatenable1.scala
@@ -12,28 +12,22 @@ import tc.{ instanceOf, Semicategory, SemicategoryClass }
 sealed abstract class ACatenable1[=>:[_, _], A, B] {
   import ACatenable1._
 
-  def compose[Z](that: ACatenable1[=>:, Z, A]): ACatenable1[=>:, Z, B] =
+  def :::[Z](that: ACatenable1[=>:, Z, A]): ACatenable1[=>:, Z, B] =
     Chain(that, this)
 
-  def andThen[Z](that: ACatenable1[=>:, B, Z]): ACatenable1[=>:, A, Z] =
+  def :++[Z](that: ACatenable1[=>:, B, Z]): ACatenable1[=>:, A, Z] =
     Chain(this, that)
 
-  def <<<[Z](that: ACatenable1[=>:, Z, A]): ACatenable1[=>:, Z, B] =
-    this compose that
+  def +:[Z](f: Z =>: A): ACatenable1[=>:, Z, B] =
+    :::(Lift(f))
 
-  def >>>[C](that: ACatenable1[=>:, B, C]): ACatenable1[=>:, A, C] =
-    this andThen that
-
-  def :+[Z](f: Z =>: A): ACatenable1[=>:, Z, B] =
-    compose(Lift(f))
-
-  def +:[Z](f: B =>: Z): ACatenable1[=>:, A, Z] =
-    andThen(Lift(f))
+  def :+[Z](f: B =>: Z): ACatenable1[=>:, A, Z] =
+    :++(Lift(f))
 
   final def foldLeft[F[_]](fa: F[A])(φ: RightAction[F, =>:]): F[B] = {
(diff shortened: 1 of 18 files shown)


edmundnoble:
Marking as WIP because testing as a free pointed set isn't done and because I want `AMaybe2` tests in here too.

base/shared/src/main/scala/scalaz/data/acatenable1.scala
```diff
@@ -54,14 +48,31 @@ sealed abstract class ACatenable1[=>:[_, _], A, B] {
...
+  /**
+   * Allows you to view the internal structure of an ACatenable1.
+   * Not stack-safe.
+   */
+  final def fold[G[_, _]](f: Forall3.Prototype[λ[(X, Y, Z) => (G[X, Y], G[Y, Z]) => G[X, Z]]])(z: =>: ~~> G): G[A, B] =
+    this match {
+      case Chain(a, b) => f.apply(a.fold[G](f)(z), b.fold[G](f)(z))
+      case Lift(a)     => z.apply(a)
+    }
+
```
emilypi: so dank

base/shared/src/main/scala/scalaz/tc/unfoldable1.scala
```diff
@@ -1,4 +1,5 @@
-package scalaz.tc
+package scalaz
+package tc
 
 import scalaz.data.Maybe
```
emilypi: 👍

base/shared/src/main/scala/scalaz/tc/debug.scala
```diff
@@ -83,6 +83,13 @@ object DebugClass {
     }
   }
 
+  implicit def tuple2Debug[A, B](implicit A: Debug[A], B: Debug[B]): Debug[(A, B)] = {
+    import Scalaz.debugInterpolator
+    DebugClass.instance[(A, B)] {
+      case (a, b) => z"($a, $b)"
```
emilypi: Can you give a quick rundown of what this interpolator does?

microsite/src/main/tut/data/Maybe2.md
```diff
@@ -6,7 +6,11 @@ title:  "Maybe2"
 
 # Maybe2 [![GitHub](../img/github.png)](https://github.com/scalaz/scalaz/blob/series/8.0.x/base/shared/src/main/scala/scalaz/data/maybe2.scala)
 
-`Maybe2` is a specialized version of [Maybe](./Maybe.html) that avoids allocation of [Tuple2](https://www.scala-lang.org/api/current/scala/Tuple2.html) instances.
+`Maybe2[A, B]` is isomorphic to `Maybe[(A, B)]`.
+The only difference between `Maybe2[A, B]` and `Maybe[(A, B)]` is that the first takes less
+memory.
+
+`Maybe2` is used to implement `IList`; each cons cell is a `Maybe2`.
```
emilypi: This is fine, but we probably need to expand this a little more for folks

tests/src/main/scala/scalaz/tests/ACatenable1Tests.scala
```diff
@@ -19,8 +19,8 @@ object ACatenable1Tests {
   // these type aliases unnerve me with their presence
   type R[A, B] = ACatenable1[Map, A, B] => Map[A, B]
   type L[A, B] = Map[A, B] => ACatenable1[Map, A, B]
-  val foldF: ACatenable1[Map, ?, ?] ~~> Map = Forall2.of[R](_.fold)
-  val liftF: Map ~~> ACatenable1[Map, ?, ?] = Forall2.of[L](lift(_))
+  val foldBalancedF: ACatenable1[Map, ?, ?] ~~> Map = Forall2.of[R](_.foldBalanced)
```
emilypi: Do we have binatural transformations documented? That should definitely be on the list as we're adding more than a few instances here.

base/shared/src/main/scala/scalaz/tc/debug.scala
emilypi: nice!

PR #1975, opened 2018-09-04
Fix compilation error from #1368

enzief:
Just swap positions of `Trampoline`/`Unit` in `ContT`.


Atry:
It seems the original PR targeted to 7.2.x, where the order is reversed.

PR #1977, opened 2018-09-06
Maybe tests

edmundnoble:
Also a little cleanup for other tests.

/cc @diminou, seeing as this touches your recent `Const` tests.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/maybe.scala
+++ b/base/shared/src/main/scala/scalaz/data/maybe.scala
@@ -33,14 +33,16 @@ sealed trait MaybeModule {
   def traversable: Traversable[Maybe]
   def debug[A: Debug]: Debug[Maybe[A]]
   def eq[A: Eq]: Eq[Maybe[A]]
+  def monoid[A: Semigroup]: Monoid[Maybe[A]]
 }
 
 object MaybeModule extends MaybeSyntax {
-  implicit def monadMaybe: Monad[Maybe]              = Maybe.monad
-  implicit def traversableMaybe: Traversable[Maybe]  = Maybe.traversable
-  implicit def isCovariantMaybe: IsCovariant[Maybe]  = Maybe.isCovariant
-  implicit def debugMaybe[A: Debug]: Debug[Maybe[A]] = Maybe.debug[A]
-  implicit def eqMaybe[A: Eq]: Eq[Maybe[A]]          = Maybe.eq[A]
+  implicit def monadMaybe: Monad[Maybe]                    = Maybe.monad
+  implicit def traversableMaybe: Traversable[Maybe]        = Maybe.traversable
+  implicit def isCovariantMaybe: IsCovariant[Maybe]        = Maybe.isCovariant
+  implicit def debugMaybe[A: Debug]: Debug[Maybe[A]]       = Maybe.debug[A]
+  implicit def eqMaybe[A: Eq]: Eq[Maybe[A]]                = Maybe.eq[A]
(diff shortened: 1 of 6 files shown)


base/shared/src/main/scala/scalaz/data/maybe.scala
```diff
@@ -127,6 +138,8 @@ trait MaybeFunctions {
 }
 
 trait MaybeSyntax {
+  implicit final class MaybeCata[A](ma: Maybe[A]) { def cata[B](f: A => B)(b: B): B = Maybe.toOption(ma).fold(b)(f) }
```
hrhino: this could just be `MaybeOps`.

also, why not reuse `MaybeModule#fold`?

tests/src/main/scala/scalaz/tests/ConstTests.scala
```diff
@@ -11,28 +11,25 @@ import testz.{ assert, Harness }
 
 import z._
 
-final class ConstTests {
+object ConstTests {
 
-  val consts = List(
-    Const[List[Int], List[Double]](List(1)),
-    Const[List[Int], List[Double]](List())
+  val consts = IList(
```
hrhino: 👍

tests/src/main/scala/scalaz/tests/ConstTests.scala
```diff
@@ -53,47 +50,47 @@ final class ConstTests {
...
+              ApplicativeLaws.applyIdentity(_)(assertEqual[Const[IList[Int], IList[Double]]])
             }
           }
         ),
         namedSection("traversable laws")(
           test("traversable composition") { () =>
-            val fst = (a: List[Double]) => a.headOption
-            val snd = (a: Double) => if (a > 0.0) scala.Some(a) else scala.None
+            val fst = (_: IList[Double]).head
+            val snd = (a: Double) => if (a > 0.0) Maybe.just(a) else Maybe.empty[Double]
```
hrhino: I would like (separately) `??` and `!?` to be added, eventually

tests/src/main/scala/scalaz/tests/MaybeTests.scala
```diff
@@ -0,0 +1,131 @@
...
+        Maybe.just("foobar"),
+      )
+
+    section(
+      namedSection("concrete")(
+        namedSection("empty")(
+          test("is cached")(
+            () => assert(Maybe.empty.asInstanceOf[scala.AnyRef] eq Maybe.empty.asInstanceOf[scala.AnyRef])
+          ),
+          test("is `None`")(() => assert(Maybe.empty.asInstanceOf[scala.AnyRef] eq scala.None)),
```
hrhino: interesting test case, but alright. I can imagine a future in which it is not.

tests/src/main/scala/scalaz/tests/MaybeTests.scala
```diff
@@ -0,0 +1,131 @@
...
+      namedSection("laws")(
+        namedSection("eq laws")(
+          test("reflexivity") { () =>
+            maybes.foldMap(
+              EqLaws.reflexivity(_)(assert)
+            )
+          },
+          test("identity") { () =>
+            maybes.cross(maybes).foldMap {
+              case (m1, m2) => assert((m1 === m2) == (m1 == m2))
```
alexknvl: This is not a very good test for `Eq`, but I'll fix all `Eq` laws later.

alexknvl: It is insufficient because it does not test that `m1 === m2` does not compare the `a: A` inside of maybes using `==`.

PR #1978, opened 2018-09-06
Dimap and lift as methods on Kleisli

edmundnoble:
Just a couple useful things.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/kleisli.scala
+++ b/base/shared/src/main/scala/scalaz/data/kleisli.scala
@@ -18,7 +18,11 @@ sealed trait KleisliModule {
 
   def second[F[_], A, B, C](k: Kleisli[F, A, B])(implicit F: Functor[F]): Kleisli[F, (C, A), (C, B)]
 
+  def dimap[F[_], A, B, C, D](k: Kleisli[F, A, B])(ca: C => A)(bd: B => D)(implicit F: Functor[F]): Kleisli[F, C, D]
+
   def compose[F[_], A, B, C](j: Kleisli[F, B, C], k: Kleisli[F, A, B])(implicit B: Bind[F]): Kleisli[F, A, C]
+
+  def lift[F[_], A, B](fb: F[B]): Kleisli[F, A, B]
 
 }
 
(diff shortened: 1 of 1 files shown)


base/shared/src/main/scala/scalaz/data/kleisli.scala
```diff
@@ -18,8 +18,12 @@ sealed trait KleisliModule {
 
   def second[F[_], A, B, C](k: Kleisli[F, A, B])(implicit F: Functor[F]): Kleisli[F, (C, A), (C, B)]
 
+  def dimap[F[_], A, B, C, D](k: Kleisli[F, A, B])(ca: C => A)(bd: B => D)(implicit F: Functor[F]): Kleisli[F, C, D]
+
   def compose[F[_], A, B, C](j: Kleisli[F, B, C], k: Kleisli[F, A, B])(implicit B: Bind[F]): Kleisli[F, A, C]
 
+  def lift[F[_], A, B](fb: F[B]): Kleisli[F, A, B]
```
TomasMikula: Not sure `lift` is a good name. What about `const`?

TomasMikula:
```scala
Function.const[B, A]   :   B  => (A =>   B )
Kleisli.const[F, A, B] : F[B] => (A => F[B])
```

edmundnoble: `lift` here is by analogy to every other monad transformer.

emilypi: @TomasMikula feel free to smack me with a newspaper when i'm being dumb 😄 

That said... _eehhhh_ `lift` still sounds a little better

hrhino: Yeah, I think I'm with @TomasMikula here; `lift` seems more apt for lifting a function rather than a wrapped value.

base/shared/src/main/scala/scalaz/data/kleisli.scala
```diff
@@ -77,18 +81,30 @@ private[data] object KleisliImpl extends KleisliModule {
   override def first[F[_], A, B, C](
     k: Kleisli[F, A, B]
   )(implicit F: Functor[F]): Kleisli[F, (A, C), (B, C)] =
-    wrapKleisli(t => F.map(runKleisli(k)(t._1))((_, t._2)))
+    t => F.map(k(t._1))((_, t._2))
```
hrhino: 😆

base/shared/src/main/scala/scalaz/data/kleisli.scala
```diff
@@ -77,18 +81,30 @@ private[data] object KleisliImpl extends KleisliModule {
...
+    t => F.map(k(t._2))((t._1, _))
+
+  override def dimap[F[_], A, B, C, D](
+    k: Kleisli[F, A, B]
+  )(
+    ca: C => A
+  )(
+    bd: B => D
+  )(implicit F: Functor[F]): Kleisli[F, C, D] =
+    c => F.map(k(ca(c)))(bd)
```
hrhino: should this delegate to the `Profunctor` instance or _vice versa_? not like either is likely to change.

base/shared/src/main/scala/scalaz/data/kleisli.scala
edmundnoble: That's fair, didn't think of separating ReaderT and Kleisli.

base/shared/src/main/scala/scalaz/data/kleisli.scala
edmundnoble: Think it should mostly be the instance delegating to the functions.

PR #1979, opened 2018-09-06
Ordering instances/tests, clearer test section names

edmundnoble:


base/shared/src/main/scala/scalaz/tc/eq.scala
```diff
@@ -102,5 +102,8 @@ object Eq {
   def always[A]: Eq[A]                   = instanceOf[EqClass[A]]((_, _) => true)
   def never[A]: Eq[A]                    = instanceOf[EqClass[A]]((_, _) => false)
```
hrhino: off topic, but ... what is this for, again? isn't it unlawful to have a irreflexive `Eq`?

tests/src/main/scala/scalaz/tests/AMaybeTests.scala
```diff
@@ -14,29 +14,31 @@ object AMaybeTests {
...
+        test("eq") { () =>
+          val testData =
+            IList(
+              empty[(?, ?), Int],
+              just[(?, ?), Int, Int]((0, 1)),
+              just[(?, ?), Int, Int]((2, 3)),
+            )
+          testData.cross(testData).foldMap {
+            case (a, b) =>
+              assert((a === b) == (a eq b))
```
hrhino: so equality on `AMaybe` is by reference? that almost seems like it shouldn't have equality at all, or that its equality should depend on some property of the relation parameter.

TomasMikula: I think this test is meant to be read as "the elements _in this case_ are all distinct, so two elements should only be equal if they are the same element". It would be better if this test more explicitly expressed that only pairs on the diagonal of the cross product are equal.

edmundnoble: How would I express that directly?

TomasMikula: `zipWithIndex` before `cross`, and then

```scala
case ((i, a), (j, b)) =>
  assert((a === b) == (i == j))
```

edmundnoble: Fixed.

PR #1981, opened 2018-09-16
removed 's'

0xflotus:


xuwei-k:
https://en.wiktionary.org/wiki/focussed

> The spelling focused is much more common in the US but also more common in the UK and Australia.

🤔

0xflotus:
Oh, I didn't know that. Nice to know. Thanks, so we can close the PR.

PR #1989, opened 2018-09-21
Use a stable ZIO release, update instances for it, rudimentary IO tests

hrhino:
Depending on a snapshot version of zio is unwise, since changes there can break us.

The `0.1-SNAPSHOT` release wasn't resolving at all.

This uses the 0.2.7 stable release, and updates `Applicative` and `Monoid`'s `Fiber` instances to use `Fiber` methods rather than directly subclassing `Fiber`.

Also, some tests for `IO`'s instances.

### DIFF
--- a/tests/src/main/scala/scalaz/tests/IOTests.scala
+++ b/tests/src/main/scala/scalaz/tests/IOTests.scala
@@ -1,0 +1,107 @@
+package scalaz
+package tests
+
+import scala.{ List, Range, StringContext }
+import scala.Predef.String
+
+import Scalaz._
+
+import testz._, z._
+import laws._, FunctorLaws._
+
+import zio._
+
+object IOTests {
+
+  case class Error(msg: String)
+
+  // a probably-too-large selection of IO actions; more could be added
+  val ios = {
+    type IOES = IO[Error, String]
+    type Endo = IOES => IOES
+    val seeds = List[IOES](
+      IO.now("now"),
+      IO.unit.map(_ => "unit"),
+      IO.point("point"),
+      IO.fail(Error("error")),
+      //IO.sleep(scala.concurrent.duration.Duration.fromNanos(1)).map(_ => "slept")
+    )
+    val endos0: List[Endo] =
+      seeds.map(s => (io: IOES) => io.flatMap(_ => s)) :::
+        List(fns.s1, fns.s2, fns.f).map(f => (io: IOES) => io.flatMap(f)) :::
+        List[Endo](
+        _.map(_ + "ishly"),
+      )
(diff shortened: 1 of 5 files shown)


tests/src/main/scala/scalaz/tests/IOTests.scala
```diff
@@ -0,0 +1,107 @@
...
+  val ios = {
+    type IOES = IO[Error, String]
+    type Endo = IOES => IOES
+    val seeds = List[IOES](
+      IO.now("now"),
+      IO.unit.map(_ => "unit"),
+      IO.point("point"),
+      IO.fail(Error("error")),
+      //IO.sleep(scala.concurrent.duration.Duration.fromNanos(1)).map(_ => "slept")
+    )
```
edmundnoble: This is good *for now* but we should probably have some tests that the laws work w.r.t `sync` and `async`, which will require more specific IO actions.

tests/src/main/scala/scalaz/tests/IOTests.scala
```diff
@@ -0,0 +1,107 @@
...
+      .flatMap(l => l)
+    seeds.flatMap(s => endos.map(_(s)))
+  }
+
+  object RTS extends RTS
+
+  def ioEql[E, A](l: IO[E, A], r: IO[E, A]): Result =
+    assert(RTS.unsafeRunSync(l.attempt) == RTS.unsafeRunSync(r.attempt))
+
+  object fns {
```
edmundnoble: Why not just define this as a list of `fns`?

tests/src/main/scala/scalaz/tests/IOTests.scala
```diff
@@ -0,0 +1,107 @@
...
+          },
+          test("success/failure") { () =>
+            ios.foldMap(BindLaws.bindAssoc(_)(fns.s1, fns.f)(ioEql))
+          },
+        )
+      ),
+      namedSection("monad")(
+        namedSection("bindLeftIdentity")(
+          test("success") { () =>
+            MonadLaws.bindLeftIdentity("mxyplyzyk")(fns.s1)(ioEql)
```
edmundnoble: You wot?

tests/src/main/scala/scalaz/tests/IOTests.scala
hrhino: I want to be able to do all pairs of success/failure in the bind associativity test

tests/src/main/scala/scalaz/tests/IOTests.scala
hrhino: woops, typo

PR #1992, opened 2018-09-22
Benchmark tests

edmundnoble:
If you'd like to benchmark the tests, this is a good way. We make sure to fork 5 JVMs and only actually run the benchmark once per iteration for realistic conditions (no JIT, classloading has to be redone on each iteration). The best way to use this is with async-profiler, as described at https://github.com/ktoso/sbt-jmh.

### DIFF
--- a/benchmarks/src/main/scala/scalaz/tests/TestBenchmark.scala
+++ b/benchmarks/src/main/scala/scalaz/tests/TestBenchmark.scala
@@ -15,9 +15,6 @@ import scalaz.Predef._
 @Measurement(iterations = 1, batchSize = 1, time = 1, timeUnit = TimeUnit.MILLISECONDS)
 class TestBenchmark {
   @Benchmark
-  def allTests(): Unit = {
+  def allTests(): Unit =
     TestMain.main(Array.empty)
-  }
 }
-
-
--- a/benchmarks/src/main/scala/scalaz/tests/TestBenchmark.scala
+++ b/benchmarks/src/main/scala/scalaz/tests/TestBenchmark.scala
@@ -1,0 +1,23 @@
+package scalaz
+package tests
+
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+import scala.Array
+import scala.Predef.{ wrapString, String }
+import scalaz.Predef._
+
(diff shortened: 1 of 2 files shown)


benchmarks/src/main/scala/scalaz/tests/TestBenchmark.scala
```diff
@@ -0,0 +1,20 @@
...
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+@Fork(value = 5, jvmArgsAppend = Array("-Xms2g", "-Xmx2g"))
+@Warmup(iterations = 0, batchSize = 1)
+@Measurement(iterations = 1, batchSize = 1, time = 1, timeUnit = TimeUnit.MILLISECONDS)
+class TestBenchmark {
+  @Benchmark
+  def allTests(): Unit =
+    TestMain.main(Array.empty)
```
hrhino: I'd be curious about benchmarking individual tests but nm

benchmarks/src/main/scala/scalaz/tests/TestBenchmark.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+import scala.Array
+import scala.Predef.{ wrapString, String }
+import scalaz.Predef._
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
```
hrhino: oh come on, they don't take _seconds_...

edmundnoble: Fair, though I don't really use the outputted time. Just the profiler. Also, it shows fractional seconds.

benchmarks/src/main/scala/scalaz/tests/TestBenchmark.scala
edmundnoble: You can definitely do that; just add a method like `allTests` that calls into a particular test suite.

PR #1993, opened 2018-09-22
Add Lazy data and Delay type class

edmundnoble:
Tests incoming.

I believe this is the beginning of a way to solve the lazy type classes problem in Scala.

### DIFF
--- a/base/shared/src/main/scala/scalaz/data/Lazy.scala
+++ b/base/shared/src/main/scala/scalaz/data/Lazy.scala
@@ -1,0 +1,30 @@
+package scalaz
+package data
+
+import tc._
+
+class Lazy[A](val run: () => A) {
+  lazy val value = run()
+}
+
+object Lazy {
+  implicit def lazyDelay[A]: Delay[Lazy[A]] =
+    instanceOf[DelayClass[Lazy[A]]](fa => new Lazy[A](() => fa().run()))
+
+  implicit def lazyMonad: Monad[Lazy] = instanceOf(new MonadClass[Lazy] {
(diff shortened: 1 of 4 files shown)


edmundnoble:
Sure, will do tomorrow.

jbgi:
I work on a similar `Thunkable` type class last year but did not have time to submit a PR. This where I left off: https://github.com/jbgi/scalaz/commit/7c116ec6468aadcbc33d06b915215f464406278b
The idea was, whenever possible, scalaz data types would come with a lazy constructor of type `(=>A) => A`.

Milyardo:
Should there be a law that says `delay(() => a) == a`?

edmundnoble:
@Milyardo We could have that, though what makes for `==` would have to not include the semantics of evaluating "whatever's inside" `a`.

base/shared/src/main/scala/scalaz/data/Lazy.scala
```diff
@@ -0,0 +1,31 @@
+package scalaz
+package data
+
+import tc._
+
+class Lazy[A](val run: () => A) {
```
hrhino: `final`?

base/shared/src/main/scala/scalaz/tc/delay.scala
```diff
@@ -0,0 +1,14 @@
+package scalaz
+package tc
+
+/**
+ * Given `Delay[A]`, one can suspend execution of an `A` inside of the `A` itself.
+ */
+trait DelayClass[A] {
```
hrhino: so what's a good use case for `Delay`?

PR #1998, opened 2018-09-25
Adding applicative, monoid, profunctor, and strong instances for schedule

pfgray:
Asked for in #1940


pfgray:
Arrow doesn't exist yet on the 8.0.x branch

base/shared/src/main/scala/scalaz/tc/monoid.scala
```diff
@@ -34,4 +34,12 @@ object MonoidClass {
         a1.zipWith(a2)(A.mappend(_, _))
       def mempty: Fiber[E, A] = Fiber.point(A.mempty)
     })
+
+  implicit def strategyMonoid[E, A](implicit A: Monoid[A]): Monoid[Schedule[E, A]] =
```
edmundnoble: Why not `scheduleMonoid`?

PR #2000, opened 2018-09-26
add syntax for Const construction with nicer inference

hrhino:


core/src/main/scala/scalaz/syntax/ConstOps.scala
```diff
@@ -0,0 +1,15 @@
+package scalaz
+package syntax
+
+/** Provides [[Const]]-related methods for any value. */
+final class ConstOps[A](val self: A) extends AnyVal {
+  /** Wrap `self` in a [[Const]].
+    *
+    * Can have better type inference than the [[Const.apply]] constructor.
+    */
+  def const[B]: Const[A, B] = Const(self)
```
edmundnoble: Excellent idea.

core/src/main/scala/scalaz/syntax/ConstOps.scala
```diff
@@ -0,0 +1,15 @@
+package scalaz
+package syntax
+
+/** Provides [[Const]]-related methods for any value. */
+final class ConstOps[A](val self: A) extends AnyVal {
```
xuwei-k: `private val self: A` ?

hrhino: Did we drop 2.10 in 7.3? I guess we did

Issue #2002, opened 2018-09-26
Add EitherT.fromOption


allantl:
I noticed that `EitherT` doesn't have this currently.
I'd like to help contribute.

WDYT?

hrhino:
A pull request would be the best way to suggest this 😉. It sounds useful.

jozi-k:
You might consider closing this issue.

hrhino:
nice catch; consider it considered! apparently github doesn't understand "addresses".

PR #2006, opened 2018-09-27
Add EitherT fromOption

allantl:
Addresses #2002

### DIFF
--- a/core/src/main/scala/scalaz/Either.scala
+++ b/core/src/main/scala/scalaz/Either.scala
@@ -397,6 +397,10 @@ object \/ extends DisjunctionInstances {
   def fromEither[A, B](e: Either[A, B]): A \/ B =
     e fold (left, right)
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/scalaz/Either.scala
```diff
@@ -397,6 +397,10 @@ object \/ extends DisjunctionInstances {
   def fromEither[A, B](e: Either[A, B]): A \/ B =
     e fold (left, right)
 
+  /** Construct a disjunction value from a standard `scala.Option`. */
+  def fromOption[A, B](o: Option[B], ifNone: => A): A \/ B =
```
edmundnoble: Personally I think this would be nicer curried as `(ifNone: => A)(o: Option[B])`

allantl:
Comments addresed

emilypi:
@edmundnoble I think we're good on this one.

PR #2007, opened 2018-09-28
Remove Eq.never

### LINKED ISSUE #1994: Delete `Eq.never`
There's only one type for which `Eq.never` is a lawful instance, and that's `Nothing`. Reflexivity is a required law.

iblamefish:
resolves #1994


alexknvl:
Theoretically there could be some instances for types that are isomorphic to `Nothing`.

hrhino:
> But the value of Eq[Nothing] is somewhat doubtful anyway, so...

also you can just as easily use `Eq.always` for that 😉 

Thanks, @iblamefish!

Issue #1991, opened 2018-09-21
get stream/vector out of our trees


hrhino:
there's no reason to use `Vector` for `StrictTree`. In no way that I know of is it superior to doing something else. `Stream` could probably benefit from becoming `EStream` in `Tree`. Benchmarks would help.

edmundnoble:
But it's *balanced* :wink:

PR #2008, opened 2018-09-29
Feature/stricty mc list tree

merlinorg:
Switch `StrictTree` from `Vector` to `List`. Addresses half of #1991.

### DIFF
--- a/core/src/main/scala/scalaz/StrictTree.scala
+++ b/core/src/main/scala/scalaz/StrictTree.scala
@@ -1,7 +1,7 @@
 package scalaz
 
 import scala.collection.mutable
-import std.vector.{vectorInstance, vectorMonoid}
+import std.list.{listInstance, listMonoid}
 
 /**
   *
@@ -11,7 +11,7 @@ import std.vector.{vectorInstance, vectorMonoid}
   */
 case class StrictTree[A](
   rootLabel: A,
-  subForest: Vector[StrictTree[A]]
+  subForest: List[StrictTree[A]]
 ) {
 
   import StrictTree._
@@ -51,7 +51,7 @@ case class StrictTree[A](
     runBottomUp(foldMapReducer(f))
 
   def foldRight[B](z: B)(f: (A, => B) => B): B =
-    Foldable[Vector].foldRight(flatten, z)(f)
+    Foldable[List].foldRight(flatten, z)(f)
 
   /** A 2D String representation of this StrictTree. */
   def drawTree(implicit sh: Show[A]): String = {
@@ -62,11 +62,11 @@ case class StrictTree[A](
     * is a function of the corresponding element in this tree
     * and the histomorphic transform of its children.
     */
-  def scanr[B](g: (A, Vector[StrictTree[B]]) => B): StrictTree[B] =
+  def scanr[B](g: (A, List[StrictTree[B]]) => B): StrictTree[B] =
     runBottomUp(scanrReducer(g))
 
   /** Pre-order traversal. */
-  def flatten: Vector[A] = {
+  def flatten: List[A] = {
     val stack = mutable.Stack(this)
 
     val result = mutable.Buffer.empty[A]
@@ -77,7 +77,7 @@ case class StrictTree[A](
       popped.subForest.reverseIterator.foreach(stack.push)
     }
 
-    result.toVector
+    result.toList
   }
 
   def size: Int = {
@@ -95,11 +95,23 @@ case class StrictTree[A](
   }
 
   /** Breadth-first traversal. */
-  def levels: Vector[Vector[A]] = {
-    val f = (s: Vector[StrictTree[A]]) => {
-      Foldable[Vector].foldMap(s)((_: StrictTree[A]).subForest)
+  def levels: List[List[A]] = {
+    val f = (s: List[StrictTree[A]]) => {
+      Foldable[List].foldMap(s)((_: StrictTree[A]).subForest)
     }
-    Vector.iterate(Vector(this), size)(f) takeWhile (!_.isEmpty) map (_ map (_.rootLabel))
+    List.iterate(List(this), depth)(f) map { _ map (_.rootLabel) }
+  }
+
+  def depth: Int = {
+    var level = List(this)
+    var result = 0
+
+    while (level.nonEmpty) {
+      level = Foldable[List].foldMap(level)(_.subForest)
+      result += 1
+    }
+
+    result
   }
 
   def toTree: Tree[A] = {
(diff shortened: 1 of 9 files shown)


core/src/main/scala/scalaz/StrictTree.scala
```diff
@@ -95,11 +95,23 @@ case class StrictTree[A](
...
 
   /** Breadth-first traversal. */
-  def levels: Vector[Vector[A]] = {
-    val f = (s: Vector[StrictTree[A]]) => {
-      Foldable[Vector].foldMap(s)((_: StrictTree[A]).subForest)
+  def levels: List[List[A]] = {
+    val f = (s: List[StrictTree[A]]) => {
+      Foldable[List].foldMap(s)((_: StrictTree[A]).subForest)
     }
-    Vector.iterate(Vector(this), size)(f) takeWhile (!_.isEmpty) map (_ map (_.rootLabel))
```
merlinorg: The choice of `size` seems to have just been profoundly wrong, it created a `Vector` the size of the entire tree and then shrunk it.

core/src/main/scala/scalaz/std/List.scala
```diff
@@ -67,17 +67,11 @@ trait ListInstances extends ListInstances0 {
         }
         (a, b) => loop(a, b, Nil)
       }
-      def traverseImpl[F[_], A, B](l: List[A])(f: A => F[B])(implicit F: Applicative[F]) = {
```
merlinorg: This crazy traverse blew the stack.

merlinorg: In a single scientific test this builder based traverse ran about 15% faster than the original.

core/src/main/scala/scalaz/std/List.scala
```diff
@@ -67,17 +67,11 @@ trait ListInstances extends ListInstances0 {
...
-            case a :: as => Maybe.just((f(a), as))
-            case Nil => Maybe.empty
-          })(Reducer.ReverseListReducer[B])
 
-        val rev: F[List[B]] = revOpt getOrElse F.point(Nil)
-
-        F.map(rev)(_.reverse)
-      }
+      def traverseImpl[F[_], A, B](l: List[A])(f: A => F[B])(implicit F: Applicative[F]) =
+        F.map(l.reverse.foldRight(F.point(l.genericBuilder[B])) { (a, flb) =>
```
merlinorg: In a single scientific test this builder based traverse ran 15% faster than the original.

core/src/main/scala/scalaz/std/List.scala
merlinorg: Anecdotally I also find the replacement traverse to be faster than the old.

core/src/main/scala/scalaz/StrictTree.scala
```diff
@@ -11,7 +11,7 @@ import std.vector.{vectorInstance, vectorMonoid}
   */
 case class StrictTree[A](
   rootLabel: A,
-  subForest: Vector[StrictTree[A]]
+  subForest: List[StrictTree[A]]
```
merlinorg: I find the `List` tree to perform better than the `Vector` tree for all operations excepting those for which `Vector` has decidedly better characteristics, the most notable of those being `foldRight`.

core/src/main/scala/scalaz/StrictTree.scala
```diff
@@ -262,44 +268,42 @@ object StrictTree extends StrictTreeInstances {
...
     f(v) match {
       case (a, bs) => Node(a, unfoldForest(bs)(f))
     }
 
   //Only used for .equals.
   private def badEqInstance[A] = new StrictTreeEqual[A] {
-    override def A: Equal[A] = new Equal[A] {
-      override def equal(a1: A, a2: A): Boolean = a1.equals(a2)
-    }
+    override def A: Equal[A] = _ equals _
```
xuwei-k:
```
[error] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/StrictTree.scala:294:32: missing parameter type for expanded function ((x$28, x$29) => x$28.equals(x$29))
[error]     override def A: Equal[A] = _ equals _
[error]                                ^
[error] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/StrictTree.scala:294:41: missing parameter type for expanded function ((x$28: <error>, x$29) => x$28.equals(x$29))
[error]     override def A: Equal[A] = _ equals _
[error]                                         ^
[error] two errors found
```

https://travis-ci.org/scalaz/scalaz/jobs/436192803#L543

core/src/main/scala/scalaz/StrictTree.scala
```diff
@@ -170,9 +153,8 @@ case class StrictTree[A](
     *
     * @return
     */
-  override def hashCode(): Int = {
-    runBottomUp(hashCodeReducer)
-  }
+  override def hashCode(): Int =
+    MurmurHash3.listHash(rflatten, "StrictTree".hashCode)
```
merlinorg: unclear how ok this is, the hashcode is invariant in tree shape, may reconsider

hrhino: what if we just delete this and `equals`?

core/src/main/scala/scalaz/StrictTree.scala
```diff
@@ -48,10 +48,21 @@ case class StrictTree[A](
...
-    runBottomUp(foldMapReducer(f))
+    foldLeft(Monoid[B].zero)((a, b) => Monoid[B].append(b, f(a)))
+
+  def foldLeft[B](z: B)(f: (A, B) => B): B = {
+    var stack = this :: Nil
+    var result = z
+    while (stack.nonEmpty) {
+      val head :: tail = stack
+      result = f(head.rootLabel, result)
+      stack = head.subForest ::: tail
```
hrhino: hmm. this could avoid a `:::` by having `stack` be a `List[List[A]]` perhaps

```scala
var stack = List(this) :: Nil
var result = z
while (stack.nonEmpty) {
  val (head :: tail) :: rest = stack
  result = f(head.rootLabel, result)
  stack = (head.subForest :: tail :: stack) dropWhile (_.isEmpty)
}
result
```

or something like that (but tested)

core/src/main/scala/scalaz/StrictTree.scala
```diff
@@ -124,42 +120,41 @@ case class StrictTree[A](
...
     * @return
     */
-  override def hashCode(): Int = {
-    runBottomUp(hashCodeReducer)
-  }
+  override def hashCode(): Int =
+    MurmurHash3.listHash(rflatten, "StrictTree".hashCode)
 
   override def equals(obj: scala.Any): Boolean = {
     obj match {
```
hrhino: yeah, this is pretty bad.

PR #2011, opened 2018-10-03
Plus for Validation accumulates errors

### LINKED ISSUE #1976: Plus for Validation is "orElse"
I think Plus for Validation should be "findSuccess" (i.e. it uses the Semigroup for errors) instead of "orElse" (which is the same implementation as Disjunction)

merlinorg:
Closes #1976.

### DIFF
--- a/core/src/main/scala/scalaz/Validation.scala
+++ b/core/src/main/scala/scalaz/Validation.scala
@@ -505,8 +505,8 @@ sealed abstract class ValidationInstances1 extends ValidationInstances2 {
 }
 
 sealed abstract class ValidationInstances2 extends ValidationInstances3 {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/Validation.scala
```diff
@@ -530,7 +530,7 @@ sealed abstract class ValidationInstances2 extends ValidationInstances3 {
         fa.fold(l => -\/(Failure(l)), \/.right)
     }
 
-  implicit def ValidationPlus[E: Semigroup, A]: Plus[Validation[E, ?]] = new Plus[Validation[E, ?]] {
```
hrhino: approving reviewers feel shame

PR #2015, opened 2018-10-11
[WIP] Need

alexknvl:


jbgi:
I guess we would want this to be thread-safe?

alexknvl:
I believe it is thread-safe to the extent that some sub-expressions may get evaluated twice and unless `A` is safely published sharing `Need[A]` is unsafe. Considering that Scala 2.13 will have safe-published `List[A]` and other data structures, I don't think it is a problem.

jbgi:
I heard that x86 is the wrong platform to check for safe-publishing. ARM would be more likely to expose flaws.
But I think the main issue is, do we want to allow by-need expression to be evaluated twice? It could be an issue if the computation of the expression is expensive. Maybe we need two variants?

alexknvl:
Adding `final class Need[+A] private (private var thunk: Any, private val x: Int = 1)` fixes safe-initialization of the `Need` itself. 

I think we can add a separate constructor for "at most once" semantics.

jbgi:
> but having "at most once" constructors and stack-safety may be difficult.

Yeah, I think a good compromise would to be to guarantee "at most once" only for the most outer expression (and the rest on a best-effort basis): this is the one that could likely be shared among threads.

edmundnoble:
It's trampolined but afaik it *only works* if you're using laziness, because it operates via graph rewriting.

alexknvl:
I have been working on a library that uses `Need` to implement some basic lazy data structures and it seems to be pretty versatile https://github.com/alexknvl/lazyboi/tree/master/src/main/scala/lazyboi

PR #2017, opened 2018-10-19
Add Comonad syntax

japgolly:


japgolly:
Silly me not realising there was Cobind. Take 2.
Also the reason I removed `final` modifiers from methods in `CobindSyntax` is because the class itself is already final.

PR #2020, opened 2018-10-22
Day convolution implementation

lemastero:


lemastero:
I was able to implement both map and cobind without any restrictions on F and G.
Day convolution is both Functor and Cobind for free!

PR #2023, opened 2018-10-31
Documentation for methods are added

chetkhatri:
What were proposed in this Pull Request?
* Method documentation of Foldable Class is added.

@jdegoes Please review and suggest.

### DIFF
--- a/base/shared/src/main/scala/scalaz/tc/foldable.scala
+++ b/base/shared/src/main/scala/scalaz/tc/foldable.scala
@@ -6,13 +6,24 @@ import scala.List
 import scala.language.experimental.macros
 
 trait FoldableClass[F[_]] {
+  /**
+  Applies a binary operator to a start value and all elements of this sequence, going left to right.
+  Note: this will not terminate for infinite-sized collections. For Example, Stream.
+  */
   def foldLeft[A, B](fa: F[A], z: B)(f: (B, A) => B): B
-
+  /**
+  Applies a binary operator to all elements of this list and a start value, going right to left.
+  */
   def foldRight[A, B](fa: F[A], z: => B)(f: (A, => B) => B): B
-
+  /**
+  It traverse the collection type A and does transform to B by using F function.
+  */
   def foldMap[A, B](fa: F[A])(f: A => B)(implicit B: Monoid[B]): B =
     foldRight(fa, B.mempty)((a, b) => B.mappend(f(a), b))
 
+  /**
+  It does addition by initializing with 0 with foldLeft and Identity value of particular monoid.
+  */
   def msuml[A](fa: F[A])(implicit A: Monoid[A]): A =
     foldLeft(fa, A.mempty)(A.mappend(_, _))
 


base/shared/src/main/scala/scalaz/tc/foldable.scala
```diff
@@ -6,13 +6,24 @@ import scala.List
...
   def foldLeft[A, B](fa: F[A], z: B)(f: (B, A) => B): B
-
+  /**
+  Applies a binary operator to all elements of this list and a start value, going right to left.
+  */
   def foldRight[A, B](fa: F[A], z: => B)(f: (A, => B) => B): B
-
+  /**
+  It traverse the collection type A and does transform to B by using F function.
+  */
```
emilypi: This one needs to be expanded. It makes no mention of the fact that the translation `a -> b` where `Monoid b => b`, allows the function to collapse the functor using `mappend`.

base/shared/src/main/scala/scalaz/tc/foldable.scala
```diff
@@ -6,13 +6,24 @@ import scala.List
...
+  /**
+  Applies a binary operator to a start value and all elements of this sequence, going left to right.
+  Note: this will not terminate for infinite-sized collections. For Example, Stream.
+  */
   def foldLeft[A, B](fa: F[A], z: B)(f: (B, A) => B): B
-
+  /**
+  Applies a binary operator to all elements of this list and a start value, going right to left.
+  */
   def foldRight[A, B](fa: F[A], z: => B)(f: (A, => B) => B): B
```
emilypi: Adding some extra comment noting that `foldRight` and `foldMap`, in contrast to `foldLeft`, will evaluate lazily would be nice. Be sure to state what is and is not evaluated according to which semantic

base/shared/src/main/scala/scalaz/tc/foldable.scala
```diff
@@ -6,13 +6,24 @@ import scala.List
...
-
+  /**
+  It traverse the collection type A and does transform to B by using F function.
+  */
   def foldMap[A, B](fa: F[A])(f: A => B)(implicit B: Monoid[B]): B =
     foldRight(fa, B.mempty)((a, b) => B.mappend(f(a), b))
 
+  /**
+  It does addition by initializing with 0 with foldLeft and Identity value of particular monoid.
+  */
```
emilypi: This one is incorrect. `Monoid a => a` can be any monoid, not just addition with 0. It could be a function composition, string appends with "", list append with [], multiplication with 1, modular arithmetic with 0, etc. This just defers to the monoid instance to choose how the fold will be executed.

base/shared/src/main/scala/scalaz/tc/foldable.scala
```diff
@@ -6,13 +6,24 @@ import scala.List
 import scala.language.experimental.macros
 
 trait FoldableClass[F[_]] {
+  /**
+  Applies a binary operator to a start value and all elements of this sequence, going left to right.
+  Note: this will not terminate for infinite-sized collections. For Example, Stream.
+  */
```
emilypi: Great, thank you. Making note of the evaluation semantic would be nice (i.e. evaluates strictly, lazily, etc) instead of saying "this will not terminate". For example, what if my Stream is finite? It works just fine! It's enough to say it is lazy/strict.

emilypi:
@chetkhatri are you still around to handle this?

emilypi:
Closing for inactivity

PR #2028, opened 2018-11-16
With Pro Strength comes monoidal and dinatuaral laws!

lemastero:
1) Add `uncurry` for `Strong` Profunctor, [based on Haskell](http://hackage.haskell.org/package/profunctors/docs/Data-Profunctor-Strong.html#v:uncurry-39-)
2) Add default implementation for `second`
3) Add laws to Strong Profunctor, [based on Haskell](http://hackage.haskell.org/package/profunctors-5.3/docs/Data-Profunctor-Strong.html#v:first-39-) and section 7 of [Notions of Computation as Monoids by E. Rivas and M. Jaskelioff](https://arxiv.org/abs/1406.4823)
4) add strong laws testing for Function1 and Kleisli

cc @puffnfresh


lemastero:
@emilypi thank you very much for the review :)
I've added tests for laws and new method. Could you please take a look :eyes:  if this is satisfactory? I am open to suggestions.

I've noticed Profunctors did not have laws as well so I've added some basic law tests for them as well.

I think it would be nice to merge this work. Strength :muscle: shall not be lawless any more!

lemastero:
All the laws from Profunctor library are translated except those:
`lmap (second f) . first' ≡ rmap (second f) . first`
`lmap (first f) . second' ≡ rmap (first f) . second'`

I've reported one issue about their formulation: https://github.com/ekmett/profunctors/pull/66 but I need to understand this a bit more to express it in Scala.

emilypi:
@lemastero looks great! Thank you. I'm much happier with the state of the laws here.

emilypi:
Great work @lemastero!

PR #2029, opened 2018-11-17
Density comonad

lemastero:
Density Comonad based on Haskell implementation: [kan-extensions/Control.Comonad.Density](https://hackage.haskell.org/package/kan-extensions/docs/Control-Comonad-Density.html)

Density complete the duality matrix of kan extensions:

| Yoneda | Codensity |  Kan |
| Coyoneda |  Density | Lan |

It is simpler version of Left Kan Extension. It is also Comonad for free like Day, CoFree.

It is already implemented in Purescript and Haskell:
  * (Purescript) [rightfold/purescript-density-codensity Density](https://github.com/rightfold/purescript-density-codensity/blob/master/src/Control/Comonad/Density.purs)
  * (Haskell) [ekmett/kan-extensions Density](https://hackage.haskell.org/package/kan-extensions/docs/Control-Comonad-Density.html)

Some methods in Haskell Control.Monad.Co use it: [diter](http://hackage.haskell.org/package/kan-extensions-5.2/docs/Control-Monad-Co.html#v:diter) [dctrlM](http://hackage.haskell.org/package/kan-extensions/docs/Control-Monad-Co.html#v:dctrlM)

It was studied on blogs:
 * [Kan Extensions](http://comonad.com/reader/2008/kan-extensions/)
 * [A Product of an Imperfect Union](http://comonad.com/reader/2011/a-product-of-an-imperfect-union/)
 * [Comonads from Monads, and a new way do the reverse - u/King_of_the_Homeless](https://www.reddit.com/r/haskell/comments/8hgub9/comonads_from_monads_and_a_new_way_do_the_reverse/)

People play around with it:
* some [old code by Edward](https://github.com/ekmett/ekmett.github.com/blob/8d3abab5b66db631e148e1d046d18909bece5893/haskell/Origami.hs#L225-L226)
* [unktomi/stuff] https://github.com/unktomi/stuff/blob/master/containers/AbstractNonsense.scala#L382
* [stephen-lazaro/persians](https://github.com/stephen-lazaro/persians/blob/master/src/main/scala/sl/persians/Density.scala)

I have turned the relationship between Density and Lan as laws:
`lanToDensity(d.densityToLan) == d`
`lanToDensity(l).densityToLan == l`

Probably it would be nice to add, as a law, relationship between density and Cobind:
`d.lowerDensity andThen liftDensity = d`

I think it is worth to add, so people might play around with it. WDYT?


core/src/main/scala/scalaz/Density.scala
```diff
@@ -0,0 +1,98 @@
...
+sealed abstract class DensityInstances0 extends DensityInstances1 {
+
+  /** Density is a Cobind for free */
+  implicit def cobindInstance[K[_]]: Cobind[Density[K, ?]] = new DensityCobind[K] {}
+}
+
+sealed abstract class DensityInstances1 {
+
+  /** Density is a Functor for free */
+  implicit def functorInstance[F[_]]: Functor[Density[F, ?]] = new DensityFunctor[F]{}
```
xuwei-k: `functorInstance` and `cobindInstance` unnecessary because same restriction as `comonadInstance`.

lemastero: Fixed: https://github.com/scalaz/scalaz/pull/2029/commits/c37a7fb88e80758d4b2208452fc89ec1803acb91

lemastero:
@stephen-lazaro I have noticed you have done some great job at implementing Density (https://github.com/stephen-lazaro/persians/blob/master/src/main/scala/sl/persians/Density.scala). In particular you got further in implementing Apply and Applicative instances and some nice work around Cofree. Do you want to contribute to this PR?

stephen-lazaro:
@lemastero Sure! I would be happy to. Thanks for thinking of me!

core/src/main/scala/scalaz/Density.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  /** Density is a free ComonadTrans */
+  implicit val comonadTransInstance: ComonadTrans[Density] = new ComonadTrans[Density] {
+    def lower[G[_], A](a: Density[G, A])(implicit C: Cobind[G]): G[A] =
+      a.lowerDensity
+  }
+}
+
+/** Density is a free Comonad */
+private trait DensityComonad[F[_]] extends Comonad[Density[F, ?]] with DensityCobind[F] {
```
xuwei-k: We can just define all methods in `DensityComonad` and remove `DensityCobind` and `DensityFunctor`.

lemastero: DensityCobind and DensityFunctor removed: https://github.com/scalaz/scalaz/pull/2029/commits/4e3d5746105a82f7cbc6ba677a21a4c367212968

lemastero:
> @lemastero Sure! I would be happy to. Thanks for thinking of me!

@stephen-lazaro It seems Density Comonad is interesting concept yet has not much application.
If you want, you can create PR with Apply/Applicative instances for Density, but chances are, that it will not end up in Scalaz :(

emilypi:
@lemastero I've proposed `Density` before for sz8 and sz7 (and implemented `Codensity` for the former), however, I was met with the same utilitarian arguments. Personally, I think that if it is a useful abstraction for study, or completeness, then it should be written, however, I also understand that this produces noise and slower compile times for users. Chances are other people will disagree, and it's certainly a valid argument going the other way. I hope it goes in though.

stephen-lazaro:
The symmetry is useful. I'm happy to add the instances regardless, as I think they are interesting.

PR #2039, opened 2019-02-09
FingerTree does not need a Monoid on node summaries, Semigroup is sufficient.

TomasMikula:
We just have to accept that an empty tree has no measure (as in `Maybe.Empty` instead of `Monoid.zero`).


TomasMikula:
@jbgi This change is similar to removing Monoid from Reducer that you have done in the past. Would you have a look at this PR?

NeQuissimus:
LGTM but I don't think I am qualified :)

TomasMikula:
Thanks for giving it a look, @NeQuissimus.

Issue #2040, opened 2019-02-19
Enum fromStepToL infinite loop


antonkw:
Enum's `|-->` can skip max value of type and go to infinite lopp.
In case if returned by `succn` instance "jump" over the max value of type it is possible that it will produce big invalid list or go to infinite loop.
Example:
`110.toByte |--> (2, 127.toByte)`
https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Enum.scala#L183 - here k value jumps from 126 to -128, so recursion evaluates forever (exit condition is "k is bigger than end of  configured range (`127.toByte`) ", but k just jumps over to -128).

This issue was discussed in gitter channel, will work on PR to fix it.
Please assign it to me.

antonkw:
PR: https://github.com/scalaz/scalaz/pull/2041

PR #2041, opened 2019-02-19
fix Enum infinite loop in |-->

antonkw:


antonkw:
Issue is described here: https://github.com/scalaz/scalaz/issues/2040

emilypi:
Thanks @antonkw !

PR #2042, opened 2019-02-28
override Functor#widen in scala std collection instances

xuwei-k:


xuwei-k:
added in https://github.com/scalaz/scalaz/commit/c8f70caadc0eace06152f8d17df652e3e632fc6a but removed in [support Scala 2.13 new collection](https://github.com/scalaz/scalaz/commit/08489c46f38233eccecafeb157a56fed4ae98fcb) unintentionally 🙇

PR #2045, opened 2019-03-17
Add ApplicativeError - Functor.compose - Traversable.compose

enzief:
A few things I need but found missing.


enzief:
I can happily move them.

base/shared/src/main/scala/scalaz/tc/functor.scala
```diff
@@ -38,3 +41,10 @@ trait FunctorSyntax {
     def void(implicit ev: Functor[F]): F[Unit]            = ev.map[A, Unit](self)(_ => ())
   }
 }
+
+class CompositionFunctorClass[F[_], G[_]](implicit F: FunctorClass[F], G: FunctorClass[G])
```
hrhino: I'd argue to make this private, like we have in sz7.

Issue #2044, opened 2019-03-09
dual of Functor#widen in Contravariant


xuwei-k:
I don't know what is suitable method name. any ideas? 🤔 

https://github.com/xuwei-k/scalaz/commit/ed2be5b3dff8b54dd222d07c22c4d8c6c7ae5b9e

```diff
diff --git a/core/src/main/scala/scalaz/Contravariant.scala b/core/src/main/scala/scalaz/Contravariant.scala
index bb62157f6..850b52d46 100644
--- a/core/src/main/scala/scalaz/Contravariant.scala
+++ b/core/src/main/scala/scalaz/Contravariant.scala
@@ -1,6 +1,7 @@
 package scalaz
 
 ////
+import scalaz.Liskov.<~<
 /**
  * Contravariant functors.  For example, functions provide a
  * [[scalaz.Functor]] in their result type, but a
@@ -26,6 +27,9 @@ trait Contravariant[F[_]] extends InvariantFunctor[F] { self =>
 
   // derived functions
 
+  def someMethod[A, B](fa: F[A])(implicit ev: B <~< A): F[B] =
+    contramap(fa)(ev.apply)
+
   def xmap[A, B](fa: F[A], f: A => B, g: B => A): F[B] =
     contramap(fa)(g)
 
diff --git a/core/src/main/scala/scalaz/std/Function.scala b/core/src/main/scala/scalaz/std/Function.scala
index c4584e50c..e71c421ec 100644
--- a/core/src/main/scala/scalaz/std/Function.scala
+++ b/core/src/main/scala/scalaz/std/Function.scala
@@ -1,6 +1,8 @@
 package scalaz
 package std
 
+import scalaz.Liskov.<~<
+
 sealed trait FunctionInstances1 {
   implicit def function1Semigroup[A, R](implicit R0: Semigroup[R]): Semigroup[A => R] =
     new Function1Semigroup[A, R] {
@@ -158,6 +160,8 @@ trait FunctionInstances extends FunctionInstances0 {
   implicit def function1Contravariant[R]: Contravariant[? => R] =
     new Contravariant[? => R] {
       def contramap[A, B](r: A => R)(f: B => A) = r compose f
+      override def someMethod[A, B](f: A => R)(implicit ev: B <~< A) =
+        Liskov.contra[-? => R, B, A](ev)(f)
     }
 
   implicit def function2Instance[T1, T2]: Monad[(T1, T2) => ?] with BindRec[(T1, T2) => ?] =

```

[...]

PR #2046, opened 2019-03-22
add Contravariant#narrow

xuwei-k:
fix https://github.com/scalaz/scalaz/issues/2044

### DIFF
--- a/core/src/main/scala/scalaz/std/Function.scala
+++ b/core/src/main/scala/scalaz/std/Function.scala
@@ -1,6 +1,8 @@
 package scalaz
 package std
 
+import scalaz.Liskov.<~<
+
 sealed trait FunctionInstances1 {
   implicit def function1Semigroup[A, R](implicit R0: Semigroup[R]): Semigroup[A => R] =
     new Function1Semigroup[A, R] {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/std/Function.scala
```diff
@@ -158,6 +160,8 @@ trait FunctionInstances extends FunctionInstances0 {
   implicit def function1Contravariant[R]: Contravariant[? => R] =
     new Contravariant[? => R] {
       def contramap[A, B](r: A => R)(f: B => A) = r compose f
+      override def narrow[A, B](f: A => R)(implicit ev: B <~< A) =
+        Liskov.contra[-? => R, B, A](ev)(f)
```
joroKr21: Can't we override this for other instances as well to avoid allocation? Maybe at least for `Kleisli`?

alexknvl: It is safe to define it as `asInstanceOf` in the parent. There are no lawful functors on JVM where `contramap(downcast)` is not the same as `asInstanceOf[F[B]]` modulo things that are unobservable in Scalazzi.

PR #2049, opened 2019-03-25
Define non-orphan typeclass instances with less boilerplate

enzief:
- Save some boilerplate when creating instances. For eg. `Monad` and `Traversable` which share `Functor`'s code.
- Users don't need to import `scalaz.Scalaz._` if they only need less powerful TC, for eg. `Functor`. This reduces implicit scope and speedup compile time.

To illustrate, console setup (note that there is no `import scalaz.Scalaz._`):
```scala
scala> import scalaz.tc._

scala> import scala.List

scala> def ft[F[_]](implicit F: Functor[F]): Functor[F] = F
ft: [F[_]](implicit F: scalaz.tc.Functor[F])scalaz.tc.Functor[F]

scala> def mn[F[_]](implicit F: Monad[F]): Monad[F] = F
mn: [F[_]](implicit F: scalaz.tc.Monad[F])scalaz.tc.Monad[F]
```
Before the change
```scala
scala> ft[List]
<console>:17: error: could not find implicit value for parameter F: scalaz.tc.Functor[List]
       ft[List]
```
After the change
```scala
scala> mn[List]
res0: scalaz.tc.Monad[List] = scalaz.tc.instances.list$$anon$1@382dacc3

scala> ft[List]
res1: scalaz.tc.Functor[List] = scalaz.tc.instances.list$$anon$1@382dacc3

scala> mf[List]
res2: scalaz.tc.Functor[List] = scalaz.tc.instances.list$$anon$1@382dacc3
```
In this PR only `List` has the new scheme of instances, but other data types can have the same too. Further steps might be to organize imports into `scalaz.syntax`, `scalaz.hierarchy`, etc. for more flexible usecases.

### DIFF
--- a/base/shared/src/main/scala/scalaz/tc/cobind.scala
+++ b/base/shared/src/main/scala/scalaz/tc/cobind.scala
@@ -20,12 +20,7 @@ object CobindClass {
       Some(f(fa))
   })
 
-  implicit val listCobind: Cobind[List] = instanceOf(new CobindClass[List] {
-    override def map[A, B](fa: List[A])(f: A => B): List[B] = fa.map(f)
-
-    override def cobind[A, B](fa: List[A])(f: List[A] => B): List[B] =
-      List(f(fa))
-  })
+  implicit val instanceList: Cobind[List] = instanceOf(instances.list.control)
 }
 
 trait CobindSyntax {
--- a/base/shared/src/main/scala/scalaz/tc/applicative.scala
+++ b/base/shared/src/main/scala/scalaz/tc/applicative.scala
@@ -1,8 +1,8 @@
 package scalaz
 package tc
 
+import scala.List
 import scala.Unit
-
 import scala.language.experimental.macros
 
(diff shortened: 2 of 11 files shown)


base/shared/src/main/scala/scalaz/tc/cobind.scala
```diff
@@ -20,12 +20,7 @@ object CobindClass {
       Some(f(fa))
   })
 
-  implicit val listCobind: Cobind[List] = instanceOf(new CobindClass[List] {
-    override def map[A, B](fa: List[A])(f: A => B): List[B] = fa.map(f)
```
enzief: Definition of `map` was duplicated in `Cobind`, `Monad`, and `Traversable`.

enzief:
I didn't mean to enhance support for `scala.List`. The idea is defining implicits in the typeclasses' companion object would save some import for users and some boilerplate for maintainers.

Instances for `List` only serve illustrating purpose.

PR #2050, opened 2019-03-26
PrioritySearchQueue

TomasMikula:
Priority queue that also supports lookup by key.

Implemented using `FingerTree`.

### DIFF
--- a/core/src/main/scala/scalaz/PrioritySearchQueue.scala
+++ b/core/src/main/scala/scalaz/PrioritySearchQueue.scala
@@ -1,0 +1,252 @@
+package scalaz
+
+import scala.annotation.tailrec
+import scalaz.syntax.order._
+import scalaz.syntax.show._
+
+import PrioritySearchQueue.Summary
+
+/** Container whose elements have two different orders: by priority and by lookup key.
+  *
+  * Supports efficient implementation of the following operations:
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/PrioritySearchQueue.scala
```diff
@@ -0,0 +1,253 @@
...
+  *
+  * @tparam A element type
+  * @tparam P priority of an element. Multiple elements can have the same priority.
+  * @tparam K lookup key. Unique—the queue holds at most one element with any given key.
+  */
+class PrioritySearchQueue[A, P, K] private(
+    private val tree: FingerTree[Summary[A, P, K], A],
+    priorityOf: A => P,
+    keyOf: A => K,
+) {
```
xuwei-k: compile error with Scala 2.11

```
[error] /home/travis/build/scalaz/scalaz/core/src/main/scala/scalaz/PrioritySearchQueue.scala:29:1: identifier expected but ')' found.
[error] ) {
[error] ^
```

https://travis-ci.org/scalaz/scalaz/jobs/511746376#L536

TomasMikula: Ah, trailing commas.

Issue #2051, opened 2019-04-03
Consistent datatypes in scalaz.


GodPlaysChess:
scalaz introduces better versions of the standard datatypes, such as
`==>>` (Map),
`Maybe`,
`IList` , but does not use it itself (for instance `Map` api returns `List` instead of `IList` and `Option` instead of `Maybe`) , which makes very cumbersome to utilise them.

I identified those places:

DataTypes:
- IList
- ==>>
- These
- Ephemeral Stream

Typeclasses:
- Enum
- Foldable

I am going to link corresponding PRs to this issue. 
I think it makes sense to change `Foldable` and `Enum` typeclass as well. What do you think?

PR #2055, opened 2019-04-05
consistent IList, ListT and Zipper

GodPlaysChess:
I left out `Option` in `List` since they both are part of standard scala collections. #2051

### DIFF
--- a/core/src/main/scala/scalaz/ListT.scala
+++ b/core/src/main/scala/scalaz/ListT.scala
@@ -1,14 +1,16 @@
 package scalaz
 
+import Maybe.just
+
 /**
  * ListT monad transformer.
  */
 
 final case class ListT[M[_], A](run: M[IList[A]]){
-  def uncons(implicit M: Applicative[M]): M[Option[(A, ListT[M, A])]] = {
+  def uncons(implicit M: Applicative[M]): M[Maybe[(A, ListT[M, A])]] = {
     M.map(run) {
-      case INil() => None
-      case ICons(listHead, listTail) => Some((listHead, new ListT(M.point(listTail))))
+      case INil() => Maybe.empty
+      case ICons(listHead, listTail) => just((listHead, new ListT(M.point(listTail))))
     }
   }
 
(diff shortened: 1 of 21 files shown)


core/src/main/scala/scalaz/ListT.scala
```diff
@@ -20,11 +22,11 @@ final case class ListT[M[_], A](run: M[IList[A]]){
...
   def headOption(implicit M: Functor[M]) : OptionT[M, A] = new OptionT(M.map(run)(_.headOption))
 
-  def find(predicate: A => Boolean)(implicit M: Functor[M]) : OptionT[M, A] = new OptionT(M.map(run)(_.find(predicate)))
+  def find(predicate: A => Boolean)(implicit M: Functor[M]) : MaybeT[M, A] = new MaybeT(M.map(run)(_.find(predicate)))
 
-  def headMaybe(implicit M: Functor[M]) : MaybeT[M, A] = new MaybeT(M.map(run)(l => Maybe.fromOption(l.headOption)))
+  def headMaybe(implicit M: Functor[M]) : MaybeT[M, A] = new MaybeT(M.map(run)(_.headMaybe))
 
-  def tailM(implicit M: Applicative[M]) : M[ListT[M, A]] = M.map(uncons)(_.get._2)
+  def tailM(implicit M: Applicative[M]) : M[ListT[M, A]] = M.map(uncons)(_.map(_._2).getOrElse(ListT.empty))
```
TomasMikula: This is changing the behavior from throwing an exception to returning an empty list.

Let's not make behavior changes as part of this PR.

core/src/main/scala/scalaz/ListT.scala
```diff
@@ -61,9 +63,9 @@ final case class ListT[M[_], A](run: M[IList[A]]){
     ListT(f(run))
 
   /**Don't use iteratively! */
-  def tail(implicit M: Functor[M]) : ListT[M, A] = new ListT(M.map(run)(_.tailOption.get))
+  def tail(implicit M: Functor[M]) : ListT[M, A] = new ListT(M.map(run)(_.tailMaybe.getOrElse(IList.empty)))
```
TomasMikula: Same change of behavior here.

GodPlaysChess: oki, I'll move it to separate PR.

PR #2056, opened 2019-04-07
safe methods in splitWith and selectSplit implementation#2051

GodPlaysChess:
This uses only safe methods in Foldable.splitWith and Foldable.selectSplit without changing the logic.

I need this in order to perform migration to scalaz IList ( #2051 ) which does not have unsafe `tail` and `head` methods.
Also I think with explicit pattern matching - the implementation is cleaner. For example in `selectSplit`
there's no need to pass `boolean` in accumulator, to check for the previous state, since it is guaranteed by nonemptiness of the List.

### DIFF
--- a/core/src/main/scala/scalaz/Foldable.scala
+++ b/core/src/main/scala/scalaz/Foldable.scala
@@ -329,14 +329,14 @@ trait Foldable[F[_]]  { self =>
    * Splits the elements into groups that alternatively satisfy and don't satisfy the predicate p.
    */
   def splitWith[A](fa: F[A])(p: A => Boolean): List[NonEmptyList[A]] =
-    foldRight(fa, (List.empty[NonEmptyList[A]], Maybe.empty[Boolean]))((a, b) => {
+    foldRight(fa, Maybe.empty[(NonEmptyList[NonEmptyList[A]], Boolean)])((a, b) => {
       val pa = p(a)
-      (b match {
-        case (_, Maybe.Empty()) => List(NonEmptyList(a))
-        case (Nil, _) => List(NonEmptyList(a))
-        case (x@(head :: tail), Maybe.Just(q)) => if (pa == q) (a <:: head) :: tail else NonEmptyList(a) :: x
-      }, Maybe.just(pa))
-    })._1
+      Maybe.just(
+        (b match {
+          case Maybe.Just((x, q)) => if (pa == q) NonEmptyList.nel(a <:: x.head, x.tail) else NonEmptyList(a) <:: x
+          case Maybe.Empty() => NonEmptyList(NonEmptyList(a))
+        }, pa))
+    }).cata(_._1.list.toList, List.empty)
 
   /**
     * Splits the elements into groups that produce the same result by a function f.
@@ -365,12 +365,14 @@ trait Foldable[F[_]]  { self =>
    * Selects groups of elements that satisfy p and discards others.
    */
   def selectSplit[A](fa: F[A])(p: A => Boolean): List[NonEmptyList[A]] =
-    foldRight(fa, List.empty[NonEmptyList[A]])((a, li) => (li, p(a)) match {
-      case (x :: xs, pa) =>
-        if (pa) (a <:: x) :: xs else li
-      case (Nil, pa) =>
-        if (pa) NonEmptyList(a) :: li else li
-    })
+    foldRight(fa, (List.empty[NonEmptyList[A]], true))((a, lb) => (lb, p(a)) match {
+      case ((l, _), false) =>
+        (l, true)
+      case ((x :: xs, false), true) =>
+        ((a <:: x) :: xs, false)
+      case ((l, _), true) =>
+        (NonEmptyList(a) :: l, false)
+    })._1
 
   /** ``O(n log n)`` complexity */
   def distinct[A](fa: F[A])(implicit A: Order[A]): IList[A] =
--- a/core/src/main/scala/scalaz/Foldable.scala
+++ b/core/src/main/scala/scalaz/Foldable.scala
@@ -1,5 +1,7 @@
 package scalaz
 
+import scalaz.Maybe.{Empty, Just, just}
+
 ////
 /**
  * A type parameter implying the ability to extract zero or more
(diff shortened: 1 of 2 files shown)


core/src/main/scala/scalaz/Foldable.scala
```diff
@@ -329,12 +329,13 @@ trait Foldable[F[_]]  { self =>
...
   def splitWith[A](fa: F[A])(p: A => Boolean): List[NonEmptyList[A]] =
-    foldRight(fa, (List[NonEmptyList[A]](), None : Option[Boolean]))((a, b) => {
+    foldRight(fa, (List.empty[NonEmptyList[A]], Maybe.empty[Boolean]))((a, b) => {
       val pa = p(a)
       (b match {
-        case (_, None) => NonEmptyList(a) :: Nil
-        case (x, Some(q)) => if (pa == q) (a <:: x.head) :: x.tail else NonEmptyList(a) :: x
-      }, Some(pa))
+        case (_, Maybe.Empty()) => List(NonEmptyList(a))
+        case (Nil, _) => List(NonEmptyList(a))
```
TomasMikula: This is an unreachable branch, right? Let's make illegal state unrepresentable by folding to `Maybe[(NonEmptyList[NonEmptyList[A]], Boolean)]` (instead of `(List[NonEmptyList[A]], Maybe[Boolean])`).

core/src/main/scala/scalaz/Foldable.scala
```diff
@@ -364,16 +365,12 @@ trait Foldable[F[_]]  { self =>
...
-            (a <:: x.head) :: x.tail else
-            NonEmptyList(a) :: x
-        else x, pa)
-      }
-    })._1
+    foldRight(fa, List.empty[NonEmptyList[A]])((a, li) => (li, p(a)) match {
+      case (x :: xs, pa) =>
+        if (pa) (a <:: x) :: xs else li
+      case (Nil, pa) =>
+        if (pa) NonEmptyList(a) :: li else li
```
TomasMikula: Doesn't this implementation return at most one `NonEmptyList`?

TomasMikula: Let's include some test to protect against it.

core/src/main/scala/scalaz/Foldable.scala
GodPlaysChess: Excellent! will do.

core/src/main/scala/scalaz/Foldable.scala
GodPlaysChess: If initial `F[_]` is empty - the result will be empty as well.

tests/src/test/scala/scalaz/FoldableTest.scala
```diff
@@ -281,6 +281,41 @@ object FoldableTest extends SpecLite {
...
+
+  "selectSplit: consistent with partition" ! forAll {
+    (l: List[Int], p: Int => Boolean) =>
+      L.selectSplit(l)(p).flatMap(_.toList) must_=== l.partition(p)._1
+  }
+
+  "selectSplit range consistent with filter" ! forAll {
+    (limit: Byte) =>
+      val l = Math.abs(limit.toInt)
+      val range = List.range(0, l)
```
TomasMikula: Could you please also take a random lower bound of the range? Otherwise, the predicate below (`_ % 2 != 0`) will never hold for the first element of this range, so it not random enough input to the method.

core/src/main/scala/scalaz/Foldable.scala
```diff
@@ -364,15 +365,13 @@ trait Foldable[F[_]]  { self =>
...
-            NonEmptyList(a) :: x
-        else x, pa)
-      }
+    foldRight(fa, (List.empty[NonEmptyList[A]], true))((a, lb) => (lb, p(a)) match {
+      case ((l, _), false) =>
+        (l, true)
+      case ((x :: xs, false), true) =>
+        ((a <:: x) :: xs, false)
+      case ((l, _), true) =>
+        (NonEmptyList(a) :: l, false)
```
TomasMikula: This last case (correctly) assumes that `((Nil, false), true)` never occurs. The previous implementation (correctly) assumed that the unsafe methods were in fact used safely. Should the assumption not hold for any reason, the previous implementation would at least fail fast with an exception. This, however, silently produces wrong result.

I suggest to write out the two cases, `((l, true), true)` and `((Nil, false), true)`, calling `sys.error` in the latter.

core/src/main/scala/scalaz/Foldable.scala
```diff
@@ -364,15 +365,13 @@ trait Foldable[F[_]]  { self =>
...
-    foldRight(fa, (List[NonEmptyList[A]](), false))((a, xb) => xb match {
-      case (x, b) => {
-        val pa = p(a)
-        (if (pa)
-          if (b)
-            (a <:: x.head) :: x.tail else
-            NonEmptyList(a) :: x
-        else x, pa)
-      }
+    foldRight(fa, (List.empty[NonEmptyList[A]], true))((a, lb) => (lb, p(a)) match {
```
TomasMikula: The cases below where `p(a)` is `true` and where it is `false` are disjoint. Checking `if (p(a))` first would save us allocating the `(lb, p(a))` tuple.

core/src/main/scala/scalaz/Foldable.scala
GodPlaysChess: Actually, I have thought that current implementation is more resilient, since if the assumption does not hold for any reason, then it creates new subNel with current element (there is no other way, because every element which satisfies the predicate, has to be in the result), and continues appending to it based on the predicate.
If you think, that's not the case, then I will change it to `sys.error`

core/src/main/scala/scalaz/Foldable.scala
GodPlaysChess: Right, will do.

core/src/main/scala/scalaz/Foldable.scala
TomasMikula: I see.

It is somewhat confusing then that it does not matter whether you start with `true` or `false`.

An auxiliary type like

```scala
sealed trait Acc {
  def add(a: A): Acc = ???
  def closeRun: Acc = ???
}
case class InsideRun(currentRun: NonEmptyList[A], tail: List[NonEmptyList[A]]) extends Acc
case class BetweenRuns(tail: List[NonEmptyList[A]]) extends Acc
```

would eliminate the ambiguity. If you find it an overkill, just leave it as is.

GodPlaysChess: actually we could write something like:
```  def selectSplit[A](fa: F[A])(p: A => Boolean): List[NonEmptyList[A]] =
    foldRight(fa, (List.empty[NonEmptyList[A]], List.empty[A]))((a, lb) =>
      if (p(a)) lb match {
        case (acc, l) =>
          (acc, a :: l)
      } else (scalaz.std.list.toNel(lb._2).toList ::: lb._1, List.empty)
    )._1.reverse```
This way we accumulate values in `List`, until the predicate fails. When it fails, we prepend this List to the result, and repeat procedure with empty list.
This way semantics of the accumulator stays clear and we do not need to use confusing boolean for tracking.

File: core/src/main/scala/scalaz/Foldable.scala [hunk 6]
```diff
@@ -363,17 +364,14 @@ trait Foldable[F[_]]  { self =>
...
-        else x, pa)
-      }
-    })._1
+  def selectSplit[A](fa: F[A])(p: A => Boolean): List[NonEmptyList[A]] = {
+    def squash(t: (List[NonEmptyList[A]], IList[A])): List[NonEmptyList[A]] = t._2.toNel.toList ::: t._1
+
+    squash(foldRight(fa, (List.empty[NonEmptyList[A]], IList.empty[A]))((a, l) =>
+      if (p(a)) (l._1, a :: l._2)
+      else (squash(l), IList.empty)
+    ))
```
TomasMikula: Very nice!

PR #2060, opened 2019-04-30
add fold(l|r)M1

enzief:


core/src/main/scala/scalaz/Foldable1.scala
```diff
@@ -70,6 +70,12 @@ trait Foldable1[F[_]] extends Foldable[F] { self =>
 
   def fold1[M: Semigroup](t: F[M]): M = foldMap1[M, M](t)(identity)
 
+  def foldlM1[M[_], A](fa: F[A])(f: (A, A) => M[A])(implicit M: Monad[M]): M[A] =
+    foldMapLeft1(fa)(M.point(_)) { (ma, a) => M.bind(ma)(f(_, a)) }
```
TomasMikula: To make the signature consistent with `Foldable#foldlM/foldLeftM` and `Foldable1#foldMapLeft1`, I think it should be

```scala
def foldLeftM1[M[_], A, B](fa: F[A])(z: A => B)(f: (B, A) => M[B])(implicit M: Monad[M]): M[B]
```

or

```scala
def foldlM1[M[_], A, B](fa: F[A])(z: A => B)(f: B => A => M[B])(implicit M: Monad[M]): M[B]
```

(or both). Analogously for `foldrM1` below.

enzief: Updated as requested. But why is `z: A => B` needed here while we can totally `.map` on `G[A]`?

TomasMikula: You mean `.map` on the result? Then you would need a function `(A, A) => G[A]`, as you had originally, which is less general than `(B, A) => G[B]`. So maybe you have `(B, A) => G[B]`, but not `(A, A) => G[A]`.

Now I noticed that `foldLeft1` doesn't use `z: A => B` either 😕. So consistency is IMO already broken (`foldLeft1` not being analogous to `foldLeft`). I see that you chose the signature analogous to `foldLeft1`, which is at least as good a choice as what I suggested. Feel free to revert.

enzief: Well I believe it is more useful as you suggested, as my first attempt was just copy/pasting from hackage. I don't see the need for reverting. Please merge if this is good enough, or I'll be happy to make further changes as requested.

enzief:
@TomasMikula is there anything else needed to merge this PR?

TomasMikula:
@enzief Looks good to me. I would like to give other collaborators a chance to have a look.

PR #2062, opened 2019-05-07
Adding filterM on Map

GrafBlutwurst:


core/src/main/scala/scalaz/Applicative.scala
```diff
@@ -70,6 +70,9 @@ trait Applicative[F[_]] extends Apply[F] with InvariantApplicative[F] { self =>
   def replicateM_[A](n: Int, fa: F[A]): F[Unit] =
     listInstance.sequence_(List.fill(n)(fa))(this)
 
+  /** Filter `map` according to an applicative predicate. **/
+  def filterM[A,B](map: A ==>> B)(f: B => F[Boolean])(implicit O:Order[A]):F[A ==>> B] = map.filterM(f)(this, O)
+
```
TomasMikula: I slightly don't like the cyclic dependency (`Applicative` depends on `Map` depends on `Applicative`), but I see that the same has already been done for `IList` and there are many more cyclic dependencies in scalaz, so OK.

PR #2063, opened 2019-05-10
Add IList#prepend.

### LINKED ISSUE #1950: IList.append is misnamed.
Rename it to `prepend` and add `append`.

jozi-k:
Closes #1950

### DIFF
--- a/tests/src/test/scala/scalaz/IListTest.scala
+++ b/tests/src/test/scala/scalaz/IListTest.scala
@@ -124,10 +124,6 @@ object IListTest extends SpecLite {
     ns.foldRight(s)(f) == ns.toList.foldRight(s)(f)
   }
 
-  "concat" ! forAll { (ns: IList[Int], ms: IList[Int]) =>
-    (ns concat ms).toList must_=== ns.toList ++ ms.toList
-  }
-
   "collect" ! forAll { (ns: IList[Int]) =>
     val pf: PartialFunction[Int, Int] = { case n if n % 2 == 0 => n + 1 }
     ns.collect(pf).toList must_=== ns.toList.collect(pf)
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/scalaz/IListTest.scala
```diff
@@ -287,6 +287,14 @@ object IListTest extends SpecLite {
     ns.prefixLength(f) must_=== ns.toList.prefixLength(f)
   }
 
+  "prepend" ! forAll { (ns: IList[Int], ms: IList[Int]) =>
+    (ns ++: ms).toList must_== ns.toList ++: ms.toList
+  }
```
TomasMikula: Seems like the exact same test is already on [line 100](https://github.com/scalaz/scalaz/pull/2063/files#diff-5b63ab3c5a167467caefa53a0c7d5eedR100).

tests/src/test/scala/scalaz/IListTest.scala
```diff
@@ -138,6 +134,10 @@ object IListTest extends SpecLite {
     ns.collectFirst(pf).toOption must_=== ns.toList.collectFirst(pf)
   }
 
+  "concat" ! forAll { (ns: IList[Int], ms: IList[Int]) =>
+    (ns concat ms).toList must_=== ns.toList ++ ms.toList
+  }
+
   "concat" ! forAll { (ns: IList[Int], ms: IList[Int]) =>
     (ns ++ ms).toList must_=== ns.toList ++ ms.toList
   }
```
TomasMikula: This one is also a duplicate of the test from line 96. (But I see that it was not you who introduced this one.)

jozi-k: Yes, firstly I wanted to get rid of it, but then I realized it might test that `concat` behaves same way as `++`. I do not see any problem with deleting it.

tests/src/test/scala/scalaz/IListTest.scala
jozi-k: Same as above, I wrote it to be sure `prepend` and `++=` should be indistinguishable.

tests/src/test/scala/scalaz/IListTest.scala
TomasMikula: I mean, they are exactly the same, both testing `++`.

Having one test for `concat` and a similar test for `++` is fine.

jozi-k:
Is there anything I can help with to conclude this PR?

TomasMikula:
Sorry, I haven't noticed your latest update. Looks good to me 👍

PR #2065, opened 2019-06-13
Consistent naming for Validation, Either and Disjunction.

jozi-k:


core/src/main/scala/scalaz/syntax/std/BooleanOps.scala
```diff
@@ -282,7 +282,10 @@ final class BooleanOps2(self: Boolean) {
    * Returns the first argument in `\/-` if this is `true`, otherwise the second argument in
    * `-\/`.
    */
-  final def either[A](a: => A): ConditionalEither[A] = new ConditionalEither(a)
+  final def toEither[A](a: => A): ConditionalEither[A] = new ConditionalEither(a)
+
+  @deprecated("Use `toEither`", "7.3.0")
+  final def either[A](a: => A): ConditionalEither[A] = toEither(a)
```
joroKr21: This one is a bit special. It's used as `bool.either(right).or(left)` so I don't think it should be renamed.

joroKr21:
I would also be interested to know if a PR deprecating `Maybe.cata` in favour of `fold` would be accepted :smile: 

It's been a thorn in my eyes for a long time.

core/src/main/scala/scalaz/syntax/std/BooleanOps.scala
jozi-k: nice catch @joroKr21 !

jozi-k:
I think you should ask at gitter channel about the `Maybe.cata` or just open PR and see what happens :)

tonymorris:
The original function was called `fold` before the stdlib started taking that name, and so was renamed to `cata` so that any implicit conversion had access to that function. An overall view of naming, such that it is consistent, and unlikely to clash with a future stdlib (unicode? :)) is definitely warranted.

jozi-k:
Can someone from maintainers have a brief look at this PR please?

core/src/main/scala/scalaz/syntax/std/EitherOps.scala
```diff
@@ -6,7 +6,10 @@ final class EitherOps[A, B](private val self: Either[A, B]) extends AnyVal {
 
   final def disjunction: A \/ B = \/ fromEither self
```
alexknvl: Shouldn't this one also be `toDisjunction`?

jozi-k: Great observation! I added it to newest version of this PR.

jozi-k:
PR has been updated. Asking for another round of reviews :)

PR #2070, opened 2019-07-01
Fix some spelling errors in Adjunct Usage

Jwhiles:
I noticed a few mistakes when reading through the adjunct usage file.


TomasMikula:
👍

PR #2073, opened 2019-07-14
 Add Apply and Applicative instances for EitherT 

joroKr21:


xuwei-k:
see https://github.com/scalaz/scalaz/pull/255

joroKr21:
Hmm, this means one can't run `EitherT[ParallelTask, Error, A]`, well... in parallel :thinking:

PR #2075, opened 2019-07-27
 Applicative.Par instance for EitherT

joroKr21:
Since we can't have the bare `Applicative` instance for `EitherT`
due to potential inconsistencies, maybe we can have the
`Applicative.Par` instance?

### DIFF
--- a/core/src/main/scala/scalaz/EitherT.scala
+++ b/core/src/main/scala/scalaz/EitherT.scala
@@ -376,6 +376,9 @@ sealed abstract class EitherTInstances extends EitherTInstances0 {
 
   implicit def eitherTShow[F[_], A, B](implicit F0: Show[F[A \/ B]]): Show[EitherT[F, A, B]] =
     Contravariant[Show].contramap(F0)(_.run)
+
+  implicit def eitherTParallelApplicative[F[_], E](implicit F0: Applicative.Par[F]): Applicative.Par[EitherT[F, E, ?]] =
+    new EitherTParallelApplicative[F, E] { def F = F0 }
 }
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/scalaz/Applicative.scala
```diff
@@ -131,5 +131,22 @@ object Applicative {
 
   implicit def monoidApplicative[M:Monoid]: Applicative[λ[α => M]] = Monoid[M].applicative
 
+  implicit def eitherTParallelApplicative[F[_], E](implicit F0: Applicative.Par[F]): Applicative.Par[EitherT[F, E, ?]] =
```
xuwei-k: please define in `EitherT.scala`

joroKr21: It doesn't work  :disappointed: 

The reason is:
```scala
type Par[F[_]] = Applicative[λ[α => F[α] @@ Tags.Parallel]]
type @@[T, Tag] = Tagged[T, Tag]
private[scalaz] type Tagged[A, T] = {type Tag = T; type Self = A}
// =>
Applicative.Par[EitherT[F, E, ?]] = Applicative[λ[α => {type Tag = Parallel; type Self = EitherT[F, E, α]}]]
```

But type member (`Self`) is not in implicit scope and therefore `EitherT` is not in implicit scope.

joroKr21: @xuwei-k see #2076

joroKr21: I think I wasn't quite clear, #2076 would enable to move the instance to `EitherT`

joroKr21:
Now moved to `EitherTInstances` and it works :tada:

PR #2121, opened 2020-12-22
macros to generate IsoSet for case classes and sealed traits

fommil:
This is a novel typeclass derivation mechanism that is much simpler than scalaz-deriving, magnolia, and shapeless, yet slightly more powerful than Magnolia (and thus scalaz-deriving). I am working up this idea in a standalone repository, but it seems that this could also be part of scalaz, see https://gitlab.com/fommil/bourbaki/ for the standalone variant.

I plan on generating further boilerplate so that for any typeclass having a Divisible or Applicative, it may be derived for any case class with a single line (!!! yes, really !!!). Likewise for Decidable and Alt for sealed traits. It is also possible to use this mechanism to derive non-lawful typeclasses such as JSON encoders and decoders (but it benefits from https://github.com/propensive/adversaria and `TypeName`). Therefore, this PR if accepted, deprecates scalaz-deriving.

@xuwei-k I would love to hear your thoughts on this, and some help with the following:

- [ ] I can't get the tests to compile (they compile in my standalone variant, so this is scalaz specific)
- [ ] the macro should work in scala 2.12 and below, even though it uses singleton type literals, but people can't do anything with it because there is no ValueOf, but I think it can be backported using the technique that shapeless used prior to its inclusion in scala 2.13
- [ ] it should be possible to rewrite the macro for scala 3, I need the most help here
- [ ] my changes to the build are unhygienic

### DIFF
--- a/core/src/main/scala/scalaz/Isomorphism.scala
+++ b/core/src/main/scala/scalaz/Isomorphism.scala
@@ -1,5 +1,7 @@
 package scalaz
 
+import scala.language.experimental.macros
+
 sealed abstract class Isomorphisms {
 
   /**Isomorphism for arrows of kind * -> * -> * */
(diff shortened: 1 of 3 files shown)


fommil:
closed due to lack of interest.

joroKr21:
You could make `V1 extends OneOf1 with OneOf2 ... with OneOf22`, `V2 extends OneOf2 with OneOf3 ... with OneOf22` etc. to generate less classes.

fommil:
@joroKr21 that's not a bad idea but it would disallow adhoc extending to larger arity. I'm not particulraly concerned about class counts.

PR #2224, opened 2021-11-19
Add the StreamTMergeMonoid instance (for #2223)

Atry:
This instance could provide the functionality similar to the [ReactiveX Merge operator](http://reactivex.io/documentation/operators/merge.html)

Issue #2223, opened 2021-11-19
ReactiveX operators for StreamT


Atry:
I found that [ReactiveX operators](http://reactivex.io/documentation/operators.html) are conceptually similar to `StreamT`. Shall 
we add these operators into `StreamT`?

I suppose our goal is to support the use cases similar to ReactiveX's, but not necessarily the same names of operators. There have been `StreamT` methods corresponding to most of ReactiveX filtering operators, and aggregating operators can be implemented easily with the help of `foldLeft`. I think the missing part is just those operators depending on `Nondeterminism`:
- [x] #2224 
- [x] #2225
- [x] #2227
- [ ] #2231

Atry:
I think we might want to reopen this issue. I should not have let all the related PRs "fix" this issue

PR #2225, opened 2021-11-19
Add the StreamTMergeMonad instance (for #2223)

Atry:
This PR adds the ReactiveX `flapMap` operator (named `mergeMap` in this PR), and the corresponding `StreamTMergeMonoid` instance

See 
http://reactivex.io/documentation/operators/flatmap.html

This PR is based on #2224


Atry:
Rebased

PR #2227, opened 2021-11-19
Add StreamT.distinctUntilChanged (fix #2223)

### LINKED ISSUE #2223: ReactiveX operators for StreamT

Atry:
This PR implements the [ReactiveX distinctUntilChanged operator](http://reactivex.io/documentation/operators/distinct.html)

### DIFF
--- a/core/src/main/scala/scalaz/StreamT.scala
+++ b/core/src/main/scala/scalaz/StreamT.scala
@@ -219,6 +219,28 @@ sealed class StreamT[M[_], A](val step: M[StreamT.Step[A, StreamT[M, A]]]) {
       }
     }
 
+  def distinctUntilChanged(latest: A)(implicit M: Functor[M], A: Equal[A]): StreamT[M, A] =
+    stepMap {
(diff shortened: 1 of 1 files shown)


Atry:
Current [Binding.scala](https://github.com/ThoughtWorksInc/Binding.scala) internally depends on mutable states, which is error-prone. This `StreamT.distinctUntilChanged` operator can be used as an immutable implementation of data-binding, which is more robust than the current Binding.scala implementation.

PR #2231, opened 2021-11-20
Add StreamT.flatMapLatest (for #2223)

Atry:
This PR adds the ReactiveX `FlatMapLatest` operator into `StreamT`.

This operator would be useful to implement data-binding.


Atry:
Note that this operator is critical for data-binding, where only the latest value matters

TomasMikula (approved):
LGTM

Atry:
Hi @xuwei-k , any feedback?

PR #2249, opened 2021-12-18
Add StreamT.collect

Atry:
Just a convenient utility, similar to `scala.collection.Seq.collect`


Atry:
Thank you for adding the test

PR #2250, opened 2021-12-19
Add StreamT.scan and StreamT.scanLeft

Atry:
scan


Atry:
I am not sure if we should provide `scanRight`. There is a `scanRight` on `LazyList`, which would evaluate the entire list. I don't think people would normally use it.

Atry:
Thank you for adding the test!

PR #2251, opened 2021-12-20
Add StreamT.liftM, which accepts Applicative instead of Monad

Atry:


TomasMikula:
```
[error] /home/runner/work/scalaz/scalaz/core/src/main/scala/scalaz/StreamT.scala:323:51: object creation impossible. Missing implementation for:
[error]   def liftM[G[_], A](a: G[A])(implicit evidence$1: scalaz.Monad[G]): scalaz.StreamT[G,A] // inherited from trait MonadTrans
[error]   implicit val StreamTHoist: Hoist[StreamT] = new StreamTHoist {}
[error]            
```

Atry:
Updated

PR #2262, opened 2021-12-29
Add Nondeterminism instance for Kleisli/ReaderT

Atry:


core/src/main/scala/scalaz/Kleisli.scala
```diff
@@ -368,6 +374,20 @@ private trait KleisliMonad[F[_], R] extends Monad[Kleisli[F, R, *]] with Kleisli
...
 
+
+private trait KleisliNondeterminism[F[_], R] extends Nondeterminism[Kleisli[F, R, *]] with KleisliMonad[F, R]{
+  implicit def F: Nondeterminism[F]
+
+  def chooseAny[A](head: Kleisli[F, R, A], tail: IList[Kleisli[F, R, A]]): Kleisli[F, R, (A, IList[Kleisli[F, R, A]])] =
+    Kleisli { r =>
+      F.map(F.chooseAny(head(r), tail.map(_(r)))) {
+        case (chosen, notChosen) =>
+          (chosen, notChosen.map(fa => Kleisli(_ => fa)))
```
Atry, line 385: `Kleisli(_ => fa)` might be incorrect, because it discard the any new values.

Shall we change the required function of `Nondeterminism` to `firstCompletedOf` as mentioned by #2255

It is hard to implement `chooseAny` correctly and efficiently

PR #2271, opened 2021-12-30
Let StreamT be nonstrict

Atry:
This diff makes StreamT be a nonstrict SAM type, similar to `LazyList` and `EphemeralStream`, unlike the deprecated `Stream`.


Atry:
Maybe we should introduce another type like `LazyListT` instead of changing the behavior of existing `StreamT`. What do you think?

PR #2273, opened 2021-12-31
Add StreamT.weakMemoize

Atry:


Atry:
Blocked by https://github.com/scala-js/scala-js/pull/4617

Atry:
`EphemeralStream.weakMemo` is not yet support Scala.js. However we didn't have a test running `weakMemo`.

armanbilge:
FYI you can use this project https://github.com/scala-js/scala-js-weakreferences/. I don't think they can accept that PR to core for historical reasons.

Atry:
Added the dependency to fix the test

Atry:
It's an optional dependency so it would not automatically introduce the dependency for scalaz users

Atry:
Rebased to trigger CI

PR #2275, opened 2021-12-31
Let codensityMonadPlus depend on PlusEmpty instead of ApplicativePlus

Atry:
ApplicativePlus is unnecessary


xuwei-k:
https://github.com/scalaz/scalaz/blob/be44a2b6b8a0f66746c34ce5fa7ea56ce09dd1df/core/src/main/scala/scalaz/Codensity.scala#L33-L35 🤔

Atry:
This instance holds neither left distribution law nor left catch law

kenji yoshida ***@***.***>于2021年12月30日 周四下午11:47写道：

Atry:
> https://github.com/scalaz/scalaz/blob/be44a2b6b8a0f66746c34ce5fa7ea56ce09dd1df/core/src/main/scala/scalaz/Codensity.scala#L33-L35
> 
> 🤔

I think the comment is correct because the property test depends on the definition of `Equal[Codensity[F, A]]`, which depends on `Applicative[F]`.

PR #2276, opened 2021-12-31
Add BindRec instance for Codensity

Atry:
`IntrusiveEphemeralStream` is an ephemeral stream whose `head` is `this`.

### DIFF
--- a/core/src/main/scala/scalaz/Codensity.scala
+++ b/core/src/main/scala/scalaz/Codensity.scala
@@ -1,5 +1,7 @@
 package scalaz
 
+import scala.annotation.tailrec
+
 abstract class Codensity[F[_], A] { self =>
   def apply[B](f: A => F[B]): F[B]
   def improve(implicit F: Applicative[F]): F[A] =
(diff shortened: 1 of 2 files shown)


Atry:
This trick can be used to implement `BindRec` for any non-final types

core/src/main/scala/scalaz/Codensity.scala
```diff
@@ -20,6 +31,23 @@ abstract class Codensity[F[_], A] { self =>
 }
 
 object Codensity extends CodensityInstances {
+
+  private[Codensity] abstract class IntrusiveEphemeralStream[F[_], A]
+      extends Codensity[F, A] {
+    def tail: Codensity[F, A]
+    @tailrec final def last: Codensity[F, A] = {
+      tail match {
+        case stream: IntrusiveEphemeralStream[F, A] =>
```
Atry, line 40: We cannot let `IntrusiveEphemeralStream` be generic because this type check must be specialized for each type

PR #2278, opened 2021-12-31
Add StreamT.noSkip and StreamT.noSkipRec

Atry:
It would be helpful for creating recursive `StreamT`s, otherwise it would create infinite `M.map` calls with `Skip` steps.


Atry:
Rebased to trigger CI

PR #2336, opened 2022-03-27
Add the BindRec instance for Future

Atry:
This PR implements `tailrecM` by running a tailrec loop instead of `flatMap` when `Future.value` is not `None`.


Atry:
I am merging this PR, because it seems low risk and I did not a review in 1 year.

PR #2341, opened 2022-04-02
Allocate 90% of memory to sbt

Atry:
Hopefully it would fix CI failures in https://github.com/scalaz/scalaz/runs/5795544306?check_suite_focus=true


xuwei-k:
- https://github.com/scalaz/scalaz/pull/2767

PR #2342, opened 2022-04-03
Make StreamT a SAM type

Atry:
This PR would reduce the size of byte code


Atry:
Not that I did not create a SAM type for `memoize` to avoid the `LazyRef` generated by the compiler.

PR #2346, opened 2022-04-06
Allocate 90% of memory to sbt

Atry:
I created this PR to trigger Github Actions for Scala 3.1.2 with more memory allocated


Atry:
It still results an out-of-memory error when running `publishLocal`. I think we should split `test` and `publishLocal` into two sbt commands, supposing there would be some memory leaks in `sbt test`.

PR #2381, opened 2022-07-28
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/scalaz/commit/ecfa03af75ed2b2f295476911f4be4dc67325e35#commitcomment-79693156

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/scalaz/commit/f440cae75d2df569f4251a5ddfa5134bc560a1d7

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

### DIFF
--- a/core/src/main/scala/scalaz/Heap.scala
+++ b/core/src/main/scala/scalaz/Heap.scala
@@ -400,7 +400,7 @@ object Heap extends HeapInstances {
     }
 
     def skewMeld[A](f: (A, A) => Boolean, ts: Forest[A], tsp: Forest[A]) =
-      unionUniq(f)(uniqify(f)(ts), uniqify(f)(tsp))
+      unionUniq(f)(uniquify(f)(ts), uniquify(f)(tsp))
 
     def ins[A](f: (A, A) => Boolean, t: Tree[Ranked[A]]): Forest[A] => Forest[A] = {
       case s if s.isEmpty    => EphemeralStream(t)
@@ -409,7 +409,7 @@ object Heap extends HeapInstances {
         ins(f, link(f)(t, tp))(ts)
     }
 
-    def uniqify[A](f: (A, A) => Boolean): Forest[A] => Forest[A] = {
+    def uniquify[A](f: (A, A) => Boolean): Forest[A] => Forest[A] = {
       case s if s.isEmpty   => emptyEphemeralStream
       case (t ##:: ts) => ins(f, t)(ts)
     }
--- a/example/src/main/scala/scalaz/example/ContravariantCoyonedaUsage.scala
+++ b/example/src/main/scala/scalaz/example/ContravariantCoyonedaUsage.scala
@@ -538,7 +538,7 @@ object ContravariantCoyonedaUsage {
       s.foldMap{case (st, i) => recItem[BinOrd](i, sortTypeBinOrd(st))}
 
     // The drawback here is that I can’t just build a separate stack
-    // willynilly for `Binfmt'.  I have to prove at each step that *the
+    // willy nilly for `Binfmt'.  I have to prove at each step that *the
     // same* `I' is used for the `Binfmt' and the `Order' for this to be
     // useful.  Once again, an idea of a fold step that can be fused
     // with the `Order' construction safely can be abstracted out here.
--- a/example/src/main/scala/scalaz/example/KleisliUsage.scala
+++ b/example/src/main/scala/scalaz/example/KleisliUsage.scala
@@ -25,7 +25,7 @@ object KleisliUsage {
       Continent("Asia",
         List(
           Country("India",
-            List(City("New Dehli"), City("Calcutta"))))))
+            List(City("New Delhi"), City("Calcutta"))))))
 
     def continents(name: String): List[Continent] =
       data.filter(k => k.name.contains(name))
--- a/example/src/main/scala/scalaz/example/ReaderWriterStateTUsage.scala
+++ b/example/src/main/scala/scalaz/example/ReaderWriterStateTUsage.scala
@@ -4,7 +4,7 @@ package scalaz.example
  A rather contrived example which shows how ReaderWriterStateT could be used.
  @author stew@vireo.org / stew@helloreverb.com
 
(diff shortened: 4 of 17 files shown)


example/src/main/scala/scalaz/example/ContravariantCoyonedaUsage.scala
```diff
@@ -538,7 +538,7 @@ object ContravariantCoyonedaUsage {
       s.foldMap{case (st, i) => recItem[BinOrd](i, sortTypeBinOrd(st))}
 
     // The drawback here is that I can’t just build a separate stack
-    // willynilly for `Binfmt'.  I have to prove at each step that *the
+    // willy nilly for `Binfmt'.  I have to prove at each step that *the
```
jsoref, line 541: Google Sheets made this (and most of the other corrections). I probably wouldn't have actively made it, but it's right.

example/src/main/scala/scalaz/example/KleisliUsage.scala
```diff
@@ -25,7 +25,7 @@ object KleisliUsage {
       Continent("Asia",
         List(
           Country("India",
-            List(City("New Dehli"), City("Calcutta"))))))
+            List(City("New Delhi"), City("Calcutta"))))))
```
jsoref, line 28: Proper name...

example/src/main/scala/scalaz/example/ReaderWriterStateTUsage.scala
```diff
@@ -28,7 +28,7 @@ case object B extends Token
 
 object Token {
 
-  implicit val euqalsRef: Equal[Token] = Equal.equalRef
+  implicit val equalsRef: Equal[Token] = Equal.equalRef
```
jsoref, line 31: ?

tests/src/test/scala/scalaz/DListTest.scala
```diff
@@ -38,7 +38,7 @@ object DListTest extends SpecLite {
     def monoid[A] = Monoid[DList[A]]
     def monadPlus = MonadPlus[DList]
     def alt = Alt[DList]
-    def bindrec = BindRec[DList]
+    def bindRec = BindRec[DList]
```
jsoref, line 41: This change is the scariest (it occurs in three files).

I can't understand how this code could compile in both forms. Based on the surrounding code, this change appears to be correct, but 🤷 .

tests/src/test/scala/scalaz/std/ListTest.scala
```diff
@@ -49,7 +49,7 @@ object ListTest extends SpecLite {
       a.groupWhenM[Id](p).map(_.list.toList).flatten must_===(a)
   }
 
-  "groupByWhenM[Id] ∀(i,j) | 0<i<resut.len & 0<j<result(i).len: p(result(i)(j), p(result(i)(j+1)) yields true" ! forAll {
+  "groupByWhenM[Id] ∀(i,j) | 0<i<result.len & 0<j<result(i).len: p(result(i)(j), p(result(i)(j+1)) yields true" ! forAll {
```
jsoref, line 52: ?

jsoref (commented):
Most changes were automatically suggested by Google Sheets. All fault mine.

core/src/main/scala/scalaz/Heap.scala
```diff
@@ -409,7 +409,7 @@ object Heap extends HeapInstances {
         ins(f, link(f)(t, tp))(ts)
     }
 
-    def uniqify[A](f: (A, A) => Boolean): Forest[A] => Forest[A] = {
+    def uniquify[A](f: (A, A) => Boolean): Forest[A] => Forest[A] = {
```
Atry, line 412: This is a public API. We should keep the original method for backward compatibility.

Atry (changes requested):
We should not rename public API. It would be nice to create a new method with the correct name and deprecate the old typo method.

Jamal985610 (approved):

Atry (changes requested):
Still contains public API changes

core/src/main/scala/scalaz/Heap.scala
jsoref: I've added a deprecated method with the original name that calls this method: https://github.com/scalaz/scalaz/compare/3b0bbbf7de95f0b02288c57aa839462f2deca9e3..561f47ff91cdcc603cf9ac048b831e88bf504730

Does that do what you're looking for?

Are there other methods that need this treatment?

scalacheck-binding/src/main/scala/scalaz/scalacheck/ScalazArbitrary.scala
```diff
@@ -492,7 +492,7 @@ object ScalazArbitrary {
     Gen.sized(fingerTree[A] _)
   }
 
-  implicit def IndSeqArbibrary[A: Arbitrary]: Arbitrary[IndSeq[A]] = Functor[Arbitrary].map(arb[List[A]])(IndSeq.fromSeq)
+  implicit def IndSeqArbitrary[A: Arbitrary]: Arbitrary[IndSeq[A]] = Functor[Arbitrary].map(arb[List[A]])(IndSeq.fromSeq)
```
Atry, line 495: ABI breakage

jsoref: Dropped. I'll have to look into how scala expects to handle such things...

tests/src/test/scala/scalaz/DListTest.scala
Atry: It's just a unused method to test whether `BindRec[DList]` compiles or not. So any name would be OK.

Atry (approved):

Atry:
Looks good to me.
@xuwei-k What do you think?

I am not sure if we could simply rename a public API considering the `master` branch does not promise backward compatibility.

PR #2467, opened 2023-01-25
Let `tailrecM` create `Future.failed` only once, not twice, when catching an exception

Atry:


Atry:
Trivial change and it passes tests. Should be safe to merge.

PR #2483, opened 2023-02-12
Added type annotations to the code

aashish2054:
Added type annotations to the code to make it more concise and readable.

### DIFF
--- a/core/src/main/scala/scalaz/std/Either.scala
+++ b/core/src/main/scala/scalaz/std/Either.scala
@@ -141,7 +141,7 @@ trait EitherInstances extends EitherInstances0 {
 
   }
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/scalaz/std/Either.scala
```diff
@@ -141,7 +141,7 @@ trait EitherInstances extends EitherInstances0 {
 
   }
 
-  implicit def eitherShow[A,B](implicit SA: Show[A], SB: Show[B]) : Show[Either[A,B]] = {
+  implicit def eithe9rShow[A,B](implicit SA: Show[A], SB: Show[B]) : Show[Either[A,B]] = {
```
xuwei-k, line 144: ?

